PR #2, opened 2018-07-16
Major rewrite

jwilander:
Still needs some work but a big step forward.


jwilander:
I need to update the README before we merge this too

plugin.json
```diff
@@ -1,40 +1,64 @@
...
+                "help_test": "Select the username of the user that the plugin will post with. This can be any user, the name and icon will be overridden when posting."
+            },
+            {
+                "key": "EncryptionKey",
+                "display_name": "At Rest Encryption Key",
+                "type": "generated",
+                "help_text": "The AES encryption key used to encrypt stored access tokens."
+            },
+            {
+                "key": "GithubOrg",
```
lieut-data: nit: capitalize `Hub`?

server/api.go
```diff
@@ -0,0 +1,399 @@
...
+		mlog.Error(err.Error())
+	}
+
+	// Post intro post
+	message := fmt.Sprintf("#### Welcome to the Mattermost GitHub Plugin!\n You've connected your Mattermost account to [%s](%s) on GitHub. Read about the features of this plugin below:\n\n##### Daily Reminders\nThe first time you log in each day, you will get a post right here letting you know what messages you need to read and what pull requests are awaiting your review.\nTurn off reminders with `/github settings reminders off`.\n\n##### Notifications\nWhen someone mentions you, requests your review, comments on or modifies one of your pull requests/issues, or assigns you, you'll get a post here about it.\nTurn off notifications with `/github settings notifications off`.\n\n##### Sidebar Buttons\nCheck out the buttons in the left-hand sidebar of Mattermost.\n* The first button tells you how many pull requests are awaiting your review\n* The second tracks the number of unread messages you have\n* The third will refresh the numbers\n\nClick on them!\n\n##### Slash Commands\n"+strings.Replace(COMMAND_HELP, "|", "`", -1), gitUser.GetLogin(), gitUser.GetHTMLURL())
```
lieut-data: nit: backticks would be easier to parse here

server/subscriptions.go
```diff
@@ -2,67 +2,158 @@ package main
...
+		subscriptions = &Subscriptions{Repositories: map[string][]*Subscription{}}
 	} else {
 		json.NewDecoder(bytes.NewReader(value)).Decode(&subscriptions)
 	}
 
 	return subscriptions, nil
 }
 
-func (s *Subscriptions) StoreInKVStore(store plugin.KeyValueStore) error {
+func (p *Plugin) StoreSubscriptions(s *Subscriptions) error {
```
lieut-data: Is there a potential for a race condition here? Either between multiple requests to the plugins, or when the plugin is deployed in an HA environment?

webapp/client/client.js
```diff
@@ -5,20 +5,41 @@ export default class Client {
...
+    getMentions = async () => {
+        return this.doGet(`${this.url}/mentions`);
+    }
+
+    getUnreads = async () => {
+        return this.doGet(`${this.url}/unreads`);
+    }
+
+    doGet = async (url, body, headers = {}) => {
+        headers['X-Requested-With'] = 'XMLHttpRequest';
```
lieut-data: What purpose does `X-Requested-With` have?

jwilander: Great question, ongoing thread specifically discussing that here https://pre-release.mattermost.com/core/pl/3wbotmckrf87mekbuse4wun9sc

server/subscriptions.go
jwilander: Yes there is, and also in other places such as updating a user's settings (though that one's less likely to be an issue). We could probably split the subscriptions by repo to reduce the chance of a race causing problems but that won't eliminate it. To eliminate it we'd need a way of atomically updating them in the KV store. Thoughts?

server/api.go
jwilander: Would be but you can't escape backticks inside backticks unfortunately (hence my whole replacing | with backticks for COMMAND_HELP)

lieut-data: Aha, learned something new! 👍

server/subscriptions.go
lieut-data: I'm not sure of a way around this for the KV store. It seems like it might be easier just to formally give plugins database access and use transactions.

jwilander: Yeah that's an option, something to look into further for sure

server/api.go
```diff
@@ -0,0 +1,399 @@
...
+		http.Error(w, "missing stored state", http.StatusBadRequest)
+		return
+	} else if string(storedState) != state {
+		http.Error(w, "invalid state", http.StatusBadRequest)
+		return
+	}
+
+	userID := strings.Split(state, "_")[1]
+
+	p.API.KVDelete(state)
```
crspeller: I'ts a rare event, but it is possible for a key not to be deleted because the user agent never makes it here. Might be a good idea to allow some kind of mechanism for cleaning up old keys.

webapp/components/sidebar_buttons/sidebar_buttons.jsx
```diff
@@ -0,0 +1,188 @@
+const React = window.react;
```
crspeller: We should use the webpack externals functionality.

server/api.go
jwilander: Yeah I was thinking about that from the general plugin perspective, might be nice to add APIs such as `KVStoreWithExpiry()` so that plugins don't have to worry about it

crspeller:
This is awesome. It even supports GitHub enterprise (although I guess you probably couldn't test that)

lieut-data:
Heads up that I've added a commit to switch to webpack externals, as per mattermost/mattermost-webapp#1473.

PR #9, opened 2018-09-10
Proposed README tweaks

it33:
Updated description of project and propose moving Hackathon mention to the bottom, rather than having it at the top, to increase trust level in the plug-in--we can acknowledge this was first developed in a Hackathon, but now it's a standard plug-in.


jwilander:
Closing as README already covers this

Issue #21, opened 2018-10-12
Plugin failed to ServeHTTP


bensig:
Running on 5.3.1 fails with this error:

{"level":"error","ts":1539313420.9237213,"caller":"plugin/client_rpc.go:154","msg":"RPC call to OnActivate plugin failed.","plugin_id":"github","error":"reading body gob: name not registered for interface: \"*plugin.ErrorString\""}
{"level":"error","ts":1539313421.156202,"caller":"plugin/client_rpc.go:295","msg":"Plugin failed to ServeHTTP, RPC call failed","plugin_id":"github","error":"connection is shut down"}
{"level":"error","ts":1539313421.16836,"caller":"plugin/client_rpc.go:295","msg":"Plugin failed to ServeHTTP, RPC call failed","plugin_id":"github","error":"connection is shut down"}
{"level":"error","ts":1539313421.187666,"caller":"plugin/client_rpc.go:295","msg":"Plugin failed to ServeHTTP, RPC call failed","plugin_id":"github","error":"connection is shut down"}
{"level":"error","ts":1539313421.194268,"caller":"plugin/client_rpc.go:295","msg":"Plugin failed to ServeHTTP, RPC call failed","plugin_id":"github","error":"connection is shut down"}
{"level":"error","ts":1539313421.2369359,"caller":"plugin/client_rpc.go:295","msg":"Plugin failed to ServeHTTP, RPC call failed","plugin_id":"github","error":"connection is shut down"f}
{"level":"error","ts":1539313421.244325,"caller":"plugin/client_rpc.go:295","msg":"Plugin failed to ServeHTTP, RPC call failed","plugin_id":"github","error":"connection is shut down"}
{"level":"error","ts":1539313421.322338,"caller":"plugin/client_rpc.go:295","msg":"Plugin failed to ServeHTTP, RPC call failed","plugin_id":"github","error":"connection is shut down"}
{"level":"error","ts":1539313426.156944,"caller":"plugin/client_rpc.go:268","msg":"Plugin failed to ServeHTTP, muxBroker couldn't Accept request body connecion","plugin_id":"github","error":"timeout waiting for accept"}
{"level":"info","ts":1539313436.7515883,"caller":"utils/i18n.go:83","msg":"Loaded system translations for 'en' from '/opt/mattermost/i18n/en.json'"}

bensig:
My mattermost server keeps crashing when trying to run this plugin on 5.3.1 - any ideas?

[...]

PR #23, opened 2018-10-15
Downstream https://github.com/mattermost/mattermost-plugin-sample

### LINKED ISSUE #20: Make prerequisites
make dist fails without the following "go get"s 

go get github.com/google/go-github/github
go get github.com/mattermost/mattermost-server
go get golang.org/x/oauth2

hanzei:
This PR sync the structure of the plugin with https://github.com/mattermost/mattermost-plugin-sample. Note that this PR is mostly for consistency.

It also updates some dependencies.

I did some smoke tests on my own server and it seamed, if everything works as expected. But since this is a major rewrite, it should be tested carefully.

Fixes #20


lieut-data:
This is awesome, @hanzei! I wonder if you could update the dependencies here to 5.4.0 (and if you're so inclined, upstream as well ;)). I'm hoping the new error handling might assist with debugging https://github.com/mattermost/mattermost-plugin-github/issues/21, but we need both 5.4.0 and the plugin itself to be upgraded.

hanzei:
Sure, done

PR #30, opened 2018-11-06
Update user_attribute.jsx

chetanyakan:


webapp/src/components/user_attribute/user_attribute.jsx
```diff
@@ -8,8 +8,6 @@ export default class UserAttribute extends React.PureComponent {
     };
 
     render() {
-        const style = getStyle(this.props.theme);
```
jwilander: If we're going to remove it here, mind removing the prop for theme as well?

Issue #36, opened 2018-11-27
Private repositories - working?


crudolf:
When I do 

/github subscribe crudolf/XYZ

where XYZ is a private repository. The plugin tells me, that the repository cannot be found "Unknown repository .."

Can this plugin only handle public repositories?

jwilander:
It currently only works with public repositories. Some permission checking would need to be added and an update of the OAuth scopes to allow for that

jwilander:
If someone wants to take a stab at this, that would be very welcome. I can help answer questions and point in the right direction

PR #40, opened 2019-01-08
Enable private repositories

zetaab:
default behaviour is disabled. However, now it is possible to allow users to use private repositories as well.

fixes issue #36

### DIFF
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"context"
 	"crypto/hmac"
 	"crypto/sha1"
 	"encoding/hex"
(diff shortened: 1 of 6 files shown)


zetaab:
@jwilander could you review it. At least for us it works very well in github enterprise

jwilander:
The only other cases I can think of is if you created a PR or issue in a repo then got removed and someone performed an action on it (comment, close, reopen, etc.). Looking at the possible notifications that could happen in most cases except being requested for a review or assigned (since you need to be in the repo for those). Might be safer to just check permissions in all cases for private repos

jwilander:
Those changes look good! There's actually one other case I've thought of. Right now if you `/github subscribe somerepo` in a channel there's no check to see if you have permissions to `somerepo`. We'd likely need a permission check when using that command.

There's another concern I have with subscribing a channel to a private repository. If I have access to a repo and subscribe a channel to it and then I am removed from the private repo, my channel is still going get those events. I'm not entirely certain what the best way to handle this is. One way would be to track the creator of a subscription and before we post to the channel make sure that creator still has access to the repo. The downside there is if that user set up the subscription for a channel many users use, when that one user is removed the subscription stops working and the channel no longer gets the posts

Thoughts @zetaab or @lieut-data ?

zetaab:
@jwilander if you do `/github subscribe somerepo` and you do not have access to repo - it will return "Unknown repository" like mentioned in issue #36. https://github.com/mattermost/mattermost-plugin-github/blob/master/server/subscriptions.go#L82-L86  So in my opinion there is no need to add permission check. 

Your another concern is true. However, I think that if user is so smart/stupid that he will add subscription from private repository to channel - he must know what he is doing.

jwilander:
Ah right, I forgot about that for the first one.

For the second one, my worry is that the user could do it maliciously to maintain getting events about a repository that they were removed from

jwilander:
@zetaab to address my second concern I think we need to:
1. Track who created the subscription
2. When a subscription event happens, check If the repo is private and then check if that the creator still has permission to the repo
3. If the user does not have permission, remove the subscription and post a message into the channel saying `@<username> does not have permission to <org/repo>. Subscription removed. Please run /github subscribe <org/repo> to recreate the subscription.`

lieut-data:
Over to @levb and @crspeller given their ownership of all things plugins related now :)

levb:
@zetaab I'd like to better familiarize myself github plugin code, will comment soon.

jwilander:
I pushed a commit to handle the case I described. My only remaining worry is that for end users already using the GitHub plugin they will need to manually do `/github disconnect` then reconnect their account to get the correct OAuth scopes needed. Might be a good idea to provide a notification telling existing users they need to do that if they want to use the plugin with private repos

server/webhook.go
```diff
@@ -158,6 +181,10 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
 			continue
 		}
 
+		if isPrivate && !p.permissionToRepo(sub.CreatorID, repoName) {
```
crspeller: Is it possible to centeralize these checks? They all do similar things but are all spread out. Is it possible to just do the check in `handleWebhook`?

jwilander: Possibly with a significant amount of refactoring

cpitclaudel:
This would be a great feature to have. Thanks everyone for working on it.

server/webhook.go
jwilander: @crspeller pushed some refactoring that should clean it up a bit

jwilander:
@crspeller pushed some refactoring that should clean it up a bit

zetaab:
I cannot approve my own PR but looks great for me. Hopefully someone has tested it :)

jwilander:
I tested it fairly thoroughly and we already have a volunteer to test a build of it before we make an official release :)

levb:
I am the least qualified in the code base, should I hit the merge button?

jwilander:
We're good to merge. Before I cut a new release I want to add the message to end users letting them know they need to reconnect their accounts to get private repos to work but I will do that in a new PR.

cpitclaudel:
Great work!

PR #41, opened 2019-01-09
implement subscribe list

### LINKED ISSUE #37: [Help Wanted] Add a command to list current channel's subscriptions
Add a command `/github subscribe list` that will list all the repository subscriptions of the current channel as an ephemeral message. It should also list the events that are subscribed to (e.g pulls,issues,labels etc.)

If you have questions feel free to ask here or join [the GitHub Plugin channel on our community Mattermost server](https://community.mattermost.com/core/channels/github-plugin).

zetaab:
fixes #37

### DIFF
--- a/server/subscriptions.go
+++ b/server/subscriptions.go
@@ -16,14 +16,10 @@ const (
 	SUBSCRIPTIONS_KEY = "subscriptions"
 )
 
-type SubWithRepo struct {
+type Subscription struct {
+	ChannelID  string
 	Features   string
 	Repository string
-}
-
-type Subscription struct {
-	ChannelID string
-	Features  string
 }
 
 type Subscriptions struct {
@@ -92,8 +88,9 @@ func (p *Plugin) Subscribe(ctx context.Context, githubClient *github.Client, use
 	}
 
 	sub := &Subscription{
-		ChannelID: channelID,
-		Features:  features,
+		ChannelID:  channelID,
+		Features:   features,
+		Repository: fmt.Sprintf("%s/%s", owner, repo),
 	}
 
 	if err := p.AddSubscription(fmt.Sprintf("%s/%s", owner, repo), sub); err != nil {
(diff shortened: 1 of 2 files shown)


server/subscriptions.go
```diff
@@ -98,6 +103,25 @@ func (p *Plugin) Subscribe(ctx context.Context, githubClient *github.Client, use
 	return nil
 }
 
+func (p *Plugin) SubscribeList(ctx context.Context, channelID string) ([]*SubWithRepo, error) {
```
jwilander: We don't actually use the context here so it can probably be removed

server/subscriptions.go
```diff
@@ -16,6 +16,11 @@ const (
 	SUBSCRIPTIONS_KEY = "subscriptions"
 )
 
+type SubWithRepo struct {
```
lieut-data: 0/5 on this one, but `SubWithRepo` seems like a redundant struct just to be able to access the repository for the subscription in question. I wonder if we could solve this more holistically by storing the repository in the existing `Subscription` struct (in addition to the repository being part of the data structure)? I could see cases where I want access to all the data.

lieut-data:
@zetaab, some solid improvements! A few suggestions above along with some comments for discussion :)

server/subscriptions.go
zetaab: also one thing that came to my mind: Is it so that currently it is not possible to subscribe to same repository in two different channels (because its map with key = repo)? In my opinion that is not perfect solution

zetaab: @lieut-data so you think that we could leave `Subscriptions` as is but add repo information also to `Subscription` struct? Then repo information is basically in two places, in map key and in subscription struct. However, at least I do not see that as big problem?

zetaab:
@lieut-data is this now fixed like should?

server/subscriptions.go
```diff
@@ -98,6 +100,28 @@ func (p *Plugin) Subscribe(ctx context.Context, githubClient *github.Client, use
...
+	var filteredSubs []*Subscription
+	subs, err := p.GetSubscriptions()
+	if err != nil {
+		return nil, err
+	}
+
+	for repo, v := range subs.Repositories {
+		for _, s := range v {
+			if s.ChannelID == channelID {
+				// this is needed to be backwards compatible
```
lieut-data: 0/5 on whether this matters, but it might make sense to do this when we load the subscriptions from the key value store (and when we add new ones, etc.), vs. special casing it here. It may not matter in the short term.

lieut-data:
Thanks, @zetaab -- will wait for @jwilander's feedback, and then we can get this merged.

server/subscriptions.go
jwilander: I would agree but I think this is good enough for now

jwilander:
Thanks @zetaab ! Great work on this

PR #43, opened 2019-02-01
MM-13841 Fetch GitHub user on profile popover open

jwilander:
Switched from populating the user's GitHub username from props to a dedicated fetch in the profile popover.

There's no tests because none of the framework for testing the server API or clients has been set up for this plugin.

https://mattermost.atlassian.net/browse/MM-13841

### DIFF
--- a/webapp/src/actions/index.js
+++ b/webapp/src/actions/index.js
@@ -151,3 +151,44 @@ export function getUnreads() {
         return {data};
     };
 }
+
+const GITHUB_USER_GET_TIMEOUT_MILLISECONDS = 1000 * 60 * 60; // 1 hour
+
+export function getGitHubUser(userID) {
+    return async (dispatch, getState) => {
+        if (!userID) {
+            return {};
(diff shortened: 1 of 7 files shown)


webapp/src/actions/index.js
```diff
@@ -151,3 +151,44 @@ export function getUnreads() {
...
+const GITHUB_USER_GET_TIMEOUT_MILLISECONDS = 1000 * 60 * 60; // 1 hour
+
+export function getGitHubUser(userID) {
+    return async (dispatch, getState) => {
+        if (!userID) {
+            return {};
+        }
+
+        const user = getState()['plugins-github'].githubUsers[userID];
+        if (user && user.last_try && Date.now() - user.last_try < GITHUB_USER_GET_TIMEOUT_MILLISECONDS) {
```
jwilander: Some basic backoff logic so we don't spam the plugin every time the popover is re-opened for an unconnected user.

PR #50, opened 2019-02-22
Serve the profile image for the plugin from the plugin itself

### LINKED ISSUE #46: Plugin user profile image is missing
I have connect the plugin to github, but I found the user profile image can't load correct.
image
The image link is `https://mattermost-server/api/v4/image?url=https%3A%2F%2Fassets-cdn.github.com%2Fimages%2Fmodules%2Flogos_page%2FGitHub-Mark.png`, it return 403 error.
And I found [GitHub-Mark.png](https://github.com/mattermost/mattermost-plugin-github/blob/2dbc427e6e6a8cbd3b381a769040186b9ea07ad6/server/api.go#L22) is 404, can the plugin use my own profile image?

jwilander:
This is the long term fix for #46 but unfortunately it requires some changes within the Mattermost server to work properly. The issue is that the Mattermost client and server use an image proxy for profile pictures that rejects URLs using the localhost or the site URL for security reasons.

@hmhealey is going to bring this up with the Mattermost security team to discuss ways around this so that plugins can serve images used by the Mattermost clients.

Long term fixes #46


hanzei:
> he issue is that the Mattermost client and server use an image proxy for profile pictures that rejects URLs using the localhost or the site URL for security reasons.

Could you elaborate a bit? If you use `$SITEURL/plugins/github/assets/profile.png` this should work fine. This is how Matterpoll solves this problem.

The only problem is that this approach doesn't work if `$SITEURL` is `localhost`. See https://community.mattermost.com/core/pl/jwbhjkqyi7ny3cu9jh6zkr4sde for a discussion about this.

jwilander:
Interesting, I was talking with @hmhealey and brought up doing that but he seemed to think that wouldn't work either. If it's working for Matterpoll then it should work here

jwilander:
I will try doing that

hanzei:
Please be aware of this bug: https://mattermost.atlassian.net/browse/MM-14314

You can find the Matterpoll code [here](https://github.com/matterpoll/matterpoll/blob/master/server/plugin/api.go#L63-L76).

jwilander:
Updated to use the SiteURL instead of a relative URL and tested to confirm it works when the SiteURL is not localhost.

server/configuration.go
```diff
@@ -96,6 +100,14 @@ func (p *Plugin) setConfiguration(configuration *configuration) {
 		panic("setConfiguration called with the existing configuration")
 	}
 
+	// PluginDirectory & ProfileImageURL should be set based on server configuration and not the plugin configuration
+	if serverConfiguration.PluginSettings.Directory != nil {
+		configuration.PluginsDirectory = *serverConfiguration.PluginSettings.Directory
+	}
+	if serverConfiguration.ServiceSettings.SiteURL != nil {
+		configuration.ProfileImageURL = path.Join(*serverConfiguration.ServiceSettings.SiteURL, "/plugins/github/assets/profile.png")
```
hanzei: You could make this even for flexible, if you use something like in https://github.com/mattermost/mattermost-plugin-github/blob/68f61d01f1f4d23a95562cdb6075b90aa556fc57/server/api.go#L217.

Issue #44, opened 2019-02-12
How can I susbcribe to a whole organisation?


cpitclaudel:
Hi!

Is there a way to subscribe a channel to all repos in a Github organization? I only found `/github subscribe org/repo`, which requires one command per repository (and does not add new repositories).

Thanks!

jwilander:
Hi @cpitclaudel, currently you need manually subscribe to each repository if you wanted to do this. I'm not opposed to adding a feature for this though. I've added the help wanted label in case someone wants to work on it

benschuster788:
I'll take this one

jasonblais:
If anyone is interested working on this ticket:

1. First PR merged here: https://github.com/mattermost/mattermost-plugin-github/pull/56
2. Second PR was closed, but not merged, here: https://github.com/mattermost/mattermost-plugin-github/pull/52

Moreover, a PR to unsubscribe from a PR was closed but not merged, here: https://github.com/mattermost/mattermost-plugin-github/pull/53

hanzei:
Submitted a PR to fix 2.

dhadiseputro:
I'll work on this, from where benschuster788 left off. I am not familiar yet with the code base and this can be my 'hello world' project if you guys allow. Am I good to take his code and add the fixes need in #53 and #56 ?

jasonblais:
@dhadiseputro Go for it! If you have any questions working through it let us know. You're also open to sharing the approach you're planning to take for review 👍

dhadiseputro:
I am planning to do something similar to benschuster.
The high level idea is, in `SubscribeOrg` I'll get all the repos in that org and do `AddSubscription` for each repo.

I have a question:
- If we pass `repo=""` to `Subscribe`, we will get here https://github.com/mattermost/mattermost-plugin-github/blob/master/server/subscriptions.go#L84, but then `ghOrg` is not getting use anywhere after that. Also, `repo` will still be empty, how can we continue to do the `AddSubscription` at the end of the method?

jasonblais:
pinging @crspeller or @aaronrothschild on the above

For `channelID` - that refers to the ID of a Mattermost channel. To find the ID of a channel, click the dropdown next to the channel name, then select "View Info". The ID is displayed at the bottom of the resulting dialog

![image](https://user-images.githubusercontent.com/13119842/62048039-d91bad00-b1d9-11e9-99c9-abc22737d7a2.png)
[...]

PR #52, opened 2019-02-27
GitHub Subscribe Organization

benschuster788:
This is a fix for issue #44 subscribing to an entire Github organization at once.  Subscriptions can occur by providing the full url for the organization or just the organization name.

`/github subscribe org mattermost`
`/github subscribe org https://github.com/mattermost`

The logic for subscribing consists of using Github's built in `ListByOrg` method to access all of the available repositories in an organization.  It then loops through each and subscribes using the previously built `AddSubscription` method.


benschuster788:
@jwilander

hanzei:
I would like to keep the organisation as the first of the two parameters of `/github subscribe` for consistency. What do you think of using just `/github subscribe mattermost`?

server/utils.go
```diff
@@ -103,6 +103,20 @@ func decrypt(key []byte, text string) (string, error) {
 	return string(unpadMsg), nil
 }
 
+func parseOwner(full, baseURL string) (string, string) {
```
levb: 1/5 would prefer `parseOrg[anization]` as a more intention-revealing name.

server/utils.go
```diff
@@ -103,6 +103,20 @@ func decrypt(key []byte, text string) (string, error) {
...
+	if baseURL == "" {
+		baseURL = "https://github.com/"
+	}
+	full = strings.TrimSuffix(strings.TrimSpace(strings.Replace(full, baseURL, "", 1)), "/")
```
levb: 0/5 use `url.Parse` and `Path.Split` - more idiomatic?

benschuster788:
Thanks for the comments all. I wasn't sure how you guys wanted the actual syntax for subscribing to an org.  `/github subscribe [ORG_NAME]` makes a little more sense so I'll update the code to reflect that.

server/utils.go
benschuster788: @levb definitely agree with you here

server/utils.go
benschuster788: @levb this was how it was already being used in parseOwnerAndRepo so just did the same.  Using the url package probably makes the most sense in this case though

levb:
@benschuster788 0/5 - I'd just parse the `git subscribe`'s 1st arg as a [URL/]Path, then if there's only 1 element in the path assume it's the org. Seems unambiguous to me.

benschuster788:
@levb yep that's what I was going to do

benschuster788:
@levb @jasonblais just pushed some updates for this.  Instead of using the new parseOwner method I previously created I modified the parseOwnerAndRepo to handle the case where only an organization is provided.  Then on the command level if it is only an org then it subscribes to the org.

I went ahead and added test coverage for the same method in utils_test.go and updated the `COMMAND_HELP` definition.

I didn't make any changes to the url replacing in that method because the way it is setup is that a full url could be provided like `https://github.com/mattermost` or just the org name `mattermost`.  If we were to do a parse URL while appending the baseURL it could fail despite providing a valid org url (https://github.com/https://github.com/mattermost for examples).  If we want to use url.Parse I can rewrite the method additionally to check if the `full` value is a url already.

server/command.go
```diff
@@ -16,6 +16,7 @@ const COMMAND_HELP = `* |/github connect| - Connect your Mattermost account to y
 * |/github disconnect| - Disconnect your Mattermost account from your GitHub account
 * |/github todo| - Get a list of unread messages and pull requests awaiting your review
 * |/github subscribe list| - Will list the current channel subscriptions
+* |/github subscribe owner [features]| - Subscribe the current channel to all available repositories within organziation and receive notifications about opened pull requests and issues
```
jasonblais: `organziation` > `organization`

benschuster788: Just fixed!

server/subscriptions.go
```diff
@@ -102,6 +98,41 @@ func (p *Plugin) Subscribe(ctx context.Context, githubClient *github.Client, use
...
+		return fmt.Errorf("Invalid organization")
+	}
+	if err := p.checkOrg(org); err != nil {
+		return err
+	}
+
+	listOrgOptions := github.RepositoryListByOrgOptions{
+		Type: "all",
+	}
+	repos, _, err := githubClient.Repositories.ListByOrg(ctx, org, &listOrgOptions)
```
hanzei: The github provides Pages results by default. Hence this will only fetch the first 100 (I think) repositories. See https://github.com/google/go-github#pagination for more infos.

server/subscriptions.go
```diff
@@ -102,6 +98,41 @@ func (p *Plugin) Subscribe(ctx context.Context, githubClient *github.Client, use
...
+
+	for _, repo := range repos {
+		sub := &Subscription{
+			ChannelID:  channelID,
+			CreatorID:  userId,
+			Features:   features,
+			Repository: fmt.Sprintf("%s/%s", org, repo.GetFullName()),
+		}
+
+		if err := p.AddSubscription(fmt.Sprintf("%s/%s", org, repo), sub); err != nil {
```
hanzei: Maybe log the error?

benschuster788: That’s a good idea, I’ll add that in

levb:
@benschuster788 When merging the ticket I accidentally overlooked that there were 2 comments from @hanzei still unresolved. Do you mind fixing those? Apologies for the confusion,

benschuster788:
@levb @hanzei just created a new PR #56 with a commit that paginates through the repos in a organization since the limit returned by Github is just 30 repos per response.

PR #53, opened 2019-02-27
GitHub Unsubscribe from Organization

benschuster788:
This is related to issue #44 and gives users to the ability to unsubscribe from an entire organization.  Use case is if a user subscribes to an entire organization and then decides they don't want to.  Instead of unsubscribing from each repo they can unsubscribe from the entire organization.  Use is below

`/github unsubscribe org https://github.com/mattermost`
`/github unsubscribe org mattermost`

This PR is built off of (and includes) the changes in pull request #52.
@jwilander


benschuster788:
Just updated this branch to paginate calls through an organizations repo's because Github only returns 30 repos per request.

server/subscriptions.go
```diff
@@ -276,3 +290,65 @@ func (p *Plugin) Unsubscribe(channelID string, repo string) error {
...
+	listOrgOptions := github.RepositoryListByOrgOptions{
+		Type: "all",
+		ListOptions: github.ListOptions{
+			Page: 1,
+		},
+	}
+
+	orgRepos := map[string]github.Repository{}
+	fetchingRepos := true
+	for fetchingRepos {
```
hanzei: Please see my comment on the other PR: https://github.com/mattermost/mattermost-plugin-github/pull/56#discussion_r262089022

jasonblais:
Hey @benschuster788 - I'll close the PR due to inactivity, but leave the branch open if the work is continued later on :)

PR #57, opened 2019-03-05
make Github icon url configurable

zetaab:
I installed 0.8 version today our mattermost, and now the problem is that Icon is not visible at all. This is because icon url is hardcoded to code, and our mattermost server cannot access to internet.

That is why I suggest that we make this url configurable

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -69,6 +69,12 @@
                 "display_name": "Enable Private Repositories",
                 "type": "bool",
                 "help_text": "(Optional) Allow the plugin to work with private repositories. Enabling private repositories will require existing users to reconnect their accounts to gain access to private repositories. A message will be automatically be sent to affected users next time they load Mattermost alerting them of this."
+            },
+            {
+                "key": "GithubIconURL",
+                "display_name": "Github bot icon URL",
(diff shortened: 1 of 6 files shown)


zetaab:
I have no idea why that test fails

lieut-data:
Is there a way we could have the plugin host the image itself? Or for the client-side code to embed the image into the webpack bundle directly? 

Presumably this doesn't need to actually be "configurable", but just accessible without an internet connection. Or, @zetaab, were you hoping to allow changing it from the default logo?

zetaab:
Yes I hope that i could change the url of the logo. This code works fine for me. I am now hosting that same image inside our internal network and it works fine. This feature also allows changing the bot icon if someone wants customize it

jwilander:
I do have a PR here for self-hosting the image from the plugin https://github.com/mattermost/mattermost-plugin-github/pull/50

However, I'm not opposed to making it configurable

crspeller:
Agree with @jasonblais https://github.com/mattermost/mattermost-plugin-github/pull/50 is the solution to this. Thanks for the contribution, but going to close this.

PR #60, opened 2019-03-10
deploy to GH when tag a release

cpanato:
Adding a functionality to deploy to GH when we tag a release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,24 +3,43 @@ jobs:
   lint:
     docker:
       - image: circleci/golang:1.12-node
-
     working_directory: /go/src/github.com/mattermost/mattermost-plugin-github
     steps:
       - checkout
-
       - run: curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
       - run: make check-style
 
   test:
     docker:
       - image: circleci/golang:1.12-node
-
     working_directory: /go/src/github.com/mattermost/mattermost-plugin-github
     steps:
       - checkout
-
       - run: curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
       - run: make test
+
+  build:
+    docker:
+      - image: circleci/golang:1.12-node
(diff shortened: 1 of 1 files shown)


jwilander:
I'm not sure we want this, I'd much rather have manual control over the releases so that we can fill in the changelog

cpanato:
@jwilander you will need to create the tag,  upload the plugin and after you will need to add the changelog. This PR reduce in one step which will upload the release when you tag.

I can research a bit later in the near future if we can automatize the changelog in the GH release.

jwilander:
I honestly prefer a manual changelog at this point. One based solely off the commit messages is likely to be unclear

cpanato:
I was thinking in a changelog file that you commit and we have some script which parses that :) @jwilander

jwilander:
That would be fine with me 👍

.circleci/config.yml
```diff
@@ -3,28 +3,78 @@ jobs:
...
     steps:
       - checkout
-
       - run: curl https://raw.githubusercontent.com/golang/dep/master/install.sh | sh
       - run: make test
 
+  build:
+    docker:
+      - image: circleci/golang:1.12-node
+    working_directory: /go/src/github.com/mattermost/mattermost-plugin-github
```
hanzei: Might be worth to move this over to a default `executor` like in https://github.com/kaakaa/mmplugin-announce-new-channel/blob/master/.circleci/config.yml#L2-L6. 2/5

hanzei:
I would like to discuss the changelog idea with a broader audience. Would you mind proposing your idea in [~Developers-Toolkit](https://community.mattermost.com/core/channels/developer-toolkit)?

.circleci/config.yml
cpanato: looks great this, will change

cpanato:
@hanzei I can add a post when i have some ideas, for now i dont have any :)

hanzei:
Re Changelog: To be fair, we wont release plugin so often. :+1: for automation in general, but I don't think this is a big topic ATM.

PR #63, opened 2019-03-24
Fix #48: Use IssueCommentEvent's action in post msg

### LINKED ISSUE #48: If a user deletes a comment, you receive a notification for a comment creation
**Description:**
When a user is deleting comments on a PR e.g., you will be notified that the user posted a new comment:

Example

Compared to:
https://github.com/mattermost/mattermost-webapp/pull/2356

The "Comment created" from @jasonblais are actually comment deletes.

**Actual Behavior:**
Comment deletion being shown as comment creation

**Expected Behavior:**
Comment deletion being shown as comment deletion

checkaayush:
Fixes #48

### DIFF
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -818,12 +818,12 @@ func (p *Plugin) handleCommentAuthorNotification(event *github.IssueCommentEvent
 	message := ""
 	switch splitURL[len(splitURL)-2] {
 	case "pull":
-		message = "[%s](%s) commented on your pull request [%s#%v](%s)"
+		message = "[%s](%s) %s comment on your pull request [%s#%v](%s)"
 	case "issues":
-		message = "[%s](%s) commented on your issue [%s#%v](%s)"
+		message = "[%s](%s) %s comment on your issue [%s#%v](%s)"
 	}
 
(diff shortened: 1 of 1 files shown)


checkaayush:
@jwilander this is ready for review.

jwilander:
Hmmm this will work but I wonder if it would be better to just not send a notification for this case? Does anybody really want notifications when someone deletes a comment?

server/webhook.go
```diff
@@ -818,12 +818,12 @@ func (p *Plugin) handleCommentAuthorNotification(event *github.IssueCommentEvent
 	message := ""
 	switch splitURL[len(splitURL)-2] {
 	case "pull":
-		message = "[%s](%s) commented on your pull request [%s#%v](%s)"
+		message = "[%s](%s) %s comment on your pull request [%s#%v](%s)"
```
hanzei: Maybe 
```suggestion
		message = "[%s](%s) %s a comment on your pull request [%s#%v](%s)"
```

Then the message would be e.g.
`someone deleted a comment on your pull request`

hanzei:
I agree. We should just filter out `edited` and `deleted` as `action`'s.

server/webhook.go
hanzei: Let's discuss first, if we want this.

checkaayush:
> I agree. We should just filter out `edited` and `deleted` as `action`'s.

Done with the change. Ready for final review. @hanzei

PR #64, opened 2019-03-25
Fix paging for Subscribe Organization

hanzei:
The old code did only fetched the first page and not more. This PR fixes this.

### DIFF
--- a/server/subscriptions.go
+++ b/server/subscriptions.go
@@ -106,18 +106,24 @@ func (p *Plugin) SubscribeOrg(ctx context.Context, githubClient *github.Client, 
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


server/subscriptions.go
```diff
@@ -106,18 +106,24 @@ func (p *Plugin) SubscribeOrg(ctx context.Context, githubClient *github.Client,
 		return err
 	}
 
-	listOrgOptions := github.RepositoryListByOrgOptions{
-		Type: "all",
+	opt := &github.RepositoryListByOrgOptions{
+		ListOptions: github.ListOptions{PerPage: 50},
```
hanzei: This number is chosen arbitrary. I have no data to back it up.

PR #68, opened 2019-04-09
Point to external profile image and bump version to 0.9.1

jwilander:
See https://community.mattermost.com/core/pl/3ukuuy7o8pngzdgxdtaxzns1fa for context


jwilander:
Sorry @crspeller, didn't answer your question. For some reason the image proxy on community was complaining about it, not really sure why because when I tested it locally it worked fine. I think I'd rather do this for now until we convert to use a proper bot account then just set the profile image of the bot account instead of using the props override

PR #74, opened 2019-05-21
MM-15566 Adding new CI

crspeller:


.circleci/config.yml
```diff
@@ -1,81 +1,58 @@
 version: 2.1
-executors:
-  default:
-    working_directory: /go/src/github.com/mattermost/mattermost-plugin-github
-    docker:
-      - image: circleci/golang:1.12-node
 
+orbs:
+  plugin-ci: mattermost/plugin-ci@volatile
```
hanzei: Out of interest: Is there a repo for this?

go.mod
```diff
@@ -0,0 +1,20 @@
+module github.com/mattermost/mattermost-plugin-github
+
+go 1.12
+
+require (
+	github.com/go-ldap/ldap v3.0.3+incompatible // indirect
+	github.com/google/go-github v17.0.0+incompatible
```
hanzei: Maybe this is a good time to update this? v17.0.0 is a year old.

cpanato: If we update that we need to change some code there are some breaking changes

cpanato: However if need i can do the update

hanzei: I'm 1/5 that this dependency should be updated at some point. 0/5 if this PR is the right place.

.circleci/config.yml
crspeller: Soon.

go.mod
crspeller: Another PR.

.circleci/config.yml
crspeller: https://github.com/mattermost/circleci-orbs

PR #78, opened 2019-05-31
Updating dependencies.

crspeller:


crspeller:
@levb I use tools (I don't edit the files myself) but I fix the problems and make the PRs.

hanzei:
@crspeller You don't use `go mod *`? What do you use instead?

crspeller:
@hanzei For go dependencies I do.

PR #83, opened 2019-06-03
Using a bot account.

crspeller:


server/api.go
```diff
@@ -211,21 +205,6 @@ func (p *Plugin) completeConnectUserToGitHub(w http.ResponseWriter, r *http.Requ
 	w.Write([]byte(html))
 }
 
-func (p *Plugin) handleProfileImage(w http.ResponseWriter, r *http.Request) {
```
hanzei: Removing this will break existing posts. 3/5 that we should not do this. At least until `v1.0`.

server/command.go
```diff
@@ -44,14 +44,13 @@ func getCommand() *model.Command {
...
-		Username:     GITHUB_USERNAME,
-		IconURL:      p.getConfiguration().ProfileImageURL,
-		Type:         model.POST_DEFAULT,
+func (p *Plugin) postCommandResponse(args *model.CommandArgs, text string) {
+	post := &model.Post{
+		UserId:    p.BotUserID,
+		ChannelId: args.ChannelId,
+		Message:   text,
 	}
+	p.API.SendEphemeralPost(args.UserId, post)
```
hanzei: Mind explicitly ignoring the return value? 2/5
```suggestion
	_ = p.API.SendEphemeralPost(args.UserId, post)
```

server/command.go
```diff
@@ -44,14 +44,13 @@ func getCommand() *model.Command {
...
 }
 
 func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*model.CommandResponse, *model.AppError) {
```
hanzei: I would like (2/5) to refactor this into something like:
```suggestion
func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*model.CommandResponse, *model.AppError) {
	p.executeCommand(c *plugin.Context, args *model.CommandArgs)
	return &model.CommandResponse{}, nil
```

and remove the return value from `executeCommand`.

But I'm fine with opening a ticket for this instead of doing this in this PR.

server/command.go
```diff
@@ -44,14 +44,13 @@ func getCommand() *model.Command {
...
 }
 
-func (p *Plugin) getCommandResponse(responseType, text string) *model.CommandResponse {
-	return &model.CommandResponse{
-		ResponseType: responseType,
-		Text:         text,
-		Username:     GITHUB_USERNAME,
-		IconURL:      p.getConfiguration().ProfileImageURL,
-		Type:         model.POST_DEFAULT,
+func (p *Plugin) postCommandResponse(args *model.CommandArgs, text string) {
```
hanzei: Maybe make this a bit more generic: 2/5
```go
func (p *Plugin) sendEphemeralPost(channelID, userID, message string) {
	ephemeralPost := &model.Post{
		ChannelId: channelID,
		UserId:    p.botUserID,
		Message:   message,
	}
	_ = p.API.SendEphemeralPost(userID, ephemeralPost)
}

```

server/plugin.go
```diff
@@ -81,13 +85,27 @@ func (p *Plugin) OnActivate() error {
...
+	if ensureBotError != nil {
+		return errors.Wrap(ensureBotError, "failed to ensure github bot")
+	}
+	p.BotUserID = botId
+
+	bundlePath, err := p.API.GetBundlePath()
 	if err != nil {
-		mlog.Error(err.Error())
-		return fmt.Errorf("Unable to find user with configured username: %v", config.Username)
+		return &model.AppError{Message: err.Error()}
```
hanzei: Why not `wrap` the error? This method returns a "normal" `error`.

server/plugin.go
```diff
@@ -81,13 +85,27 @@ func (p *Plugin) OnActivate() error {
...
 	}
 
-	p.BotUserID = user.Id
+	profileImage, err := ioutil.ReadFile(filepath.Join(bundlePath, "assets", "profile.png"))
+	if err != nil {
+		return &model.AppError{Message: err.Error()}
```
hanzei: Same question.

server/plugin.go
```diff
@@ -81,13 +85,27 @@ func (p *Plugin) OnActivate() error {
...
+	}
+	p.API.SetProfileImage(botId, profileImage)
```
hanzei: Would you please check the return value

server/plugin.go
```diff
@@ -81,13 +85,27 @@ func (p *Plugin) OnActivate() error {
 		return err
 	}
 	p.API.RegisterCommand(getCommand())
-	user, err := p.API.GetUserByUsername(config.Username)
+
+	botId, ensureBotError := p.Helpers.EnsureBot(&model.Bot{
```
hanzei: Nit: Why not call this `appErr`?
```suggestion
	botUserID, appErr := p.Helpers.EnsureBot(&model.Bot{
```

server/webhook.go
```diff
@@ -776,16 +658,16 @@ func (p *Plugin) handleCommentMentionNotification(event *github.IssueCommentEven
 			continue
 		}
 
-		userID := p.getGitHubToUserIDMapping(username)
-		if userID == "" {
+		userId := p.getGitHubToUserIDMapping(username)
```
hanzei: Why was this changes?

webapp/package-lock.json
```diff
@@ -1,1519 +1,1570 @@
 {
   "name": "github",
-  "version": "0.9.1",
+  "version": "0.7.1",
```
hanzei: Are these changes intended?

hanzei:
Would you also please set `"min_server_version": "5.12.0",` in the manifest?

server/api.go
crspeller: It won't because it was not actually used. See: https://github.com/mattermost/mattermost-plugin-github/pull/83/files#diff-97fcb5163c5df39ace07e8d58616214eL108

server/command.go
crspeller: Separate ticket.

server/command.go
crspeller: With above.

server/webhook.go
crspeller: I think `go vet` was glitching out.

webapp/package-lock.json
crspeller: Are package-lock changes ever intended :P
Removed.

crspeller:
@hanzei Ready for re-review.

server/plugin.go
```diff
@@ -81,13 +85,31 @@ func (p *Plugin) OnActivate() error {
 		return err
 	}
 	p.API.RegisterCommand(getCommand())
-	user, err := p.API.GetUserByUsername(config.Username)
+
+	botId, appErr := p.Helpers.EnsureBot(&model.Bot{
+		Username:    "github",
```
lieut-data: There's a non-zero chance someone used to customized this away from `github`. Worth updating the README.md to help guide those upgrading? Ideally they would rename the existing user to github to preserve the "flow" from past message interactions.

crspeller: Ya. I need to upgrade the README to make sure people do the bot account conversion if they have an existing account named `github`. Will mention this.

server/plugin.go
```diff
@@ -81,13 +85,31 @@ func (p *Plugin) OnActivate() error {
 		return err
 	}
 	p.API.RegisterCommand(getCommand())
-	user, err := p.API.GetUserByUsername(config.Username)
+
+	botId, appErr := p.Helpers.EnsureBot(&model.Bot{
```
hanzei: Late catch: This is potential dangerous. `EnsureBot` returns an `error`, not `*model.AppErr`. Because of this the checks later can fail. @crspeller Would you please fix this?

Issue #90, opened 2019-06-05
Do not post a daily reminder if there are no updates


jasonblais:
When there are no updates, the plugin sends a daily reminder with "no updates".

![image](https://user-images.githubusercontent.com/13119842/58991275-e1e19b00-87b5-11e9-84ec-0f6b9e04710a.png)

Propose not sending a daily reminder when there are no updates.

jasonblais:
Thanks @ivenk!

PR #91, opened 2019-06-07
GH-90 Added checking for updates before sending daily update

ivenk:
Issue : #90

Current:
Currently the daily reminder is always send even if there is nothing the user has missed. 

PR: 
Adds a check before sending the daily update. Now there either has to be an unread message, a pull request needing reviewing, an open pull request or a new assignment. 
If nothing is found to report there will be no daily update.

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -296,7 +296,7 @@ func (p *Plugin) getConnected(w http.ResponseWriter, r *http.Request) {
 		resp.GitHubClientID = config.GitHubOAuthClientID
 		resp.Settings = info.Settings
 
-		if info.Settings.DailyReminder && r.URL.Query().Get("reminder") == "true" {
+		if info.Settings.DailyReminder && r.URL.Query().Get("reminder") == "true" && p.HasNews(info) {
 			lastPostAt := info.LastToDoPostAt
 
 			var timezone *time.Location
--- a/server/api.go
+++ b/server/api.go
@@ -296,7 +296,7 @@ func (p *Plugin) getConnected(w http.ResponseWriter, r *http.Request) {
 		resp.GitHubClientID = config.GitHubOAuthClientID
 		resp.Settings = info.Settings
 
-		if info.Settings.DailyReminder && r.URL.Query().Get("reminder") == "true" && p.HasNews(info) {
+		if info.Settings.DailyReminder && r.URL.Query().Get("reminder") == "true" {
 			lastPostAt := info.LastToDoPostAt
 
 			var timezone *time.Location
@@ -308,9 +308,11 @@ func (p *Plugin) getConnected(w http.ResponseWriter, r *http.Request) {
 			nt := time.Unix(now/1000, 0).In(timezone)
 			lt := time.Unix(lastPostAt/1000, 0).In(timezone)
 			if nt.Sub(lt).Hours() >= 1 && (nt.Day() != lt.Day() || nt.Month() != lt.Month() || nt.Year() != lt.Year()) {
-				p.PostToDo(info)
-				info.LastToDoPostAt = now
-				p.storeGitHubUserInfo(info)
+				if p.HasUnreads(info) {
+					p.PostToDo(info)
+					info.LastToDoPostAt = now
+					p.storeGitHubUserInfo(info)
+				}
 			}
 		}
 
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -367,6 +367,43 @@ func (p *Plugin) GetToDo(ctx context.Context, username string, githubClient *git
 	return text, nil
 }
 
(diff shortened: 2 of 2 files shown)


ivenk:
Hi,
i took a swing at this. Tell me if there is anything i can change to improve it.

jasonblais:
Great, thanks @ivenk! Queuing for review

server/api.go
```diff
@@ -296,7 +296,7 @@ func (p *Plugin) getConnected(w http.ResponseWriter, r *http.Request) {
 		resp.GitHubClientID = config.GitHubOAuthClientID
 		resp.Settings = info.Settings
 
-		if info.Settings.DailyReminder && r.URL.Query().Get("reminder") == "true" {
+		if info.Settings.DailyReminder && r.URL.Query().Get("reminder") == "true" && p.HasNews(info) {
```
jwilander: Might be better to do this inside this if block after doing the time comparison since it will be a little heavy making network requests

server/plugin.go
```diff
@@ -367,6 +367,43 @@ func (p *Plugin) GetToDo(ctx context.Context, username string, githubClient *git
 	return text, nil
 }
 
+func (p *Plugin) HasNews(info *GitHubUserInfo) bool {
```
jwilander: How `HasUnreads` to be a little more consistent/clear?

server/plugin.go
```diff
@@ -367,6 +367,43 @@ func (p *Plugin) GetToDo(ctx context.Context, username string, githubClient *git
...
+	githubClient := p.githubConnect(*info.Token)
+	config := p.getConfiguration()
+
+	issues, _, err := githubClient.Search.Issues(ctx, getReviewSearchQuery(username, config.GitHubOrg), &github.SearchOptions{})
+	if err != nil {
+		mlog.Error(err.Error())
+		return false
+	}
+
+	notifications, _, err := githubClient.Activity.ListNotifications(ctx, &github.NotificationListOptions{})
```
jwilander: We actually don't consider all the GitHub notifications as notifications in the plugin since some of them are super spammy. So you'll need to a bit of post-processing here to filter out the ones we don't care about. See https://github.com/mattermost/mattermost-plugin-github/blob/master/server/plugin.go#L298 for what we skip/ignore

server/plugin.go
ivenk: Sounds good. I will change that.

server/plugin.go
ivenk: Thanks! I will read up on that.

server/api.go
ivenk: Will do.

server/plugin.go
```diff
@@ -367,6 +367,43 @@ func (p *Plugin) GetToDo(ctx context.Context, username string, githubClient *git
...
+	if err != nil {
+		mlog.Error(err.Error())
+		return false
+	}
+
+	yourPrs, _, err := githubClient.Search.Issues(ctx, getYourPrsSearchQuery(username, config.GitHubOrg), &github.SearchOptions{})
+	if err != nil {
+		mlog.Error(err.Error())
+		return false
```
crspeller: Why return false on error? Why not just post anyway?

server/plugin.go
```diff
@@ -367,6 +367,43 @@ func (p *Plugin) GetToDo(ctx context.Context, username string, githubClient *git
...
+	}
+
+	yourAssignments, _, err := githubClient.Search.Issues(ctx, getYourAssigneeSearchQuery(username, config.GitHubOrg), &github.SearchOptions{})
+	if err != nil {
+		mlog.Error(err.Error())
+		return false
+	}
+
+	if issues.GetTotal() == 0 && len(notifications) == 0 && yourPrs.GetTotal() == 0 && yourAssignments.GetTotal() == 0 {
```
crspeller: Would it make sense to only check the notifications? That way you only get the message if something has changed.
At least for me, this will never be false.

@jwilander Thoughts?

server/plugin.go
ivenk: I will update that.

server/plugin.go
jwilander: Yes I was thinking that would be better originally but I was wondering if some people would still want the reminders even if there are no new unreads. e.g. if they read a PR but didn't review it because they didn't have time. That would make it less useful to me and you, though, and I think we are representative of the core users of the

@ivenk or @jasonblais any thoughts?

ivenk: I personally would appreciate reminders even if there are no new unreads. However i see why one might disagree.

crspeller: Ok. I am fine with leaving it as-is.

jwilander:
Much thanks @ivenk !

PR #96, opened 2019-06-17
README update: format setting config for better usability

jasonblais:
Two people today had issues with this README 
 - (while screensharing) one tried to copy-paste the provided webhook and OAuth URLs, but had trouble because they are formatted as links with quotation marks. Updated to use in-line code formatting instead
 - both missed "Content Type" value in configuring webhook
 - one missed the correct authorization callback URL

PM review: @aaronrothschild tagging you for review, so that you can help review the proposed changes


jasonblais:
I forgot to add dev review, cc @jwilander / @hanzei if you habe any feedback, I can update on another PR

Issue #94, opened 2019-06-17
Include a redirect_uri in the authorize request


ThiefMaster:
Currently this is not the case, so I need to set the correct URI on the GitHub app registration. But this is a bad idea because without looking at the code I have no way of knowing what the exact redirect URI needs to be (`https://.../plugins/github/oauth/complete`)!

jasonblais:
Thanks @ThiefMaster - what happens when it's not set. Also, where do you configure it when registering the GitHub app?

![image](https://user-images.githubusercontent.com/13119842/59603238-e754b480-90d6-11e9-92d1-2ec6ad5f0b61.png)

ThiefMaster:
The callback URL is used if no URL is provided. This works OK but the advantage of the plugin sending the redirect_uri is that specifying the base url (e.g. `https://mattermost.example..com`) for the callback URL is on Github enough, since (as mentioned in the oauth docs linked there) it allows any subpath within what's specified to be used as the redirect_uri

jasonblais:
@ThiefMaster So we do have the `https://.../plugins/github/oauth/complete` URL in our doc, attached below. Wondering if in the short term it would be helpful if this URL was also in the System Console settings page where the GitHub plugin is configured? This would at least unblock the initial configuration.
 
![image](https://user-images.githubusercontent.com/13119842/59643619-61b52100-9137-11e9-9605-d396734fc35d.png)

Then in the long term we could have a redirect_uri included in the authorize request of the plugin, so that someone could just configure their Mattermost site URL as the callback URL. Not sure if there are any security or functional implications of this change, though.

Let me know what you think, and whether I understood your case correctly?

(We could also do a better job of linking back to this README doc for configuration steps directly from the System Console settings page, or alternatively add some of the instructions there.)

ThiefMaster:
Yes, I saw it in the github readme only later when I noticed it didn't work without the exact URL configured on GitHub. So I think having the URL showed in the system console would indeed be useful.

> Not sure if there are any security or functional implications of this change, though.

I think Github is actually the exception where you don't *have to* specify it.

jasonblais:
PR submitted for a proposed system console text update as an immediate improvement, let me know what you think https://github.com/mattermost/mattermost-plugin-github/pull/99

hanzei:
Is this issue resolved with #99?

aaronrothschild:
[...]

PR #99, opened 2019-06-19
Update GitHub plugin introduction with authorization callback URL

jasonblais:
In response to https://github.com/mattermost/mattermost-plugin-github/issues/94

@aaronrothschild would be curious to hear your thoughts on other options, e.g. we could link to docs from the Introduction page, or incorporate some of the steps directly in there.

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -15,7 +15,7 @@
         "bundle_path": "webapp/dist/main.js"
     },
     "settings_schema": {
-        "header": "To set up the GitHub plugin, you need to register a GitHub OAuth app here https://github.com/settings/applications/new.",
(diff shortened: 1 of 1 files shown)


aaronrothschild:
I think we should definitely add more context within the integration area as is reasonable AND we should link to the documentation that outlines this in a bit more detail.  I agree that having a value you can simply cut/paste is less prone to error and is generally useful as a UX for other providers in the future too.

jasonblais:
> I think we should definitely add more context within the integration area as is reasonable

Thanks @aaronrothschild - do you mean we should add more context in addition to the changes in this PR?

aaronrothschild:
Context such as what else they may need to do when installing.   I think we should have links to the instruction at the top of each plugin section.  

I'll change the text to also include a link to https://github.com/mattermost/mattermost-plugin-github/blob/master/README.md

PR #100, opened 2019-06-21
Post response if unkown action is provided

hanzei:


server/command.go
```diff
@@ -229,5 +229,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 		return &model.CommandResponse{}, nil
 	}
 
+	p.postCommandResponse(args, fmt.Sprintf("Unknown action %v", action))
```
jasonblais: Future improvement may be to return the output of `/github help`, so the user can immediately see what commands are available

hanzei: 0/5 that just printing "Unknown action" is enough here. Your call.

jasonblais: Yes, my note was intended as a future improvement, this PR alone is an improvement (it's better to return a message than no message at all)

hanzei:
@jwilander Gentle reminder to review this

jwilander:
Thanks for the reminder!

PR #110, opened 2019-07-22
Use new RegisterSlashCommand helper

irbrad:
This PR uses new helper commands implemented in mattermost/mattermost-server#11678.

From the commit against mattermost-server:
- adds SplitCommandArguments helper that takes CommandArgs and splits them into a command (string), action (string), and parameters ([]string).
- adds RegisterSlashCommand helper that matches a given slash command trigger with a callback function.

Jira issue: https://mattermost.atlassian.net/browse/MM-16801


irbrad:
cc @lieut-data

PR #122, opened 2019-09-11
replacing apache.com/thrift with github.com to solve network issue when building

niklabh:
## Summary
go mudule unable to access git.apache.org as it is timing out. Replacing it with github.com for thrift.

## Issue

https://github.com/mattermost/mattermost-server/issues/12125

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,4 +31,7 @@ require (
 )
 
 // Workaround for https://github.com/golang/go/issues/30831 and fallout.
-replace github.com/golang/lint => github.com/golang/lint v0.0.0-20190227174305-8f45f776aaf1
+replace (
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -31,4 +31,7 @@ require (
 )
 
 // Workaround for https://github.com/golang/go/issues/30831 and fallout.
-replace github.com/golang/lint => github.com/golang/lint v0.0.0-20190227174305-8f45f776aaf1
+replace (
+	git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
+	github.com/golang/lint => github.com/golang/lint v0.0.0-20190227174305-8f45f776aaf1
```
hanzei: Would you please move the comment from L33 before this?

niklabh: done

PR #137, opened 2019-10-07
MM-8880: Show code block previews of Github permalinks

### LINKED ISSUE #128: Show code block preview of Github code line permalink
When a user post contains a Github code line permalink, it should attach a markdown code block that will contain the preview of that section of the code with possibly 3 lines before and after (much like the display shown in code compares in Github:

Screen Shot 2019-09-28 at 5 03 43 PM

The [Github plugin](https://github.com/mattermost/mattermost-plugin-github) should subscribe to the [`MessageWillBePosted`](https://github.com/mattermost/mattermost-server/blob/cee19b0332d66851cd7d9a16ea9984d8275a747d/plugin/hooks.go#L104|https://github.com/mattermost/mattermost-server/blob/cee19b0332d66851cd7d9a16ea9984d8275a747d/plugin/hooks.go#L104) hook and also include a setting to enable/disable this feature. 

----

JIRA: https://mattermost.atlassian.net/browse/MM-18880

agnivade:
#### Summary

For every message we now hook into the MessageWillBePosted event, and check
if it contains a permalink to a Github file.

If yes, we parse out the link, and make an API call to the Github contents API.
Get the file contents and filter out the required lines given in the permalink anchor.

Then we construct a code block in markdown and replace the original string.

Added a config key to gate this feature.

Screenshots:

With config disabled:
before-mm

After enabling it:
after-mm
The newly formed hyperlink is a link to the original permalink

Config setting:
codepreview

#### Ticket Link

[MM-18880](https://mattermost.atlassian.net/browse/MM-18880)

Fixes #128

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ endif
 govet:
 ifneq ($(HAS_SERVER),)
 	@echo Running govet
-	@# Workaroung because you can't install binaries without adding them to go.mod
+	@# Workaround because you can't install binaries without adding them to go.mod
 	env GO111MODULE=off $(GO) get golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow
 	$(GO) vet ./server/...
 	$(GO) vet -vettool=$(GOPATH)/bin/shadow ./server/...
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ endif
 govet:
 ifneq ($(HAS_SERVER),)
 	@echo Running govet
-	@# Workaroung because you can't install binaries without adding them to go.mod 
+	@# Workaroung because you can't install binaries without adding them to go.mod
 	env GO111MODULE=off $(GO) get golang.org/x/tools/go/analysis/passes/shadow/cmd/shadow
 	$(GO) vet ./server/...
 	$(GO) vet -vettool=$(GOPATH)/bin/shadow ./server/...
@@ -154,7 +154,7 @@ endif
 
 ## Extract strings for translation from the source code.
 .PHONY: i18n-extract
-i18n-extract: 
+i18n-extract:
 ifneq ($(HAS_WEBAPP),)
 	@[[ -d $(MM_UTILITIES_DIR) ]] || echo "You must clone github.com/mattermost/mattermost-utilities repo in .. to use this command"
 	@[[ -d $(MM_UTILITIES_DIR) ]] && cd $(MM_UTILITIES_DIR) && npm install && npm run babel && node mmjstool/build/index.js i18n extract-webapp --webapp-dir ../mattermost-plugin-demo/webapp
--- a/server/permalinks.go
+++ b/server/permalinks.go
@@ -105,12 +105,11 @@ func (p *Plugin) makeReplacements(msg string, replacements []replacement, ghClie
 		reqctx, cancel := context.WithTimeout(context.Background(), permalinkReqTimeout)
 		fileContent, _, _, err := ghClient.Repositories.GetContents(reqctx,
 			r.permalinkInfo.user, r.permalinkInfo.repo, r.permalinkInfo.path, &opts)
+		defer cancel()
 		if err != nil {
 			p.API.LogError("error while fetching file contents", "error", err.Error(), "path", r.permalinkInfo.path)
-			cancel()
 			continue
 		}
-		cancel()
 		// this is not a file, ignore.
 		if fileContent == nil {
 			p.API.LogWarn("permalink is not a file", "file", r.permalinkInfo.path)
--- a/server/permalinks.go
+++ b/server/permalinks.go
@@ -1,0 +1,119 @@
+package main
+
+import (
+	"context"
+	"encoding/hex"
+	"fmt"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/google/go-github/v25/github"
+)
+
+const maxPermalinkReplacements = 10
+const permalinkReqTimeout = 2 * time.Second
+
+// replacement holds necessary info to replace github permalinks
+// in messages with a code preview block
+type replacement struct {
+	index      int               // index of the permalink in the string
+	word       string            // the permalink
+	captureMap map[string]string // named regex capture group of the link
+}
+
+// getReplacements returns the permalink replacements that needs to be performed
+// on a message.
+func (p *Plugin) getReplacements(msg string) []replacement {
+	// find the permalinks from the msg using a regex
+	matches := p.githubRegex.FindAllStringSubmatch(msg, -1)
+	indices := p.githubRegex.FindAllStringIndex(msg, -1)
+	replacements := make([]replacement, 0, maxPermalinkReplacements)
+	for i, m := range matches {
+		// have a limit on the no. of replacements to do
+		if i > maxPermalinkReplacements {
+			break
+		}
+		word := m[0]
+		index := indices[i][0]
+		r := replacement{
+			index: index,
+			word:  word,
+		}
+		// ignore if the word is inside a link
+		if isInsideLink(msg, index) {
+			continue
+		}
+		// populate the captureMap with the extracted groups of the regex
+		for j, name := range p.githubRegex.SubexpNames() {
+			if j == 0 {
+				r.captureMap = make(map[string]string)
+				continue
+			}
+			r.captureMap[name] = m[j]
+		}
+		replacements = append(replacements, r)
+	}
+	return replacements
+}
+
+// makeReplacements perform the given replacements on the msg and returns
+// the new msg.
+func (p *Plugin) makeReplacements(msg string, replacements []replacement) string {
+	for _, r := range replacements {
+		// quick bailout if the commit hash is not proper
+		_, err := hex.DecodeString(r.captureMap["commit"])
+		if err != nil {
+			p.API.LogError("bad git commit hash", "error", err.Error(), "hash", r.captureMap["commit"])
+			continue
+		}
+
+		// get the file contents
+		opts := github.RepositoryContentGetOptions{
+			Ref: r.captureMap["commit"],
+		}
+		// TODO: make all of these requests concurrently.
+		reqctx, cancel := context.WithTimeout(context.Background(), permalinkReqTimeout)
+		fileContent, _, _, err := p.githubClient.Repositories.GetContents(reqctx,
+			r.captureMap["user"], r.captureMap["repo"], r.captureMap["path"], &opts)
+		if err != nil {
+			p.API.LogError("error while fetching file contents", "error", err.Error(), "path", r.captureMap["path"])
+			cancel()
+			continue
+		}
+		cancel()
+		// this is not a file, ignore.
+		if fileContent == nil {
+			continue
+		}
+		decoded, err := fileContent.GetContent()
+		if err != nil {
+			p.API.LogError("error while decoding file contents", "error", err.Error(), "path", r.captureMap["path"])
+			continue
+		}
+
+		// get the required lines
+		start, end := getLineNumbers(r.captureMap["line"])
+		// bad anchor tag, ignore.
+		if start == -1 || end == -1 {
+			continue
+		}
+		lines := filterLines(decoded, start, end)
+
+		// construct the final string
+		final := fmt.Sprintf("\n[%s/%s/%s](%s)\n",
+			r.captureMap["user"], r.captureMap["repo"], r.captureMap["path"], r.word)
+		ext := path.Ext(r.captureMap["path"])
+		// remove the preceding dot
+		if len(ext) > 1 {
+			ext = ext[1:]
+		}
+		final += "```" + ext + "\n"
+		final += lines
+		final += "```\n"
+
+		// replace word in msg starting from r.index only once
+		msg = msg[:r.index] + strings.Replace(msg[r.index:], r.word, final, 1)
+	}
+	return msg
+}
--- a/server/permalinks.go
+++ b/server/permalinks.go
@@ -68,7 +68,7 @@ func (p *Plugin) getReplacements(msg string) []replacement {
 
 // makeReplacements perform the given replacements on the msg and returns
 // the new msg.
-func (p *Plugin) makeReplacements(msg string, replacements []replacement) string {
+func (p *Plugin) makeReplacements(msg string, replacements []replacement, ghClient *github.Client) string {
 	// iterating the slice in reverse to preserve the replacement indices.
 	for i := len(replacements) - 1; i >= 0; i-- {
 		r := replacements[i]
@@ -84,7 +84,7 @@ func (p *Plugin) makeReplacements(msg string, replacements []replacement) string
 		}
 		// TODO: make all of these requests concurrently.
 		reqctx, cancel := context.WithTimeout(context.Background(), permalinkReqTimeout)
-		fileContent, _, _, err := p.githubClient.Repositories.GetContents(reqctx,
+		fileContent, _, _, err := ghClient.Repositories.GetContents(reqctx,
 			r.captureMap["user"], r.captureMap["repo"], r.captureMap["path"], &opts)
 		if err != nil {
 			p.API.LogError("error while fetching file contents", "error", err.Error(), "path", r.captureMap["path"])
--- a/server/permalinks.go
+++ b/server/permalinks.go
@@ -9,13 +9,13 @@ import (
 	"github.com/google/go-github/v25/github"
 )
 
-// maxPermalinkReplacements sets the maximum limit to the no. of
+// maxPermalinkReplacements sets the maximum limit to the number of
 // permalink replacements that can be performed on a single message.
 const maxPermalinkReplacements = 10
 
 const permalinkReqTimeout = 5 * time.Second
 
-// maxPreviewLines sets the maximum no. of preview lines that will be shown
+// maxPreviewLines sets the maximum number of preview lines that will be shown
 // while replacing a permalink.
 const maxPreviewLines = 10
 
@@ -24,22 +24,29 @@ const maxPreviewLines = 10
 const permalinkLineContext = 3
 
 // replacement holds necessary info to replace github permalinks
-// in messages with a code preview block
+// in messages with a code preview block.
 type replacement struct {
-	index      int               // index of the permalink in the string
-	word       string            // the permalink
-	captureMap map[string]string // named regex capture group of the link
+	index         int      // index of the permalink in the string
+	word          string   // the permalink
+	permalinkInfo struct { // holds the necessary metadata of a permalink
+		haswww string
+		commit string
+		user   string
+		repo   string
+		path   string
+		line   string
+	}
 }
 
 // getReplacements returns the permalink replacements that needs to be performed
-// on a message.
+// on a message. The returned slice is sorted by the index in ascending order.
 func (p *Plugin) getReplacements(msg string) []replacement {
 	// find the permalinks from the msg using a regex
 	matches := p.githubPermalinkRegex.FindAllStringSubmatch(msg, -1)
 	indices := p.githubPermalinkRegex.FindAllStringIndex(msg, -1)
 	var replacements []replacement
 	for i, m := range matches {
-		// have a limit on the no. of replacements to do
+		// have a limit on the number of replacements to do
 		if i > maxPermalinkReplacements {
 			break
 		}
@@ -53,13 +60,25 @@ func (p *Plugin) getReplacements(msg string) []replacement {
 		if isInsideLink(msg, index) {
 			continue
 		}
-		// populate the captureMap with the extracted groups of the regex
-		r.captureMap = make(map[string]string)
+		// populate the permalinkInfo with the extracted groups of the regex
 		for j, name := range p.githubPermalinkRegex.SubexpNames() {
 			if j == 0 {
 				continue
 			}
-			r.captureMap[name] = m[j]
+			switch name {
+			case "haswww":
+				r.permalinkInfo.haswww = m[j]
+			case "user":
+				r.permalinkInfo.user = m[j]
+			case "repo":
+				r.permalinkInfo.repo = m[j]
+			case "commit":
+				r.permalinkInfo.commit = m[j]
+			case "path":
+				r.permalinkInfo.path = m[j]
+			case "line":
+				r.permalinkInfo.line = m[j]
+			}
 		}
 		replacements = append(replacements, r)
 	}
@@ -67,44 +86,44 @@ func (p *Plugin) getReplacements(msg string) []replacement {
 }
 
 // makeReplacements perform the given replacements on the msg and returns
-// the new msg.
+// the new msg. The replacements slice needs to be sorted by the index in ascending order.
 func (p *Plugin) makeReplacements(msg string, replacements []replacement, ghClient *github.Client) string {
 	// iterating the slice in reverse to preserve the replacement indices.
 	for i := len(replacements) - 1; i >= 0; i-- {
 		r := replacements[i]
 		// quick bailout if the commit hash is not proper.
-		if _, err := hex.DecodeString(r.captureMap["commit"]); err != nil {
-			p.API.LogError("bad git commit hash in permalink", "error", err.Error(), "hash", r.captureMap["commit"])
+		if _, err := hex.DecodeString(r.permalinkInfo.commit); err != nil {
+			p.API.LogError("bad git commit hash in permalink", "error", err.Error(), "hash", r.permalinkInfo.commit)
 			continue
 		}
 
 		// get the file contents
 		opts := github.RepositoryContentGetOptions{
-			Ref: r.captureMap["commit"],
+			Ref: r.permalinkInfo.commit,
 		}
 		// TODO: make all of these requests concurrently.
 		reqctx, cancel := context.WithTimeout(context.Background(), permalinkReqTimeout)
 		fileContent, _, _, err := ghClient.Repositories.GetContents(reqctx,
-			r.captureMap["user"], r.captureMap["repo"], r.captureMap["path"], &opts)
+			r.permalinkInfo.user, r.permalinkInfo.repo, r.permalinkInfo.path, &opts)
 		if err != nil {
-			p.API.LogError("error while fetching file contents", "error", err.Error(), "path", r.captureMap["path"])
+			p.API.LogError("error while fetching file contents", "error", err.Error(), "path", r.permalinkInfo.path)
 			cancel()
 			continue
 		}
 		cancel()
 		// this is not a file, ignore.
 		if fileContent == nil {
-			p.API.LogWarn("permalink is not a file", "file", r.captureMap["path"])
+			p.API.LogWarn("permalink is not a file", "file", r.permalinkInfo.path)
 			continue
 		}
 		decoded, err := fileContent.GetContent()
 		if err != nil {
-			p.API.LogError("error while decoding file contents", "error", err.Error(), "path", r.captureMap["path"])
+			p.API.LogError("error while decoding file contents", "error", err.Error(), "path", r.permalinkInfo.path)
 			continue
 		}
 
 		// get the required lines.
-		start, end := getLineNumbers(r.captureMap["line"])
+		start, end := getLineNumbers(r.permalinkInfo.line)
 		// bad anchor tag, ignore.
 		if start == -1 || end == -1 {
 			continue
@@ -116,13 +135,13 @@ func (p *Plugin) makeReplacements(msg string, replacements []replacement, ghClie
 		}
 		lines, err := filterLines(decoded, start, end)
 		if err != nil {
-			p.API.LogError("error while filtering lines", "error", err.Error(), "path", r.captureMap["path"])
+			p.API.LogError("error while filtering lines", "error", err.Error(), "path", r.permalinkInfo.path)
 		}
 		if lines == "" {
-			p.API.LogError("line numbers out of range. Skipping.", "file", r.captureMap["path"], "start", start, "end", end)
+			p.API.LogError("line numbers out of range. Skipping.", "file", r.permalinkInfo.path, "start", start, "end", end)
 			continue
 		}
-		final := getCodeMarkdown(r.captureMap, r.word, lines, isTruncated)
+		final := getCodeMarkdown(r.permalinkInfo.user, r.permalinkInfo.repo, r.permalinkInfo.path, r.word, lines, isTruncated)
 
 		// replace word in msg starting from r.index only once.
 		msg = msg[:r.index] + strings.Replace(msg[r.index:], r.word, final, 1)
--- a/server/permalinks.go
+++ b/server/permalinks.go
@@ -3,16 +3,25 @@ package main
 import (
 	"context"
 	"encoding/hex"
-	"fmt"
-	"path"
 	"strings"
 	"time"
 
 	"github.com/google/go-github/v25/github"
 )
 
+// maxPermalinkReplacements sets the maximum limit to the no. of
+// permalink replacements that can be performed on a single message.
 const maxPermalinkReplacements = 10
-const permalinkReqTimeout = 2 * time.Second
+
+const permalinkReqTimeout = 5 * time.Second
+
+// maxPreviewLines sets the maximum no. of preview lines that will be shown
+// while replacing a permalink.
+const maxPreviewLines = 10
+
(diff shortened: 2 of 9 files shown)


marianunez:
@agnivade Could you include a screenshot of an example of how it looks to the description so we can get a quick PM review? cc @aaronrothschild

agnivade:
Yep, done !

Makefile
```diff
@@ -56,7 +56,7 @@ endif
 govet:
 ifneq ($(HAS_SERVER),)
 	@echo Running govet
-	@# Workaroung because you can't install binaries without adding them to go.mod 
+	@# Workaroung because you can't install binaries without adding them to go.mod
```
ThiefMaster: *Workaround

agnivade: Yes, this was already there. I would rather change it in a separate PR to keep related things in a single PR. The trailing space got removed because it's an editor setting of mine. Probably I should revert that too. But let's wait for the review to be done.

aaronrothschild:
LGTM.  Do we cut off the code at a certain number of characters?

agnivade:
Good point actually. I think number of lines is better than number of characters. How about a max of 10 lines with an ellipsis at end if there are more lines to indicate that lines have been cut off in the preview ?

agnivade:
Yep, absolutely. See https://github.com/mattermost/mattermost-plugin-github/pull/137/files#diff-e5d8a1761b1b05bb4bb6df984d0f1334R82 and it's testcases.

agnivade:
gentle ping @marianunez

marianunez:
@agnivade sorry for the delay. Will be getting to this PR today.

Feel free to re-requets review when necessary given that the PR falls out of our review list and sometimes gets out of our radar.

server/permalinks.go
```diff
@@ -0,0 +1,119 @@
...
+		// ignore if the word is inside a link
+		if isInsideLink(msg, index) {
+			continue
+		}
+		// populate the captureMap with the extracted groups of the regex
+		for j, name := range p.githubRegex.SubexpNames() {
+			if j == 0 {
+				r.captureMap = make(map[string]string)
+				continue
+			}
```
marianunez: Could we move this initialization outside of the for loop?

server/permalinks.go
```diff
@@ -0,0 +1,119 @@
...
+	return replacements
+}
+
+// makeReplacements perform the given replacements on the msg and returns
+// the new msg.
+func (p *Plugin) makeReplacements(msg string, replacements []replacement) string {
+	for _, r := range replacements {
+		// quick bailout if the commit hash is not proper
+		_, err := hex.DecodeString(r.captureMap["commit"])
+		if err != nil {
```
marianunez: Nit: these two can be combined into one line

server/permalinks.go
```diff
@@ -0,0 +1,119 @@
...
+			p.API.LogError("bad git commit hash", "error", err.Error(), "hash", r.captureMap["commit"])
```
marianunez:
```suggestion
			p.API.LogError("bad git commit hash in permalink", "error", err.Error(), "hash", r.captureMap["commit"])
```

server/permalinks.go
```diff
@@ -0,0 +1,119 @@
...
+			r.captureMap["user"], r.captureMap["repo"], r.captureMap["path"], &opts)
+		if err != nil {
+			p.API.LogError("error while fetching file contents", "error", err.Error(), "path", r.captureMap["path"])
+			cancel()
+			continue
+		}
+		cancel()
+		// this is not a file, ignore.
+		if fileContent == nil {
+			continue
```
marianunez: Could we include a warning message here ? It seems this would be a false positive case

server/permalinks.go
```diff
@@ -0,0 +1,119 @@
...
+		final := fmt.Sprintf("\n[%s/%s/%s](%s)\n",
+			r.captureMap["user"], r.captureMap["repo"], r.captureMap["path"], r.word)
+		ext := path.Ext(r.captureMap["path"])
+		// remove the preceding dot
+		if len(ext) > 1 {
+			ext = ext[1:]
+		}
+		final += "```" + ext + "\n"
+		final += lines
+		final += "```\n"
```
marianunez: Would move this to a helper function `getCodeMarkdown` so that is more clear what this is doing

server/utils.go
```diff
@@ -173,3 +175,106 @@ func fixGithubNotificationSubjectURL(url string) string {
...
+		l := getLine(parts[0])
+		if l == -1 {
+			return -1, -1
+		}
+		if l < 3 {
+			return 0, l + 3
+		}
+		return l - 3, l + 3
+	case 2: // a line rage
+		start := getLine(parts[0])
```
marianunez: We should add a setting to limit the amount of lines we want to show at a time like @aaronrothschild suggested

agnivade: Done

server/plugin.go
```diff
@@ -116,6 +126,29 @@ func (p *Plugin) OnActivate() error {
...
+	msg := post.Message
+	info, err := p.getGitHubUserInfo(post.UserId)
+	if err != nil {
+		p.API.LogError("error in getting user info", "error", err)
+		return nil, ""
+	}
+	// TODO: apply the same optimization for /slash commands too.
+	p.initOnce.Do(func() {
+		p.githubClient = p.githubConnect(*info.Token)
+	})
```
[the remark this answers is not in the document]
agnivade: Agree with both points. However, I think doing this completely would require applying this logic for the /slash commands too.

It might be better to remove this optimization from this PR and then later do another PR which stores the `githubClient` at instance creation time, and reuse that for all calls. And also have an `updateClient` method which gets triggered on those user change events. That might be easier to review.

server/permalinks.go
agnivade: Sure, but we still need to `continue` if j == 0. But moving out the `captureMap` initialization is more readable. Done.

server/permalinks.go
agnivade: Done

server/permalinks.go
```diff
@@ -0,0 +1,107 @@
...
+			continue
+		}
+
+		// get the required lines
+		start, end := getLineNumbers(r.captureMap["line"])
+		// bad anchor tag, ignore.
+		if start == -1 || end == -1 {
+			continue
+		}
+		lines := filterLines(decoded, start, end, maxPreviewLines)
```
ali-farooq0: passing and handling `maxPreviewLines` is really confusing to a reader, I would suggest to either let `getLineNumbers` return the final truncated `start` and `end` values OR do `if end>maxPreviewLines {end=maxPreviewLines}` here before calling this function.

server/utils.go
```diff
@@ -173,3 +176,129 @@ func fixGithubNotificationSubjectURL(url string) string {
...
+}
+
+// getCodeMarkdown returns the constructed markdown for a permalink.
+func getCodeMarkdown(captureMap map[string]string, word, lines string, start, end int) string {
+	final := fmt.Sprintf("\n[%s/%s/%s](%s)\n",
+		captureMap["user"], captureMap["repo"], captureMap["path"], word)
+	ext := path.Ext(captureMap["path"])
+	// remove the preceding dot
+	if len(ext) > 1 {
+		ext = ext[1:]
```
ali-farooq0: can we use `strings.TrimPrefix` here to simplify this

server/utils.go
```diff
@@ -173,3 +176,129 @@ func fixGithubNotificationSubjectURL(url string) string {
...
+	}
+	final += "```" + ext + "\n"
+	final += lines
+	if end-start > maxPreviewLines { // add an ellipsis if lines were cut off
+		final += "...\n"
```
ali-farooq0: nice 👍

server/permalinks.go
ali-farooq0: Ah i see that you need the origEnd to figure out if the lines were truncated. In that case maybe we can have a `truncatedEnd`? which is passed to the `filterLines`

server/utils.go
```diff
@@ -173,3 +176,129 @@ func fixGithubNotificationSubjectURL(url string) string {
...
 	return fmt.Sprintf("%s/%s", owner, repo)
 }
+
+// filter lines in a string from start to end
+func filterLines(s string, start, end, limit int) string {
+	scanner := bufio.NewScanner(strings.NewReader(s))
+	var buf strings.Builder
+	i := 1
+	w := 1
+	for scanner.Scan() {
```
ali-farooq0: can you merge `i` into this for loop since we don't need `w` to track the lines

server/permalinks.go
```diff
@@ -0,0 +1,107 @@
...
+		final := getCodeMarkdown(r.captureMap, r.word, lines, start, end)
+
+		// replace word in msg starting from r.index only once
+		msg = msg[:r.index] + strings.Replace(msg[r.index:], r.word, final, 1)
```
ali-farooq0: Hmmm, after updating the text here, wouldn't `r.index` become invalid for the next iteration as things have moved? Unless maybe I'm missing something... 🤔 

One way to fix this would be to always do the replacement starting from the right-most index and work your way in towards the start of the string. That way the indexes are valid across all replacements.

server/permalinks.go
```diff
@@ -0,0 +1,107 @@
...
+			continue
+		}
+
+		// get the file contents
+		opts := github.RepositoryContentGetOptions{
+			Ref: r.captureMap["commit"],
+		}
+		// TODO: make all of these requests concurrently.
+		reqctx, cancel := context.WithTimeout(context.Background(), permalinkReqTimeout)
+		fileContent, _, _, err := p.githubClient.Repositories.GetContents(reqctx,
```
ali-farooq0: Also it looks like `GetContents` has a max 1mb limit, we should probably make sure we return something valid in cases where we can't find the start/end from the file.

https://github.com/google/go-github/blob/0830f2846d799e0c1b07b643f79c4a2af7e54295/github/repos_contents.go#L114

server/permalinks.go
ali-farooq0: Also we should make sure we return something valid when `start` `end` are not in `decoded`

go.mod
```diff
@@ -33,5 +33,6 @@ require (
 replace (
 	git.apache.org/thrift.git => github.com/apache/thrift v0.0.0-20180902110319-2566ecd5d999
 	// Workaround for https://github.com/golang/go/issues/30831 and fallout.
-	github.com/golang/lint => github.com/golang/lint v0.0.0-20190227174305-8f45f776aaf1
+	github.com/golang/lint => github.com/golang/lint v0.0.0-20181217174547-8f45f776aaf1
+	willnorris.com/go/imageproxy v0.8.1-0.20190326225038-d4246a08fdec => willnorris.com/go/imageproxy v0.8.1-0.20190422234945-d4246a08fdec
```
ali-farooq0: is this a result of updating go dependencies? Don't think `imageproxy` is being used anywhere 🤔

server/utils.go
```diff
@@ -173,3 +176,129 @@ func fixGithubNotificationSubjectURL(url string) string {
...
+	}
+
+	switch len(parts) {
+	case 1: // just a single line
+		l := getLine(parts[0])
+		if l == -1 {
+			return -1, -1
+		}
+		if l < 3 {
+			return 0, l + 3
```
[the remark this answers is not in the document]
marianunez: @ali-farooq0 I think it makes sense to start from 0 so at least you get the first couple of lines before. For example, if l = 2, you get from the beginning of the file in the code lines and not start just from the line 2. This way is consistent as much possible with the case of `return l - 3, l + 3`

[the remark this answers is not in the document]
ali-farooq0: @marianunez Hmm, fair enough, but then it's behaving differently for a rage where we return `start, end`. My thinking was that if someone wants to show a certain range they'll use that range i.e. `L13-L15`, but if they just have `L13` then the intent would be to show just that line.

I believe that's what github does as well.

image

link: https://github.com/mattermost/mattermost-server/pull/12390#discussion_r335118081

server/plugin.go
```diff
@@ -53,6 +55,13 @@ type Plugin struct {
 	configuration *configuration
 }
 
+// New returns an instance of a Plugin
+func New() *Plugin {
+	return &Plugin{
+		githubPermalinkRegex: regexp.MustCompile(`https?://(?P<haswww>www\.)?github\.com/(?P<user>[\w-]+)/(?P<repo>[\w-]+)/blob/(?P<commit>\w+)/(?P<path>[\w-/.]+)#(?P<line>[\w-]+)?`),
```
ali-farooq0: Would be good to doc what does this match.

server/utils.go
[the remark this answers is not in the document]
marianunez: We were following a different approach of showing the surrounding lines to give more context like it does in the code reviews screens. For example:

Screen Shot 2019-10-16 at 8 08 17 PM

2/5 we should keep this because most of the time one line does not say enough.
However, we could also give the range the surrounding lines or limit the one liners to exactly one line to be consistent if you feel strongly about this.

go.mod
agnivade: This is a result of using Go 1.13. Because 1.13 validates the commit hash with the timestamp, we need these replace blocks to correctly get the dependency. `imageproxy` is a transitive dependency that is used somewhere else in the module graph.

I have kept this in a separate commit so that they remain separate.

server/plugin.go
```diff
@@ -53,6 +56,13 @@ type Plugin struct {
 	configuration *configuration
 }
 
+// New returns an instance of a Plugin
+func New() *Plugin {
```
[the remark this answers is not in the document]
agnivade: Totally agree. The package name [stutter](https://golang.org/doc/effective_go.html#package-names) was so hammered in me, I didn't realise this is the main package :stuck_out_tongue: Good call.

server/plugin.go
```diff
@@ -53,6 +56,13 @@ type Plugin struct {
 	configuration *configuration
 }
 
+// New returns an instance of a Plugin
```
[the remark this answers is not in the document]
agnivade: Thanks. I always miss to add the periods.

server/permalinks.go
```diff
@@ -0,0 +1,119 @@
...
+	"fmt"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/google/go-github/v25/github"
+)
+
+const maxPermalinkReplacements = 10
+const permalinkReqTimeout = 2 * time.Second
```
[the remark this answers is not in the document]
agnivade: Sure.

server/permalinks.go
```diff
@@ -0,0 +1,119 @@
...
+	"encoding/hex"
+	"fmt"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/google/go-github/v25/github"
+)
+
+const maxPermalinkReplacements = 10
```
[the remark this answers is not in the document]
agnivade: Sure.

server/permalinks_test.go
```diff
@@ -0,0 +1,160 @@
...
+						"haswww": "",
+						"line":   "L15-L22",
+						"path":   "app/authentication.go",
+						"user":   "mattermost",
+						"repo":   "mattermost-server",
+					},
+				},
+			},
+		},
+	}
```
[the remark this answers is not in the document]
agnivade: Already added a one-line replacement from your comment earlier. Let me know if you meant something else.

server/utils.go
```diff
@@ -173,3 +176,129 @@ func fixGithubNotificationSubjectURL(url string) string {
...
+		}
+		return l - 3, l + 3
+	case 2: // a line rage
```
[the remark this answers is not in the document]
agnivade: Not really :wink:  https://golang.org/src/debug/gosym/symtab.go#L374. But I don't mind. We should just remember to be consistent whenever we do this.

server/plugin.go
```diff
@@ -116,6 +125,27 @@ func (p *Plugin) OnActivate() error {
...
+	}
+
+	msg := post.Message
+	info, err := p.getGitHubUserInfo(post.UserId)
+	if err != nil {
+		p.API.LogError("error in getting user info", "error", err)
+		return nil, ""
+	}
+	// TODO: store it during instance creation time and reuse it for all calls.
+	p.githubClient = p.githubConnect(*info.Token)
```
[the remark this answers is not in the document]
agnivade: Personally, I still think it's wasteful to create a new client for every single message unnecessarily, even if there is no change. These message handlers are on the fast path and should take the minimum time possible. Thoughts ?

server/utils.go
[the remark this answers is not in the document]
agnivade: I agree with @marianunez about trying to show more context. Keeping it the same for now. Let me know if you want this changed.

server/permalinks.go
```diff
@@ -0,0 +1,107 @@
...
+		// TODO: make all of these requests concurrently.
+		reqctx, cancel := context.WithTimeout(context.Background(), permalinkReqTimeout)
+		fileContent, _, _, err := p.githubClient.Repositories.GetContents(reqctx,
+			r.captureMap["user"], r.captureMap["repo"], r.captureMap["path"], &opts)
+		if err != nil {
+			p.API.LogError("error while fetching file contents", "error", err.Error(), "path", r.captureMap["path"])
+			cancel()
+			continue
+		}
+		cancel()
```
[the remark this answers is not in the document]
agnivade: Ah no, we need this. The context needs to be cancelled after the call is done. We cannot use `defer` because it won't get cancelled until the function returns which is something we don't want. Then we need to declare a closure and that hurts readability.

server/plugin.go
agnivade: I thought the name was clear in expressing that it matches a github permalink. What do you have in mind ?

server/permalinks.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+	return replacements
+}
+
+// makeReplacements perform the given replacements on the msg and returns
+// the new msg.
+func (p *Plugin) makeReplacements(msg string, replacements []replacement) string {
+	for _, r := range replacements {
+		// quick bailout if the commit hash is not proper
+		if _, err := hex.DecodeString(r.captureMap["commit"]); err != nil {
```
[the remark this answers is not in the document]
agnivade: The only issue is that setting the fields of `captureMap` becomes slightly unwieldy:
```
switch name {
		case "haswww":
			// set field
		case "user":
			// set field
		case "repo":
			// set field
		case "commit":
			// set field
		case "path":
			// set field
		case "line":
			// set field
		}
```

compared to just `r.captureMap[name] = m[j]`. Let me know your thoughts. I don't have strong opinions on both approaches. Just wanted to clarify this.

server/utils.go
```diff
@@ -173,3 +176,129 @@ func fixGithubNotificationSubjectURL(url string) string {
...
+		if i > end {
+			break
+		}
+		buf.Write(scanner.Bytes())
+		buf.WriteByte(byte('\n'))
+		i++
+		w++
+	}
+	if err := scanner.Err(); err != nil {
+		fmt.Println(err)
```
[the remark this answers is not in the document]
agnivade: Sure

server/permalinks.go
agnivade: Yes, I had mentioned this in my design sketch. I think it is good to have this limit to prevent unnecessarily downloading huge files just to show a 2 line preview. In this case, it will just not expand the link, that's all.

server/permalinks.go
agnivade: Simplified it. Good call on start/end out of range. Fixed that too.

server/utils.go
agnivade: This is now gone because of the simplification.

server/permalinks.go
agnivade: > Hmmm, after updating the text here, wouldn't r.index become invalid for the next iteration as things have moved?

Yeah :man_facepalming:  .. I was also stuck trying to understand how is it working. The reason is because it is trying to replace the string just once. So even if the starting index is incorrect. It just needs the string to be present in the entire string. So accidentally, this worked :sweat_smile: 

Fixed it to do it in reverse.

server/permalinks_test.go
```diff
@@ -0,0 +1,160 @@
+package main
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetReplacements(t *testing.T) {
```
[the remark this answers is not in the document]
agnivade: The non-commit permalink is tested in `makeReplacements` unfortunately. I added the one for single line commit. The incorrect line range is tested again in `makeReplacements`, I added the test in `TestGetLineNumbers`.

See my main comment about why it is hard to test `makeReplacements` as a whole.

agnivade:
Oops my bad ! I mistakenly amended to the previous commit. So you won't see the exact changes done for this review. You will also see Maria's review changes. Sorry about that :(

server/utils.go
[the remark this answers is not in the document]
ali-farooq0: @marianunez @agnivade Alright lets keep it as is for now and we can always update this later if need be. 👌

server/plugin.go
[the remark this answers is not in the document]
marianunez: Agreed @agnivade. However, as discussed on the [this review comment](https://github.com/mattermost/mattermost-plugin-github/pull/137#discussion_r335300453) keeping this in the struct would also require updating it anytime there is a change in the associated account info which we should defer to another PR. 

+1 we should remove `githubClient` from the struct for the scope of this PR as @ali-farooq0 suggested.

agnivade:
Sounds good. Will update.

ali-farooq0:
@agnivade just a gentle reminder not to force push an on-going review. https://developers.mattermost.com/blog/submitting-great-prs/#4-avoid-force-pushing

server/utils.go
```diff
@@ -173,3 +176,127 @@ func fixGithubNotificationSubjectURL(url string) string {
 func fullNameFromOwnerAndRepo(owner, repo string) string {
 	return fmt.Sprintf("%s/%s", owner, repo)
 }
+
+// filter lines in a string from start to end
+func filterLines(s string, start, end int) (string, error) {
+	scanner := bufio.NewScanner(strings.NewReader(s))
+	var buf strings.Builder
+	i := 1
+	for scanner.Scan() {
```
ali-farooq0: Any reason not to merge `i` into this line?
`for i := 1; i <= end || scanner.Scan(); i = i++`

server/permalinks.go
```diff
@@ -0,0 +1,131 @@
...
+			continue
+		}
+
+		// get the file contents
+		opts := github.RepositoryContentGetOptions{
+			Ref: r.captureMap["commit"],
+		}
+		// TODO: make all of these requests concurrently.
+		reqctx, cancel := context.WithTimeout(context.Background(), permalinkReqTimeout)
+		fileContent, _, _, err := ghClient.Repositories.GetContents(reqctx,
```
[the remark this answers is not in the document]
ali-farooq0: Apologies for not bringing this up in the first pass.

server/permalinks.go
```diff
@@ -0,0 +1,131 @@
...
+
+// getReplacements returns the permalink replacements that needs to be performed
+// on a message.
+func (p *Plugin) getReplacements(msg string) []replacement {
+	// find the permalinks from the msg using a regex
+	matches := p.githubPermalinkRegex.FindAllStringSubmatch(msg, -1)
+	indices := p.githubPermalinkRegex.FindAllStringIndex(msg, -1)
+	var replacements []replacement
+	for i, m := range matches {
+		// have a limit on the no. of replacements to do
```
ali-farooq0: nit: `no.` -> `number`

server/permalinks_test.go
```diff
@@ -0,0 +1,178 @@
...
+					},
+				},
+			},
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.name, func(t *testing.T) {
+			replacements := p.getReplacements(tc.input)
+			require.Equalf(t, tc.numReplacements, len(replacements), "unexpected no. of replacements for %s", tc.input)
```
ali-farooq0: nit `number`

server/plugin.go
```diff
@@ -41,7 +42,8 @@ const (
 
 type Plugin struct {
 	plugin.MattermostPlugin
-	githubClient *github.Client
+	githubClient         *github.Client
```
ali-farooq0: should this be removed?

server/permalinks.go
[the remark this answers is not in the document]
agnivade: Hmm .. I can move the commit hash checking logic up to `getReplacements`. But if you don't want to make the github call in `makeReplacements`, do you want to make it in `getReplacements` ? That will make `getReplacements` untestable. Could you clarify ?

I can add tests for `filterLines` too if you would prefer that. And once we move the commit hash checking to `getReplacements` we can test that too. That just leaves the network call part not tested.

server/plugin.go
agnivade: Yeah, it can. But now it's unrelated to the PR. This field was never used. I can do it if you want.

server/utils.go
agnivade: ohh I like it :+1:

agnivade:
Just one clarification needed here @ali-farooq0 - https://github.com/mattermost/mattermost-plugin-github/pull/137#discussion_r340173972.

Thanks for the reminder on force pushing. In my past, I have always force pushed for resolving merge conflicts with master. So it's like a reflex ;) . Getting used to the new flow.

P.S. I will be AFK this week till next Monday. So I will get back to this once I'm back.

server/plugin.go
ali-farooq0: Since it's not being used I'd recommend to remove it. No harm if it's not being used.

ali-farooq0:
> Just one clarification needed here @ali-farooq0 - [#137 (comment)](https://github.com/mattermost/mattermost-plugin-github/pull/137#discussion_r340173972).
> 
> Thanks for the reminder on force pushing. In my past, I have always force pushed for resolving merge conflicts with master. So it's like a reflex ;) . Getting used to the new flow.
> 
> P.S. I will be AFK this week till next Monday. So I will get back to this once I'm back.

@agnivade Sounds good 👍 . Also replied to the comment.

server/permalinks.go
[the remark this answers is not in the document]
agnivade: >  It will only make the github fetching part untestable.

IIUC, it will make the entire `getReplacements` method untestable. With the current state of the code, there is no way to test everything in `getReplacements` except the github fetch.

Actually the mock client code in the test looks decent and not that complicated. Let me try using that. That seems like the most elegant way forward, rather than juggling with our logic.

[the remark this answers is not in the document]
ali-farooq0: agreed. :)

server/permalinks.go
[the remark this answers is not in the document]
ali-farooq0: Shouldn't this be updated to `defer cancel()`?

server/permalinks_test.go
```diff
@@ -0,0 +1,472 @@
...
+
+	mockPluginAPI.AssertCalled(t, "LogError", "bad git commit hash in permalink", "error", "encoding/hex: invalid byte: U+0068 'h'", "hash", "badhash")
+	mockPluginAPI.AssertCalled(t, "LogError", "error while fetching file contents", "error", "unmarshalling failed for both file and directory content: unexpected end of JSON input and unexpected end of JSON input", "path", "path/file.go")
+}
+
+const (
+	baseURLPath = "/api-v3"
+)
+
+func getClient() (*github.Client, func()) {
```
ali-farooq0: nice! 👍

server/permalinks.go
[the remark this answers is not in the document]
ali-farooq0:
```
// TODO: make all of these requests concurrently.
		reqctx, cancel := context.WithTimeout(context.Background(), permalinkReqTimeout)
		fileContent, _, _, err := ghClient.Repositories.GetContents(reqctx, r.permalinkInfo.user, r.permalinkInfo.repo, r.permalinkInfo.path, &opts)
		defer cancel()
```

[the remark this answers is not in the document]
agnivade: Ah ok, it was not recommended to use `defer` in loops as it has performance implications. But now with open-coded defers in tip, this should be fine. Will change.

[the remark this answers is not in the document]
ali-farooq0: @agnivade hmmm fair point. Yeah I see folks are recommending to close at the end of the loop rather than doing `defer cancel()` Maybe we could do that instead?

agnivade:
@ali-farooq0 - It looks like the build was cancelled ? Can you restart it ?

ali-farooq0:
@DHaussermann this is ready for QA.

DHaussermann:
Hey @agnivade I'm not seeing this work as expected. Once I enable `Enable Code Previews` the next subscription event does get delivered but I see no preview and the the logs  show 
```
{"level":"info","ts":1574089720.0073502,"caller":"go-plugin/stream.go:15","msg":"2019/11/18 15:08:40 RPC call to KVGet API failed: connection is shut down","plugin_id":"github","source":"plugin_stderr"}
{"level":"info","ts":1574089720.00742,"caller":"go-plugin/stream.go:15","msg":"2019/11/18 15:08:40 RPC call to KVGet API failed: connection is shut down","plugin_id":"github","source":"plugin_stderr"}
```
After this subscriptions and @ mentions stop working.

Once I disable `Enable Code Previews` It looks like the plugin crashes and restarts
```
{"level":"error","ts":1574090255.8316064,"caller":"plugin/health_check.go:91","msg":"Health check failed for plugin github, error: Plugin RPC connection is not responding"}
{"level":"warn","ts":1574090255.831667,"caller":"plugin/hclog_adapter.go:51","msg":"error closing client during Kill","plugin_id":"github","wrapped_extras":"errconnection is shut down"}
{"level":"warn","ts":1574090255.8316815,"caller":"plugin/hclog_adapter.go:53","msg":"plugin failed to exit gracefully","plugin_id":"github"}
```
Subscriptions and @ mentions then start working again. 

Can you help me isolate this behavior? Can you confirm if these errors are related to the new feature and if I need to collect more info somehow?

agnivade:
Hi @DHaussermann - Yes I have seen the " RPC call to KVGet API failed: connection is shut down," several times. But I worked around it by just restarting the plugin process. Everything starts working again. I thought it was an unrelated issue and didn't have time to look deeper, because everything works fine with a restart.

On a deeper look, the error comes from the `p.API.KVGet(userID + GITHUB_TOKEN_KEY)` call in `(*Plugin).getGitHubUserInfo` method. I am calling this method in my code, but this is also used in a lot of places. If it is failing, I believe there is a central reason this is failing. Does this help ?

Please let me know how can I help further.

DHaussermann:
Thanks @agnivade can you also give me some guidance on the functionality? 
Once I enable the feature I do get the 1st event delivered but I don't see the file preview. I'm not sure if it's related to the error or I'm trying to use this correctly. Maybe it does not work with all events or the link I'm using? Should these steps work?
1. Create a subscription to a repo that includes `issues`
2. Enable the feature
3. In the target repo, create an issue where the description contains a file like such as:
```
Link Test Issue
https://github.com/DHaussermann/hello-world/blob/master/hello.go
```
**Expected**: Link included in GitHub post shows a preview.

agnivade:
Ah no, you are using it wrong. You don't need any subscriptions or delivery of any event whatsoever. And secondly, that link is not a permalink. It's just a URL to a file.

Here is what I would suggest:

Once you enable the feature, just copy paste any text containing a github permalink. That's all. Then you should be able to see the preview.

As an example type `https://github.com/mattermost/mattermost-server/blob/4d1c6cd446ba5a40fb90d530ceadc4ed7a387424/store/constants.go#L7` in the text box.

You can get a permalink by clicking on a line number of a file in github. Then click on the 3 dots on the left, click on copy permalink. You can also test a line range, by pressing the shift key and clicking on a close-by line to get a line range. Then do the same to get its permalink.

Let me know if that helps.

DHaussermann:
Thanks @agnivade this does help. I think I mistakenly focus on the subscription events because I noticed they stop working :) . With a proper link I see the functionality working. 

I also notice the same behavior with the preview. After it is generated once, the plugin stops working. I then have to disable and re-enable the plugin to see a preview again.

To try and isolate this I deployed the plugin from the `maste`r branch of the Mattermost repo and I can't reproduce the crash or log errors.

Is there anyway to determine conclusively if the issue is introduced by this PR?

agnivade:
@DHaussermann - It was a corner case where an empty userID was being passed. Apparently, restarting the plugin did not send these erroneous messages.

Please check again with the latest patch.

PR #138, opened 2019-10-07
add ability to refresh RHS details

### LINKED ISSUE #131: Github Plugin: Add Refresh ability to RHS
The [GitHub plugin](https://github.com/mattermost/mattermost-plugin-github) RHS lists notifications, pull requests, assigned issues and review requests. It currently displays the data that is available from the current state of the store but does not have the ability to refresh the data in case is outdated. Add support to Refresh the data  by:

  1. Calling Refresh whenever the RHS bar is opened to guarantee latest data is available. 
  1. Add an action icon on the title bar to offer the ability to manually refresh as needed (similar to the one currently available in the team sidebar buttons). Locate the icon in the RHS section title right aligned (in screenshot below, in the “Your Open Pull Requests” title header).

**Current RHS:** 
Screen Shot 2019-09-28 at 5 29 19 PM

**Sidebar Buttons:**

Screen Shot 2019-09-28 at 5 26 57 PM

----

JIRA: https://mattermost.atlassian.net/browse/MM-18879

manland:
mattermost-rhs-right-github

Fixes #131

With part of #132


marianunez:
Thanks @manland! Could you add a closer screenshot of the new refresh button for UX review?

marianunez:
@manland thanks for adding that extra part of #132. While testing it out, did you feel it added value to gray out the read item even though you refreshed manually clicking the button? 

I believe the original intent of graying out was if you lose focus on the browser and when it returned, it would automatically refresh which could've removed items from the list possibly confusing the user of where they went. Thoughts @lieut-data?

manland:
@marianunez I think it's a good feature even when I refresh manually. On my gif I switch between github and MM, but in real life, I imagine I will open PR view and stay on it every day (when I have done the dev in gitlab-plugin because I use gitlab... but it's another story :D) and when a coworker, accept/close a MR I have a clear picture of what happened. Like that I will be able to inactive notif `/github settings notif off`, which is a lot perturbing when I develop. 

But IMHO I think adding a label `merged` or `accepted` which explain why the item is grey will be a good improvement ;) And when I was playing with the RHS, I want to be able to do action directly in MM (close issue, mark as read notif, or accept PR...)

aaronrothschild:
This is a great improvement!  Thanks @manland for your contribution.

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -7,10 +7,32 @@ import PropTypes from 'prop-types';
 import {Badge} from 'react-bootstrap';
 import {makeStyleFromTheme, changeOpacity} from 'mattermost-redux/utils/theme_utils';
 
+//retain previous item to show deleted items between 2 refresh call
+let previousItems = [];
```
lieut-data: @manland, can we promote the `GithubItems` to a `React.Component` and store these as state? (Or, if you're adventurous, to a React Hook?)

While this likely works because we only render one `GitHubItems` at a time, it runs the risk of breaking in edge cases where React decides not to re-render the component, and it makes it hard to properly manipulate this state and achieve a re-render.

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -7,10 +7,32 @@ import PropTypes from 'prop-types';
...
+        newElements = props.items.filter((item) => !previousItems.find((i) => i.id === item.id));
+    } else {
+        previousStateRhs = props.rhsState;
+        newElements = [...props.items];
+    }
+
+    const elements = [...oldElements, ...newElements];
+    previousItems = [...elements];
+
+    return elements.length > 0 ? elements.map((item) => {
```
lieut-data: Pre-existing code, but could we eliminate the ternary here and just return early?
```js
if (elements.length == 0) {
    return <div style={style.container}>{'You have no active items'}</div>;
}
```

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -7,10 +7,32 @@ import PropTypes from 'prop-types';
...
+    let newElements = [];
+
+    if (previousStateRhs === props.rhsState) {
+        oldElements = previousItems.map((item) => {
+            if (!props.items.find((i) => i.id === item.id)) {
+                return {...item, missing: true};
+            }
+            return item;
+        });
+        newElements = props.items.filter((item) => !previousItems.find((i) => i.id === item.id));
```
lieut-data: If the incoming `props.items` have updates to existing elements (e.g. new labels applied to a PR), it seems like we ignore them. Should we take from the new elements if we find it there instead to get any updates?

webapp/src/components/sidebar_right/sidebar_right.jsx
```diff
@@ -44,11 +45,57 @@ export default class SidebarRight extends React.PureComponent {
...
+        };
+    }
+
+    refresh = async (e) => {
+        if (this.state.refreshing) {
+            return;
+        }
+
+        if (e) {
+            e.preventDefault();
```
lieut-data: Is this `preventDefault` necessary?

webapp/src/components/sidebar_right/sidebar_right.jsx
```diff
@@ -44,11 +45,57 @@ export default class SidebarRight extends React.PureComponent {
...
+        case RHSStates.UNREADS:
+            refreshAction = this.props.actions.getUnreads;
+            break;
+        case RHSStates.ASSIGNMENTS:
+            refreshAction = this.props.actions.getYourAssignments;
+            break;
+        default:
+            return;
+        }
+        await refreshAction();
```
lieut-data: If `refreshAction` were to throw an exception, I believe we'd be stuck in `refreshing: true` permanently. Should we add a `try/catch` here?

webapp/src/components/sidebar_right/sidebar_right.jsx
```diff
@@ -44,11 +45,57 @@ export default class SidebarRight extends React.PureComponent {
...
+            refreshAction = this.props.actions.getReviews;
+            break;
+        case RHSStates.UNREADS:
+            refreshAction = this.props.actions.getUnreads;
+            break;
+        case RHSStates.ASSIGNMENTS:
+            refreshAction = this.props.actions.getYourAssignments;
+            break;
+        default:
+            return;
```
lieut-data: On the off chance we hit this block, we'll be stuck in `refreshing: true` -- should we check this earlier instead?

webapp/src/components/sidebar_right/github_items.jsx
manland: I'm adventurous^^ please review again :sweat_smile: I keep this conversation open because tests are missing ;)

manland:
@lieut-data thank you very much for this very smart review! I'm sorry for the delay to continue this PR but last week was hot :fire: (wife work travel... :smile:) and I had read lot of blog around react.hooks (i'm angular dev :man_shrugging:), anyway it was funny!

I'll implem test as soon as possible :man_cook:

manland:
> Thanks @manland! Could you add a closer screenshot of the new refresh button for UX review?

@marianunez this is some screenshot of refresh button (default, mouse hover, dark default, dark mouse hover)... sorry for delay [explication](https://github.com/mattermost/mattermost-plugin-github/pull/138#issuecomment-542908421)!

Sélection_296
Sélection_297
Sélection_294
Sélection_295

manland:
I found a bug in my last implementation : if i click refresh 2 times, the second time will remove the item previously marked has `missing`. Because compo render is called and `usePrevious` return not previous but last items : 

* first call 8 items
* second call (on refresh) 7 items ==> 1 marked has missing
* third call (on refresh again) 7 items ==> compared to previous no missing

To fix it, I can implement the second parameter of React.memo to check if a render is needed depending on items ;) But it will increase the complexity of this component ;)

@lieut-data what do you think about that?

marianunez:
> @marianunez this is some screenshot of refresh button (default, mouse hover, dark default, dark mouse hover)... sorry for delay [explication](https://github.com/mattermost/mattermost-plugin-github/pull/138#issuecomment-542908421)!
> 
> Sélection_296
> Sélection_297
> Sélection_294
> Sélection_295

Thanks @manland! cc @asaadmahmood for UX review 👆

asaadmahmood:
@marianunez it should use center-channel-color opacity .6 for its default state, and then center-channel-color opacity .8 for its hovered state.

lieut-data:
@manland, I admit that I'm not yet well versed enough with React Hooks to describe the right solution in that context.

But as a `React.Component`, I'd expect to update my state's copy of the items only when `refresh` is called, and not try to tie it to re-renders due to props changes. Perhaps you could keep the last timestamp the refresh button was pressed (in the sidebar right), and pass that as a prop to the items component? So long as that value hasn't changed, you can keep your "old" props around and use them to render dimmed, etc.

I'll leave it to you to figure out if it makes sense to model that a hook or as a `React.Component` with old-school state.

manland:
I'm always on it... now I have test I can try to find a solution :D

Sélection_299

@lieut-data i'm not a big fan of this kind of hack, because if user click on refresh in LHS, compo will not be updated on RHS. I continue to play with hooks, and if nothing works I will make a real old school component^^

lieut-data:
> i'm not a big fan of this kind of hack, because if user click on refresh in LHS, compo will not be updated on RHS. I continue to play with hooks, and if nothing works I will make a real old school component

@manland, if I understand hooks correctly, there no more or less powerful than components. Just a cleaner way to write certain patterns. So we'll still need to solve the same problem. 

With respect to the refresh, my gut instinct would be to model the timestamp in the Redux store instead of internal state. This would make the the transition across a manual / automatic refreshes and open / close transitions easy to follow.

manland:
> With respect to the refresh, my gut instinct would be to model the timestamp in the Redux store instead of internal state. This would make the the transition across a manual / automatic refreshes and open / close transitions easy to follow.

I'm sorry @lieut-data but I don't understand your solution :sweat: 

I have fix the issue by adding an `areEqual` method in React.memo to be able to not re-render if props are the same.

Just tell me if it's the `react way` or not ;)

@asaadmahmood css updated :muscle: :stars:

asaadmahmood:
@manland Can I get an updated screenshot?

.gitignore
```diff
@@ -1,3 +1,4 @@
 dist
 .DS_STore
-node_modules
\ No newline at end of file
+node_modules
+.vscode
```
marianunez: Missing a new line here

webapp/.eslintrc.json
```diff
@@ -34,6 +34,14 @@
   "settings": {
     "import/resolver": "webpack"
   },
+  "overrides": [
+    {
+        "files": ["tests/setup.js"],
+        "rules": {
+            "no-console": "off"
```
marianunez: Do we need this? I believe this is turned on by default on all repos

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -125,4 +158,10 @@ const notificationReasons = {
...
 };
 
-export default GithubItems;
+function areEquals(prevProps, nextProps) {
+    return prevProps.theme === nextProps.theme &&
+        prevProps.rhsState === nextProps.rhsState &&
+        prevProps.items.reduce((acc, i) => `${acc}-${i.id}`, '') === nextProps.items.reduce((acc, i) => `${acc}-${i.id}`, '');
+}
+
+export default React.memo(GithubItems, areEquals);
```
marianunez: Nice! 👌

webapp/src/components/sidebar_right/sidebar_right.css
```diff
@@ -0,0 +1,11 @@
...
+    float: right;
+    border: none;
+    background: none;
+    color: var(--center-channel-color);
+    opacity: 0.6;
+}
+
+.GithubPlugin__SidebarRight__refresh-button:hover {
+    opacity: 0.8;
+}
```
marianunez: Missing new line here

webapp/src/components/sidebar_right/github_items.jsx
lieut-data: Hmm, `React.memo` is a strategy to boost performance by reducing unnecessary renders, but it seems were relying on it here to help enforce business logic about when to transition "old state" so as to achieve the desired rendering logic.

This approach is still buggy. As an example, if I change my theme, the dimmed items will disappear, and it's really hard to fix this.

I'd like to suggest we move all of this logic into a reducer instead. It should be the reducer's responsibility to decide, based on the incoming actions, when to clear the "old state" and when to preserve it.

jasonblais:
@manland let us know if you have any questions about the feedback above. We'd be happy to clarify! :)

lieut-data:
@manland, would you still be interested in contributing towards this pull request? Let me know if there's any clarification I can help with :)

manland:
@lieut-data oups sorry this issue was out of my radar :man_shrugging: 

Yes I want to finish it, but I have no idea how to implement a good solution. Could you, please, give me more details on your proposed implementation?

lieut-data:
Absolutely! I don't yet have enough experience to frame this as a React hook, but let me just talk about fundamentals.

To render some items as "dimmed", we effectively need to remember the set of items we loaded when the RHS was first opened or when it was last manually refreshed, in addition to the set of items we've received via a websocket event.

Let's take `mentions` as a concrete example. We already model these in the Redux reducer:
```js
function mentions(state = [], action) {
    switch (action.type) {
    case ActionTypes.RECEIVED_MENTIONS:
        return action.data;
    default:
        return state;
    }
}
```

We should start by modifying the `RECEIVED_MENTIONS` handling to merge the old `state` with the incoming `action.data`. If we find an item in the old `state` that's not in the new `action.data`, we can add a boolean, say `stale`. The client-side can directly use this bit to decide whether to render the item as dimmed. In pseudo-code, something like:
```js
    case ActionTypes.RECEIVED_MENTIONS:
        // clone state (can't mutate in place)
        // iterate through items in action.data
        //     if item.id is not in state, set state[item.id].stale = true
        //     otherwise overwrite with state[item.id] = action.data[item.id]
        // return cloned state
}
```

For the initial load event -- i.e. opening the RHS -- we can probably just purge the entire state -- stale and all -- and wait for the incoming mentions. When manually refreshing, we want to both get rid of the stale mentions and replace with the new mentions. The easiest way to do that is to add a new action that also makes the same HTTP request to GitHub, but gets handled the old way:
```
    case ActionTypes.REFRESHED_MENTIONS:
        return action.data;
```

I'm sure there's different ways to slice this, but the key take away is to model the desired state in Redux, use actions to transform the data based on events, and avoid tying up too much of this business logic into the component itself.

hanzei:
Hi @manland,

Given that this PR has been stale for a long time and I don't expect an update happening in the near future, I will close it for now. Please let me know, if you want to continue your work.

manland:
Hey @hanzei yes thank you... I really want to... But I can't... If someone want to finish it, I have done 80% of the dev, the last thing it's to move the logic from component to reducer requested by @lieut-data

PR #145, opened 2019-10-18
MM-18593: updated manifest metadata

lieut-data:
Record the HomepageURL, SupportURL and IconPath for use with the Plugin Marketplace.

Relates-to: https://mattermost.atlassian.net/browse/MM-18593


plugin.json
```diff
@@ -2,6 +2,9 @@
     "id": "github",
     "name": "GitHub",
     "description": "GitHub plugin for Mattermost.",
+    "homepage_url": "https://github.com/mattermost/mattermost-plugin-github",
```
lieut-data: These fields are modelled in https://github.com/mattermost/mattermost-plugin-github/pull/145, but it doesn't hurt to add them early. I have similar PRs prepared for other plugins once I prove that the generator works with a new release of this one.

lieut-data:
@prapti, this change relates to https://github.com/mattermost/mattermost-server/pull/12823, in that we're updating metadata here, but nothing is yet consuming it. The pending marketplace changes re: same will enable.

PR #149, opened 2019-10-24
Bump version to 0.12.0

lieut-data:
Bumping version to v0.12.0 to cut a new release, with the primary purpose of including the recent `plugin.json` metadata changes for programmatic consumption by the marketplace.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,8 +10,6 @@ require (
 	github.com/huandu/xstrings v1.2.0 // indirect
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/mattermost/mattermost-server v0.0.0-20191017141203-48c06e9bce3b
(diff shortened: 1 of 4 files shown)


go.mod
```diff
@@ -10,6 +10,8 @@ require (
 	github.com/huandu/xstrings v1.2.0 // indirect
 	github.com/imdario/mergo v0.3.7 // indirect
 	github.com/mattermost/mattermost-server v0.0.0-20191017141203-48c06e9bce3b
+	github.com/modern-go/concurrent v0.0.0-20180306012644-bacd9c7ef1dd // indirect
```
levb: just wondering why these weren't in before, tidy-ed?

lieut-data: Hmm, uncertain -- I'll push a `go mod tidy` to fix this for now, but I note they come back with a `make clean test`.

Issue #125, opened 2019-09-24
Allow excluding organisation members from subscription events 


hanzei:
There should be an option to not post issue and PR events from organisation members. If the flag `--exclude-org-member` is provided when subscribing to a repo, events from organisation members will not trigger notification.

If not organisation is set (https://github.com/mattermost/mattermost-plugin-github/blob/master/plugin.json#L45-L48), this flag has no effect.

enolal826:
I'd like to work on this! Thanks.

hanzei:
That's awesome, thanks @enolal826 :tada:

enolal826:
Hello!

Two aspects:
- Since currently no flags are supported, I've thought it'd be more convenient for the Subscription struct to have a "flags" field. What do you think?

- As of now, when subscribing there is just one field "features" supported and is supposed to be a comma-separated string, but other features are also considered:
`features = strings.Join(parameters[1:], " ")`

It means that for the command
`/github subscribe owner/repo pulls,issues anotherThing`
both "pulls,issues" and "anotherThing" would be inserted in the "feature" field. I don't know if this is something expected

Thanks!

hanzei:
1. I would like to have the "flags" as boolean fields of in the `Subscription` struct. This way we make sure that we do the parsing of the input only once and then have all data in the struct
2. That is not expected. "anotherThing" should not go into `Features`.

Let me know, if this resolves your questions.

enolal826:
Ok, I see flags as Subscription fields a good idea. It should also be more efficient than having it in array of strings as we won't have to iterate just to check if the flag is set. Maybe another struct SubscriptionFlags could be included.

Thanks @hanzei!

PR #154, opened 2019-11-06
Exclude organization members feature

enolal826:
- Command: we iterate through every parameter, identifying if it's a "feature" or a "flag" (for this purpose, some helper functions have been included in "utils")

- Subscription: a new struct for flags has been created.

- Plugin: a new method isUserOrganizationMember has been added.

- Webhook: a new method excludeConfigOrgMember has been included. It's responsible for fetching configuration and creating the GitHub client.

I have created two different methods so isUserOrganizationMember could be reused without any links to the purpose of this task.

Thanks!

Fixes https://github.com/mattermost/mattermost-plugin-github/issues/125

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -99,6 +99,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 	case "subscribe":
 		config := p.getConfiguration()
 		features := "pulls,issues,creates,deletes"
+		flags := []string{}
 
 		txt := ""
 		if len(parameters) == 0 {
@@ -122,12 +123,27 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 			p.postCommandResponse(args, txt)
 			return &model.CommandResponse{}, nil
 		} else if len(parameters) > 1 {
-			features = strings.Join(parameters[1:], " ")
+			featureList := []string{}
+
+			for _, element := range parameters[1:] {
+				if isFlag(element) {
+					flags = append(flags, parseFlag(element))
+				} else {
+					featureList = append(featureList, element)
+				}
+			}
+
+			if len(featureList) > 1 {
+				p.postCommandResponse(args, "Just one {features} parameter is allowed")
+				return &model.CommandResponse{}, nil
+			} else if len(featureList) == 1 {
+				features = featureList[0]
+			}
 		}
 
 		_, owner, repo := parseOwnerAndRepo(parameters[0], config.EnterpriseBaseURL)
 		if repo == "" {
-			if err := p.SubscribeOrg(context.Background(), githubClient, args.UserId, owner, args.ChannelId, features); err != nil {
+			if err := p.SubscribeOrg(context.Background(), githubClient, args.UserId, owner, args.ChannelId, features, flags); err != nil {
 				p.postCommandResponse(args, err.Error())
 				return &model.CommandResponse{}, nil
 			}
@@ -136,7 +152,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 			return &model.CommandResponse{}, nil
 		}
 
-		if err := p.Subscribe(context.Background(), githubClient, args.UserId, owner, repo, args.ChannelId, features); err != nil {
+		if err := p.Subscribe(context.Background(), githubClient, args.UserId, owner, repo, args.ChannelId, features, flags); err != nil {
 			p.postCommandResponse(args, err.Error())
 			return &model.CommandResponse{}, nil
 		}
--- a/server/command.go
+++ b/server/command.go
@@ -102,7 +102,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 	case "subscribe":
 		config := p.getConfiguration()
 		features := "pulls,issues,creates,deletes"
-		flags := []string{}
+		flags := SubscriptionFlags{}
 
 		txt := ""
 		if len(parameters) == 0 {
@@ -130,7 +130,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 
 			for _, element := range parameters[1:] {
 				if isFlag(element) {
-					flags = append(flags, parseFlag(element))
+					flags.AddFlag(parseFlag(element))
 				} else {
 					optionList = append(optionList, element)
 				}
--- a/server/command.go
+++ b/server/command.go
@@ -123,21 +123,21 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 			p.postCommandResponse(args, txt)
 			return &model.CommandResponse{}, nil
 		} else if len(parameters) > 1 {
-			featureList := []string{}
+			optionList := []string{}
 
 			for _, element := range parameters[1:] {
 				if isFlag(element) {
 					flags = append(flags, parseFlag(element))
 				} else {
-					featureList = append(featureList, element)
+					optionList = append(optionList, element)
 				}
 			}
 
-			if len(featureList) > 1 {
-				p.postCommandResponse(args, "Just one {features} parameter is allowed")
+			if len(optionList) > 1 {
+				p.postCommandResponse(args, "Just one option is allowed")
 				return &model.CommandResponse{}, nil
-			} else if len(featureList) == 1 {
-				features = featureList[0]
+			} else if len(optionList) == 1 {
+				features = optionList[0]
 			}
 		}
 
--- a/server/command.go
+++ b/server/command.go
@@ -134,7 +134,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 			}
 
 			if len(optionList) > 1 {
-				p.postCommandResponse(args, "Just one option is allowed")
+				p.postCommandResponse(args, "Just one list of features is allowed")
 				return &model.CommandResponse{}, nil
 			} else if len(optionList) == 1 {
 				features = optionList[0]
--- a/server/command.go
+++ b/server/command.go
@@ -16,7 +16,7 @@ const COMMAND_HELP = `* |/github connect| - Connect your Mattermost account to y
 * |/github disconnect| - Disconnect your Mattermost account from your GitHub account
 * |/github todo| - Get a list of unread messages and pull requests awaiting your review
 * |/github subscribe list| - Will list the current channel subscriptions
-* |/github subscribe owner[/repo] [features]| - Subscribe the current channel to receive notifications about opened pull requests and issues for an organization or repository
+* |/github subscribe owner[/repo] [features] [flags]| - Subscribe the current channel to receive notifications about opened pull requests and issues for an organization or repository
   * |features| is a comma-delimited list of one or more the following:
     * issues - includes new and closed issues
 	* pulls - includes new and closed pull requests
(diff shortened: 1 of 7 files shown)


jasonblais:
@marianunez / @ali-farooq0 friendly reminder to help review of this PR some time this coming week :)

server/webhook.go
```diff
@@ -163,6 +163,23 @@ func (p *Plugin) permissionToRepo(userID string, ownerAndRepo string) bool {
...
 
+func (p *Plugin) excludeConfigOrgMember(user *github.User, subscription *Subscription) bool {
+	if !subscription.ExcludeOrgMembers() {
+		return false
+	}
+
+	info, err := p.getGitHubUserInfo(subscription.CreatorID)
+
+	if err != nil {
+		return false
```
marianunez: We should log the error here before returning possibly as a warning

server/command.go
```diff
@@ -122,12 +123,27 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+			for _, element := range parameters[1:] {
+				if isFlag(element) {
+					flags = append(flags, parseFlag(element))
+				} else {
+					featureList = append(featureList, element)
+				}
+			}
+
+			if len(featureList) > 1 {
+				p.postCommandResponse(args, "Just one {features} parameter is allowed")
```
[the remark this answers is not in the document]
enolal826: Hello @marianunez, thanks for your comments!

I don't actually know if what the expected behaviour is. Currently we're allowing:

`/github subscribe {owner} {features}`

What we're checking looping over the expression is basically that one (or none) parameter is supplied (in this case "features", which is a comma-separated string). I mean, this command:

`/github subscribe repo issues,pulls anotherThing --exclude-org-member`

shouldn't be valid. With the previous implementation we would be inserting in the `Subscription` struct:

`Features: issues,pulls anotherThing`

but "Features" is just the command-separated string.

I don't know if this makes sense

Thanks!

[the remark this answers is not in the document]
marianunez: I see. Agreed that before it was not taking that into account and was taking everything after the first parameter. Makes sense your implementation then 👍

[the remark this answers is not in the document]
enolal826: Having a look at it again, I think the naming might be confusing. Instead of featureList we should use parameterList.

I'll change that too if you're ok with it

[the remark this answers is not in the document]
marianunez: Yes I guess `featureList` now is not enough considering we are adding flag support too. However, parameterList not sure if it will be confused with the whole list of initial `parameters` [here](https://github.com/mattermost/mattermost-plugin-github/blob/b89d7b68e4c5b8760c76df45bcc207f4521131a7/server/command.go#L59)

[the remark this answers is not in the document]
enolal826: Maybe optionList?

Once you are inside a command (in this case, "subscribe"), you have "options" (in this case, only features)

What do you think?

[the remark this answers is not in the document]
marianunez: `optionList` sounds better 💯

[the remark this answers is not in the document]
enolal826: Great! Done. Thanks a lot.

server/plugin.go
```diff
@@ -442,6 +442,20 @@ func (p *Plugin) checkOrg(org string) error {
...
 
+func (p *Plugin) isUserOrganizationMember(githubClient *github.Client, user *github.User, organization string) bool {
+	if organization == "" {
+		return false
+	}
+
+	isMember, _, err := githubClient.Organizations.IsMember(context.Background(), organization, *user.Login)
+
+	if err != nil {
+		return false
```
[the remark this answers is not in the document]
enolal826: Done, thanks!

server/webhook.go
enolal826: Done, thanks!

server/command.go
```diff
@@ -122,12 +123,27 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+			for _, element := range parameters[1:] {
+				if isFlag(element) {
+					flags = append(flags, parseFlag(element))
+				} else {
+					optionList = append(optionList, element)
+				}
+			}
+
+			if len(optionList) > 1 {
+				p.postCommandResponse(args, "Just one option is allowed")
```
marianunez: To be even more clear here I suggest: 
```suggestion
				p.postCommandResponse(args, "Just one list of features is allowed")
```

Given that we defined options = feature list + flags

DHaussermann:
@marianunez would you be able to provide any high-level tests steps for this? I can regression test subscriptions but I'm not sure how validate `Plugin: a new method isUserOrganizationMember has been added.`

DHaussermann:
@enolal826 when I try to pass in the `exclude-org-member` flag my subscription does not work and I see no event deliveries. I have ensure that the cofig locks the organization so, the feature should be working. 

There may be unrelated problem where there is not proper feedback when the subscription command is invalid. but, for now, can you help by providing a sample subscribe command where 1 or multiple members are excluded. 
Can we also add this to the help?

server/utils_test.go
```diff
@@ -78,3 +78,51 @@ func TestParseOwnerAndRepo(t *testing.T) {
 		assert.Equal(t, repo, tc.ExpectedRepo)
 	}
 }
+
+func TestIsFlag(t *testing.T) {
```
ali-farooq0: nice tests! 👍

server/utils_test.go
```diff
@@ -78,3 +78,51 @@ func TestParseOwnerAndRepo(t *testing.T) {
...
+
+func TestContainsValue(t *testing.T) {
+	tcs := []struct {
+		List     []string
+		Value    string
+		Expected bool
+	}{
+		{List: []string{"value1", "value2"}, Value: "value1", Expected: true},
+		{List: []string{}, Value: "value1", Expected: false},
+		{List: []string{"value1", "value2"}, Value: "value1", Expected: true},
```
ali-farooq0: Should this case be removed since it looks to be the same as on line 120? unless you wanted to test `value2` as opposed to `value1`?

server/webhook.go
```diff
@@ -163,6 +163,24 @@ func (p *Plugin) permissionToRepo(userID string, ownerAndRepo string) bool {
 	return true
 }
 
+func (p *Plugin) excludeConfigOrgMember(user *github.User, subscription *Subscription) bool {
+	if !subscription.ExcludeOrgMembers() {
+		return false
+	}
+
+	info, err := p.getGitHubUserInfo(subscription.CreatorID)
+
```
ali-farooq0: minor nit: empty line.

server/plugin.go
```diff
@@ -442,6 +442,21 @@ func (p *Plugin) checkOrg(org string) error {
 	return nil
 }
 
+func (p *Plugin) isUserOrganizationMember(githubClient *github.Client, user *github.User, organization string) bool {
+	if organization == "" {
+		return false
+	}
+
+	isMember, _, err := githubClient.Organizations.IsMember(context.Background(), organization, *user.Login)
+
```
ali-farooq0: nit :empty line

server/subscriptions.go
```diff
@@ -108,6 +117,9 @@ func (p *Plugin) Subscribe(ctx context.Context, githubClient *github.Client, use
 		CreatorID:  userId,
 		Features:   features,
 		Repository: fullNameFromOwnerAndRepo(owner, repo),
+		Flags: SubscriptionFlags{
+			ExcludeOrgMembers: containsValue(flags, "exclude-org-member"),
```
ali-farooq0: I think it'll be slightly better if we move this parsing logic to https://github.com/mattermost/mattermost-plugin-github/pull/154/files#diff-7278f3371df5ef589578d375bde66febR126 and just pass the `SubscriptionFlags` struct around.

The benefit of that will be that we'll have all the `SubscriptionFlags` parsing logic in one spot. With the current approach if we decided to add a new flag you'll have to update in two spots.

enolal826: Thanks for your comments Ali! That was a doubt I had while implementing it. I eventually decided to place it here just because I wasn't sure if this struct should be created within its own domain (within subscriptions.go in this case)

ali-farooq0: @enolal826 
Fair point. Ideally I think what we should be doing is inside `commands.go` we should be creating a `Subscription` struct and passing that in, however that's a much bigger change. 

Also, after your comment I'm thinking maybe we should just put `ExcludeOrgMembers` directly under `Subscription` and get rid of `SubscriptionFlags` and just pass in the list of flags? what do you think?

server/utils_test.go
enolal826: Updated!

server/webhook.go
enolal826: Updated!

server/plugin.go
enolal826: Updated!

server/subscriptions.go
enolal826: Not sure. I think it might be good to have the flags in a different struct (in case we add new flags it might confusing with the "Features" field), but you probably have more knowledge on the codebase, so I can update it if you prefer it

DHaussermann:
Hey @enolal826 thanks for updating the help to include the flag. 

I'm still not seeing this work on my server. (The subscription not delivering before was because I tried  to append values to the the flag which was invalid)  When I use the flag I'm not seeing members being excluded. 

Can you please look over these steps and let me know if you see anything else I'm doing wrong?
1. Build GitHub from `org-members-excluded` branch
2. Configure the plugin with an app and webhook on the GitHub side and enable it 
3. Lock the plugin to an organization such as "DHaussermann"
4. Subscribe to a repo in the Org using the new flag `/github subscribe DHaussermann/hello-world --exclude-org-member`
5. As the owner of the repo (or other member) create a branch or PR
**Observed** The even is deliver in the subscribed channel.

server/subscriptions.go
ali-farooq0: Fair enough, in that case can we move the creation of `SubscriptionFlags` close to where we parse it?

ali-farooq0:
@enolal826 do you mind updating the branch to resolve conflicts as well as answer @DHaussermann's comment above? Thanks!

server/subscriptions.go
enolal826: I've updated it. What do you think now?

ali-farooq0: @enolal826 looks good 👌

ali-farooq0: @enolal826 can you please resolve conflicts and merging master into your branch? thanks!

enolal826: Done, thanks!

enolal826:
I spoke to DHaussermann in MM and he got it working

ali-farooq0:
@DHaussermann this PR is good to go then?

DHaussermann:
@enolal826 thanks for the help on this! I will do the last round of testing today to ensure all events are supported as expected with last change you made. 

One more thing I was curious about is adding documentation to the read-me about the new flags.
Some details we should consider adding for clarity:
1. The flag only works for Members of an Organization and not for Collaborators on a personal account. (0/5 on this one. May be obviously already to GitHub users based on the name)
2. The flag is ignored if the plugin is not actually locked to an Organization
3. The flag works does not work in cases where the user does not make their membership Public. (during testing when inviting members it seems to make this private by default so, while it seems obvious enough it's still worth mentioning)    

Maybe @aaronrothschild has some thoughts to share on the read-me update.

DHaussermann:
@aaronrothschild  I added the `PM Review` label to this one. Sorry, I'm not sure what the process is for updating the read-me in cases like this. Would updates be done by the contributor and approved by you?

enolal826:
I can update the README and @aaronrothschild could validate it if it's ok for him

aaronrothschild:
> I can update the README and @aaronrothschild could validate it if it's ok for him

Yes, that would be great and helpful for me @enolal826 !  Happy to review, then can remove the "Docs Needed" label when this is all set.

hanzei:
@enolal826 Do you want to include the doc changes in this PR or open a separate one?

jasonblais:
Hi @enolal826! Wasn't sure if you and @aaronrothschild were planning to make the doc changes in this PR or separately. Looks like this is otherwise good to merge. Thanks for your contribution! :)

hanzei:
@aaronrothschild Would you be fine merging the PR as it is and doing the doc changes in a different PR?

aaronrothschild:
> @aaronrothschild Would you be fine merging the PR as it is and doing the doc changes in a different PR?

@hanzei Yes, no problem.

hanzei:
Thanks for adding this valuable feature @enolal826 :man_dancing:

PR #158, opened 2019-11-22
Fix gosimple using using golangci-lint

agnivade:
- S1021: should merge variable declaration with assignment on next line.
- S1009: should omit nil check; len() for nil slices is defined as zero.

### DIFF
--- a/server/webhook.go
+++ b/server/webhook.go
@@ -151,8 +151,7 @@ func (p *Plugin) permissionToRepo(userID string, ownerAndRepo string) bool {
 	if apiErr != nil {
 		return false
 	}
-	var githubClient *github.Client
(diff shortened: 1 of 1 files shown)


DHaussermann:
Thanks @agnivade for the submission. Maybe @hanzei or @jwilander can provide a bit of guidance or high level test steps on this one.  

Seems like maybe just regression testing? I think the 1st change is simply the way the declaration is being made but, for the 2nd change it's no clear to me what regression testing is needed or if there may actually be a functional change in cases where there are no subscription to the events.

agnivade:
For the 2nd change, there will be no functional change as the code is equivalent. As is noted in the PR message: `len() for nil slices is defined as zero`.

PR #159, opened 2019-11-22
Fix deadcode issues using golangci-lint

agnivade:


DHaussermann:
@hanzei  or @agnivade can you please advise if there is any targeted regression testing suitable for this change?

agnivade:
Will let Ben/Maria take a call.

hanzei:
Because this is removing code that is not used, i'm very confident that there will not introduce regressions. A general smoke test is enough.

server/plugin.go
```diff
@@ -24,20 +24,16 @@ import (
...
 	GITHUB_TOKEN_KEY        = "_githubtoken"
-	GITHUB_STATE_KEY        = "_githubstate"
 	GITHUB_USERNAME_KEY     = "_githubusername"
 	GITHUB_PRIVATE_REPO_KEY = "_githubprivate"
 	WS_EVENT_CONNECT        = "connect"
 	WS_EVENT_DISCONNECT     = "disconnect"
 	WS_EVENT_REFRESH        = "refresh"
 	SETTING_BUTTONS_TEAM    = "team"
-	SETTING_BUTTONS_CHANNEL = "channel"
-	SETTING_BUTTONS_OFF     = "off"
```
marianunez: These are also defined in the webapp here : https://github.com/mattermost/mattermost-plugin-github/blob/f67b80b3daca4b99b036b59759148e241ac44c2a/webapp/src/constants/index.js#L6-L7

We should remove them there to be consistent.

Issue #124, opened 2019-09-24
Improve "Attach to Issue" function


aaronrothschild:
This was just completed by @niklabh https://github.com/mattermost/mattermost-plugin-github/pull/120

Comments from @DHaussermann are quoted below.

> @aaronrothschild I do see some opportunities to add to this further that would improve the experience. Let me know of any of this would be worth adding and I can create separate issues.
> 
> 1. When I attach a comment, I should get feedback that this was done successfully (Similar integrations post a reply with such a message and a link but, a simpler way to provide feedback wold work as well)

Yes, I've had a few instances where the confirmation is re-assuring - certainly during the first experience.  I agree with @DHaussermann , the utility of having the link referenced within the conversation is pretty critical to the long-term user experience.  Over time, without the reference to the attached issue, a critical piece of development conversations can get lost.  it also serves the dual purpose of confirming that it was successful in attaching to the issue.  I'm 4/5 we should add this before merge.

> 2. In the search results, ideally we would return the issue number in addition to the title to cover cases where there titles are similar. Though the issue numbers could overlap across multiple repos within the same list.

This request has come up from a few different customers that use the equivalent function in the Jira plugin.  Displaying the issue numbers clears ambiquity and lets the user be confident they chose the right option/value.  Not sure how much effort is involved.

> 3. This design is limited to adding comments where the user is the assignee. There would also be value in adding comments to other issues as well. Would we need a more complex design to handle cases where you select an issue and don'y have permissions to comment?

Hmmm.  Not sure about this one, could we even accomodate a scenario where we couldn't post as the user (b/c permissions)?

hector2:
Hello! Can i work on this issue?

aaronrothschild:
@hector2 Absolutely, let me know if you have any questions at all!

Removing "Up for Grabs" label for you.

hector2:
Hello @aaronrothschild  !

I managed to check if the issue has locked the conversation. However, there is a problem checking if the user is blocked from the organization.

I created an organization called "GabuTest", and the owner is "gabu9". In this organization, I blocked myself (hector2).

If I check if the user is blocked (using gabu9 credentials):

curl -u "gabu9" -H "Accept: application/vnd.github.giant-sentry-fist-preview+json" https://api.github.com/orgs/GabuTest/blocks/hector2

The response is 204, it means it is blocked.

However, if I do:
[...]

PR #162, opened 2019-11-30
GH-124 Improve "Attach to Issue" function 

hector2:
### Summary 
- Feedback in mattermost channel when an issue comment is created.
- Permalink in issue comment. 
- Showing org/owner, repo and issue number in the search list.
- Locked issues are shown but can't be selected.
- More precise feedback when the usser doesn't have enough permissions (maybe banned or blocked from org) or the repo has turned private.
- List issues in which the user is not the assignee.

### Issue link

Fixes https://github.com/mattermost/mattermost-plugin-github/issues/124

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -9,9 +9,9 @@ import (
 	"strings"
 	"time"
 
-	"github.com/mattermost/mattermost-server/mlog"
-	"github.com/mattermost/mattermost-server/model"
-	"github.com/mattermost/mattermost-server/plugin"
+	"github.com/mattermost/mattermost-server/v5/mlog"
+	"github.com/mattermost/mattermost-server/v5/model"
+	"github.com/mattermost/mattermost-server/v5/plugin"
 
 	"github.com/google/go-github/v25/github"
 	"golang.org/x/oauth2"
--- a/server/api.go
+++ b/server/api.go
@@ -195,9 +195,11 @@ func (p *Plugin) completeConnectUserToGitHub(w http.ResponseWriter, r *http.Requ
 	p.API.PublishWebSocketEvent(
 		WS_EVENT_CONNECT,
 		map[string]interface{}{
-			"connected":        true,
-			"github_username":  userInfo.GitHubUsername,
-			"github_client_id": config.GitHubOAuthClientID,
+			"connected":           true,
+			"github_username":     userInfo.GitHubUsername,
+			"github_client_id":    config.GitHubOAuthClientID,
+			"enterprise_base_url": config.EnterpriseBaseURL,
+			"organization":        config.GitHubOrg,
 		},
 		&model.WebsocketBroadcast{UserId: userID},
 	)
--- a/server/api.go
+++ b/server/api.go
@@ -533,6 +533,30 @@ func getPermaLink(siteUrl string, postId string, currentTeam string) string {
 	return fmt.Sprintf("%v/%v/pl/%v", siteUrl, currentTeam, postId)
 }
 
+func getFailReason(code int, repo string, username string) string {
+	cause := ""
+	switch code {
+	case http.StatusInternalServerError:
+		cause = "Internal server error"
+		break
+	case http.StatusBadRequest:
+		cause = "Bad request"
+		break
+	case http.StatusNotFound:
+		cause = fmt.Sprintf("Sorry, either you don't have access to the repo %s with the user %s or it is no longer available", repo, username)
+		break
+	case http.StatusUnauthorized:
+		cause = fmt.Sprintf("Sorry, your user %s is unauthorized to do this action", username)
+		break
+	case http.StatusForbidden:
+		cause = fmt.Sprintf("Sorry, you don't have enough permissions to comment in the repo %s with the user %s", repo, username)
+		break
+	default:
+		cause = fmt.Sprintf("Unknown status code %d", code)
+	}
+	return cause
+}
+
 func (p *Plugin) createIssueComment(w http.ResponseWriter, r *http.Request) {
 	if r.Method != http.MethodPost {
 		http.Error(w, fmt.Sprintf("Request: %s is not allowed, must be POST", r.Method), http.StatusMethodNotAllowed)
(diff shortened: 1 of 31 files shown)


server/api.go
```diff
@@ -529,6 +529,34 @@ func (p *Plugin) searchIssues(w http.ResponseWriter, r *http.Request) {
 	w.Write(resp)
 }
 
+func getPermaLink(siteUrl string, postId string, currentTeam string) string {
+	return fmt.Sprintf("%v/%v/pl/%v", siteUrl, currentTeam, postId)
+}
+
+func getFailReason(code int, repo string, username string) string {
```
aaronrothschild: Thanks for being thorough and finding the edge cases on this @hector2

hector2: No problem Aaron :D

DHaussermann:
@aaronrothschild I will deploy this to a test server and ping you credentials.

jasonblais:
@aaronrothschild / @DHaussermann quick reminder to help with reviews after the holidays. Thank you!

server/api.go
```diff
@@ -582,17 +620,65 @@ func (p *Plugin) createIssueComment(w http.ResponseWriter, r *http.Request) {
...
+	req.Comment = permalinkMessage + req.Comment
 	comment := &github.IssueComment{
 		Body: &req.Comment,
 	}
 
-	result, _, err := githubClient.Issues.CreateComment(ctx, req.Owner, req.Repo, req.Number, comment)
-	if err != nil {
-		mlog.Error(err.Error())
-	}
+	if result, rawResponse, err := githubClient.Issues.CreateComment(ctx, req.Owner, req.Repo, req.Number, comment); err != nil {
```
crspeller: Generally we like to not use the else block like this. See: https://golang.org/doc/effective_go.html#if

hector2: Done without else block!

DHaussermann:
Hi @hector2, Sorry for the delay on this. Thanks for the round of improvements. The feed back that the comment has been added really makes this nicer to use! 

I've done some testing on this and have a couple questions:
 
1. Was there previously any type of restriction on which repos I could search in for issues or, did it just seem that way because I could only search for issues where I was the assignee? Currently, it seems like you can search across any public repo. Is that correct?

2. Would you be able to make two small improvements to the post menu. The option reads `Github` instead of `GitHub` and the text seems to be top justified instead of middle justified. This is not directly related to your PR so, if not, I'd be happy to create a separate issue.

hector2:
Hey @DHaussermann !

1 - Yes, before these changes you could only list the issues in which the user it's the asignee. I did that because I though that was desired, but maybe I'm wrong, let me know! :D

-> his design is limited to adding comments where the user is the assignee. There would also be value in adding comments to other issues as well. Would we need a more complex design to handle cases where you select an issue and don't have permissions to comment?

2 - What do you mean by "the text seems top justified"? I looked at the post menu but it seemed fine. The options are aligned to the left. 

image

And yes I can make those improvements of course!

DHaussermann:
1. @aaronrothschild do you have any thoughts on this? The current design certainly offers the most functionality but, the draw back is that it's a bit confusing since your search can return issues from countless repositories.

2. @hector2 This is a bit easier to notice when compared to other plugins like Jira. The text is not vertically centered to line up with the middle of the icon. It sits a bit high such that the top of the text lines up with the top of the icon.
Icon-text

hector2:
@DHaussermann hey, I think the alignment problem is related to https://github.com/mattermost/mattermost-plugin-jira/pull/407 . Let me check if I can do the same in this case.

DHaussermann:
@hector2 Thanks for fixing the alignment issues! I had a couple follow up questions about error cases...

**A**. When the user does not have permission to make the comment, it's great that we return allot of info but, the error looks a bit cryptic. 
GitHubAttch
Can we trim off the beginning and about `id` and `message` and loose the braces? Or is there limited control over what goes up to the UI?
**B**. When a conversation becomes locked in GitHub I noticed that the issue becomes non-selectable in the UI. The user's pointer changes to a crossed out circle. Just curious if there is a reason this is different rather than have the user select the option and see an error in the modal         

@aaronrothschild to clarify my comment from above, the UX of opening the search and seeing issues from countless different projects to search for would only be the case if the plugin is not locked to an organization or a users' work space. When locked it only returns issues from within these repos in the issue search results. Please confirm you're okay with this design. 
Also, both of the issues above a very minor, maybe I should simply create a follow-up issue for the Error UX and not hold up this PR.

aaronrothschild:
@asaadmahmood Any thoughts on this?   

I think ideally we can "scope" the search to only include repositories they own or belong to at first.  It becomes pretty unwieldy.  Perhaps a radio button select:  [ ] My/MyOrg's Repositories [ ] All Public Github Repos.  OR alternatively should we break the search into two parts: ask for the repo keywords first, then allow search by keywords through issues in that repo. 

 You start to realize how commonly used some words are in issues and it becomes almost impossible to find the relevant issue unless you add the keyword of the gituhub repo name, then it becomes more manageable.  I think the UI should encourage that behavior perhaps - by splitting the search box into two fields, one for the repo name and one for the keywords being searched within that repo.  I'm 2/5 though.  

cc @DHaussermann ? 

I'm not suggesting we make a heavy interface like this, but I like how Gitbook's Github Repo selection allows me to choose based on which repos are related to me:


Just an example when you use some words that are unusual it reduces the results set, but I had a hard time finding some of the issues I was expecting to be able to find easily.

aaronrothschild:
> @aaronrothschild to clarify my comment from above, the UX of opening the search and seeing issues from countless different projects to search for would only be the case if the plugin is not locked to an organization or a users' work space. When locked it only returns issues from within these repos in the issue search results. Please confirm you're okay with this design.
> Also, both of the issues above a very minor, maybe I should simply create a follow-up issue for the Error UX and not hold up this PR.

OK, that's great and ideal then.  Ignore my comment above, I had half-written it earlier today and just finished typing it when I saw this.

DHaussermann:
Okay Thanks @aaronrothschild I'd be curious how common it is for people to use the plugin not locked to an organization. There may be a worthwhile story in the future IF we get feedback that it would be valuable to restrict attachment functionality.

webapp/src/components/github_issue_selector.jsx
```diff
@@ -41,7 +41,15 @@ export default class GithubIssueSelector extends PureComponent {
         const textEncoded = encodeURIComponent(text.trim().replace(/"/g, '\\"'));
 
         return Client.searchIssues(textEncoded).then((data) => {
-            return Array.isArray(data) ? data.map((item) => ({value: item, label: item.title})) : [];
+            return Array.isArray(data) ?
```
mickmister:
```suggestion
           if (!Array.isArray(data)) {
               return [];
           }
           return data.map(...)
```

hector2: I like this, it's more readable! thanks @mickmister

Issue #139, opened 2019-10-09
octopus icon randomly changes where it points to


wwalker:
GitHub (github - 0.9.2)
GitHub plugin for Mattermost.

We are running on GithubEnterprise.

We have Enterprise Base URL and Enterprise Upload URL set to https://github.ourdomain.com

the octocat in the left bottom, sometimes points to https://github.com/settings/connections/applications/09a864219be324547cea1
and other times points to https://github.ourdomain.com/settings/connections/applications/09a864219be324547cea1

I can't get it to change, it just does what it wants.  Currently, all the icons down there point to github.com and therefore don't work.

wwalker:
![gh](https://user-images.githubusercontent.com/23478/66520749-64090580-eaaf-11e9-82e8-3b7cff77589d.png)

wwalker:
Also, I have tested this in Google Chrome Version 77.0.3865.90 (Official Build) (64-bit) and Mozilla Firefox 69.0.1 both on Fedora 30

wwalker:
And now they are pointing again where they should. (no config change, no restart, nothing done to change the behavior
![ghe](https://user-images.githubusercontent.com/23478/66522685-a0d6fb80-eab3-11e9-8d6b-34af6d10d0b1.png)

hanzei:
This is likely a duplicate of https://github.com/mattermost/mattermost-plugin-github/issues/113 and  https://github.com/mattermost/mattermost-plugin-gitlab/issues/74

larkox:
May I take this one? I just uploaded a solution for gitlab and I am pretty sure the solution here is the same.

PR #164, opened 2019-12-02
[GH-139] octopus icon randomly changes where it points to

larkox:
Summary:
Fix "octopus icon randomly changes where it points to"

Issue link:
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/139

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -195,9 +195,11 @@ func (p *Plugin) completeConnectUserToGitHub(w http.ResponseWriter, r *http.Requ
 	p.API.PublishWebSocketEvent(
 		WS_EVENT_CONNECT,
 		map[string]interface{}{
-			"connected":        true,
-			"github_username":  userInfo.GitHubUsername,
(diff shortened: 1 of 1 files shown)


DHaussermann:
This is performing as expected when no Enterprise Base URL is configure but, I'm unsure of how I can validate this when there is an Enterprise Base URL in place. 

I can add a token URL to the config and hit refresh to see the icon pointing to it but I can't validate that it works while having a connected user. @crspeller does Mattermost have an enterprise URL I may not be aware of?

crspeller:
@jwilander Did we get something to test the GitHub enterprise integration?

jwilander:
I'm part of the GitHub Enterprise developer program which is how I tested this stuff in the past. I don't have that instance running anymore though unfortunately

It's pretty easy to get into the developer program and spin up an instance though. Might be a good idea to have a "permanent" instance for testing the plugin

PR #170, opened 2019-12-22
[GH-127] Improve RHS UI and content

larkox:
#### Summary

Improve RHS UI and content
#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-github/issues/127

#### Screenshots
image
image
image
image


larkox:
The functionality is practically complete, but there are a few concerns.

First of all, I would like to know if there are any standard styles you want to use for several parts of the code.

Also, to retrieve status and reviews imply making several more calls, which might become a problem due to the rates from the API. We are talking that retrieving the PRs now implies: 1 call for the list of PRs + 3 calls for each PR.

Finally, there are some details that I am not sure whether they have been done in the most desirable way. Due to how I was able to check the reviews, we get the "Changes requested" message whenever someone asked for changes, even if we have clicked the "Re-request review" button. Is this the expected behaviour?

marianunez:
> The functionality is practically complete, but there are a few concerns.

Thanks @larkox! Sorry for the delay response as some of us where away for the holidays.

> The functionality is practically complete, but there are a few concerns.
> 
> First of all, I would like to know if there are any standard styles you want to use for several parts of the code.

Could you provide some screenshots of your changes in the PR description that would help get an initial UX review? We will also take a look at the code and provide any styles suggestions there are necessary. 
 
> Also, to retrieve status and reviews imply making several more calls, which might become a problem due to the rates from the API. We are talking that retrieving the PRs now implies: 1 call for the list of PRs + 3 calls for each PR.

Agreed that it seems too many extra calls to do each time we get all PRs. Two possible suggestions:

- Given that `getYourPrs` is called each time the plugin data is refreshed to update the numbers in the left hand sidebar, I suggest we decouple the extra information to a separate plugin call and only get that extra information when the RHS UI actually needs it. This way the extra calls will only occur when the UI requests it and not each time the plugin is updated. 
- Regarding the PR status specifically (x or check next to the PR), not sure if it provides enough value at to merit another API call. Thoughts @mickmister @asaadmahmood ?

> Finally, there are some details that I am not sure whether they have been done in the most desirable way. Due to how I was able to check the reviews, we get the "Changes requested" message whenever someone asked for changes, even if we have clicked the "Re-request review" button. Is this the expected behaviour?

I believe this is the way Github handles the review state correct?

larkox:
Thanks @marianunez !

I uploaded a new version of this, with pr fetching and extra info fetching decoupled. Whenever you tell me what information is needed and what not, I will remove that part. As a small remainder:
- Fetching reviewers: 1 more call per PR
- Fetching status: 2 more calls per PR

I also have been taking a look into it, and using GraphQL (github API v4, right now we are using v3) might allow us to put all this in a single call. If you want, we can continue this PR as it is right now, and at some point consider migrating the api calls to GraphQL and use all the potential they have.

Regarding UX, I added one screenshot on how it looks right now. My point is that if we have already some CSS classes created that will work for this, let me know which ones, so I use them directly. If they are not created, let me know the definition (colors, sizes, theme parameters...) so I can implement them.

Thanks!

asaadmahmood:
@larkox What exactly does the x button does?
Does it clear that item? I don't have much context on this PR.

larkox:
Hello @asaadmahmood !

The X should be red, and is the same X as in this screens (taken from github directly):
image

It states the status of the PR. Whether it failed any check, or passed all the checks or it is pending something. It can be also a green tick:
image
or an orange dot (I have no image of that one at the moment).

This image is used in several other places with similar meaning. For example:
image
image

Is it clear? Or do you need more context?

asaadmahmood:
@larkox Got it, can you setup a test server with that a PR that has that x icon on it.
I can make the changes live, and tell you which css to apply.
I don't have Github configured locally so might not be able to set it up.

mickmister:
> I also have been taking a look into it, and using GraphQL (github API v4, right now we are using v3) might allow us to put all this in a single call.

@larkox That sounds like an excellent improvement! Thank you for pointing out this option.

When we are talking about API requests per PR, are we referring to API to calls to the MM server, or API calls to the GitHub server? If it is the former, we can consolidate some of the API requests into endpoints that return several pieces of data. If it is the latter, GraphQL seems like a good way to get everything we need from GitHub in (ideally) one request.

mickmister:
@larkox Here's a test server you can use:

https://github-plugin-pull-170.test.mattermost.cloud

Username: `sysadmin`
Password: `Sys@dmin123`

larkox:
@mickmister Yes, it is the second case. 3 different calls to github. I haven't explored enough the GraphQL api yet, but hopefully it can make things easier.

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -35,29 +37,140 @@ function GithubItems(props) {
...
+        }
+
+        let status = '';
+
+        // Status images pasted directly from GitHub. Change to our own version when styles are decided.
+        if (item.status) {
+            switch (item.status) {
+            case 'success':
+                status = (
+                    <svg
```
mickmister: @larkox Can we move these svg's to a separate file with named exports?

@marianunez Does this make sense or do you think they should each have their own file? I think it clutters this file having them laid out like they are currently.

marianunez: +1 on their own file.

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -35,29 +37,140 @@ function GithubItems(props) {
...
         }
 
+        let milestone = '';
+        if (item.milestone) {
+            milestone = (<span><i className='fas fa-bullseye'/>{item.milestone.title}</span>);
+        }
+
+        let reviewers = '';
+        if (item.reviewers && item.reviewers > 0) {
+            reviewers = (<span>{item.reviewers} pending reviews.</span>);
```
marianunez: I believe these are the amount of reviewers requested regardless if pending or not. Not sure if there is any value on providing the total reviewers. 

Thoughts @mickmister ?

mickmister: Could we have a `3 out of 4 reviews complete` display or something of the sort?

larkox:
@asaadmahmood The test server that @mickmister provided already has one version of my changes. You have several cases. If you want to try any other case, please let me know. Here is the info for the server again:

https://github-plugin-pull-170.test.mattermost.cloud

Username: sysadmin
Password: Sys@dmin123

@mickmister In order to let Asaad try this, I created a mockClient for the webapp part. I am not sure whether this will be useful on the future, but I can clean it up and have upload it also if you find it useful.

webapp/src/components/sidebar_right/github_items.jsx
larkox: This is a mistake on my part. The reviewers value comes from "requested_reviewers". As far as I have seen, a requested reviewer is marked as that since the it is requested for review until it makes its first review. After that, even if you request a rereview, it will not appear on "requested_reviewers" anymore.

That being said... reviewers is indeed the pending reviews. Also, because of that, my first question about how to deal with the reviews. I can "do the maths" and leave it as `3 out of 4 reviews complete`, or I could go even overkill with something as:
```
- 1 approved review.
- 2 changes requested reviews.
- 1 pending review.
- 1 out of 4 reviews complete.
```

What I cannot do (as far as I have seen at least) is distinguish between a Changes Requested review and a rerequested review.

All that being said, what do you expect to be the most useful way to give this information?

mickmister:
> @mickmister In order to let Asaad try this, I created a mockClient for the webapp part.

He is testing the view so that sounds reasonable. Could you post a summary of what the mock client is doing in a comment? Thanks @larkox!

asaadmahmood:
@larkox Here's how it should look:
Screenshot 2020-01-09 at 11 08 39 PM

Here's the css:
```
    top: 3px;
    position: relative;
    left: 6px;
    fill: #2b9643;
    height: 18px;
    display: inline-flex;
    align-items: center;
```

The fill for the check, the cross, and the circle will be different:
`#2b9643` is for the check.
`#c59e17` is for the circle.
`#c11b28` is for the cross.

Wrap the svgs in a `<div>` tag, and give that div tag this css.
Like this:
Screenshot 2020-01-09 at 11 08 59 PM

larkox:
@mickmister It is really simple. It just substitute all the calls to the server on client.js by mock functions that just return the values you want. For example, client.getYourPrs(), instead of doing a get call to the server side, it just returns a list of PRs done by hand. This way we can test with controlled data that the application behave the way we want with the data we expect to receive.

Another great thing about this is that the client.getConnected always returns true, so there is no need to set up everything on github, or connect to github. You just put enable the plugin, and you are "connected".

At the moment most of the functions just return empty lists, since I do not need more in my example, but as more test are made, it can be filled up.

webapp/src/components/sidebar_right/github_items.jsx
mickmister: @larkox I think saying `1 out of 4 reviews complete` and a small indicator to express that at least one reviewer currently has requested changes (just like the red icon that shows up in GitHub) would be sufficient. I think this solution is both compact and informative.

larkox:
All previous commented errors should be solved on the latest commit. Latest version is also uploaded to the test server.

Still missing some small things:
- @asaadmahmood Do you like the current style? I feel the "x out of y reviews complete" text is a bit bland, and could use a bit more of formatting, and all being black is also a bit bland and could use more color, similar to how github does it, but I think I should not be the person deciding the style.
- @marianunez I tried using for milestone the bullseye icon, but it does not load correctly. I am using the notation `<i class="fas fa-bullseye"></i>` which worked for the branch icon, but for this one is not working. Any idea or alternative for this?
- @mickmister I still have not ported anything to GraphQL (I think that may go better on a different PR). Do you want me to port this to GraphQL on this PR? If not, do you want to keep this PR on hold until that is done? Or do you prefer to remove some of the functionality (specially the status icon) to reduce the number of calls?

I think once all that is clear, I can remove the draft status of this PR.

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -35,29 +49,97 @@ function GithubItems(props) {
...
+                </div>
                 <GithubLabels labels={item.labels}/>
                 <div
                     className='light'
                     style={style.subtitle}
                 >
-                    {userName ? 'Created by ' + userName + ' ' : ''}
-                    {'at ' + repoName + '.'}
+                    {'Opened ' + formatTimeSince(item.created_at) + ' ago'}
+                    {userName ? ' by ' + userName : '.'}
```
mickmister:
```suggestion
                    {userName && ' by ' + userName}{'.'}
```

I think the punctuation should be consistent here whether there is a username or not.

mickmister:
> Do you want me to port this to GraphQL on this PR?

@larkox I believe this PR is out of scope to introduce the GraphQL portion. A separate PR to improve performance would be very much appreciated :slightly_smiling_face:

marianunez:
> * @asaadmahmood Do you like the current style? I feel the "x out of y reviews complete" text is a bit bland, and could use a bit more of formatting, and all being black is also a bit bland and could use more color, similar to how github does it, but I think I should not be the person deciding the style.

I agree. It also seems the font color and size is off compared to the other lines. Myabe @asaadmahmood can provide some css styling there too.

Also, the original design I believe the PR number had a link color like the screenshot below:

marianunez:
> * @marianunez I tried using for milestone the bullseye icon, but it does not load correctly. I am using the notation `<i class="fas fa-bullseye"></i>` which worked for the branch icon, but for this one is not working. Any idea or alternative for this?

Not sure why it's not coming up. @asaadmahmood could it be we have an older version of font awesome (v4.7) in the webapp? Do you have an alternative icon to use instead?

asaadmahmood:
@marianunez @larkox Yup, don't like it, it should be similar to the light text color that says `Opened an hour ago` in the screenshot that Maria shared but perhaps semibold.

asaadmahmood:
@marianunez Yup, we are using an older version. But we can just just use `fa fa-bullseye`. instead of `fas`.

server/api.go
```diff
@@ -493,6 +503,114 @@ func (p *Plugin) getYourPrs(w http.ResponseWriter, r *http.Request) {
...
+
+	if info, err := p.getGitHubUserInfo(userID); err != nil {
+		writeAPIError(w, err)
+		return
+	} else {
+		githubClient = p.githubConnect(*info.Token)
+	}
+
+	var prInfo []*PRExtraInfo
+	json.NewDecoder(r.Body).Decode(&prInfo)
```
marianunez: Why is this `PRExtraInfo` struct? Is this not the list of yourPRs that are passed in?

server/api.go
```diff
@@ -493,6 +503,114 @@ func (p *Plugin) getYourPrs(w http.ResponseWriter, r *http.Request) {
...
+	for range prInfo {
+		pr := <-ch
+		prList = append(prList, pr)
+	}
+
+	resp, _ := json.Marshal(prList)
+	w.Write(resp)
+}
+
+func fetchExtraPRInfo(ctx context.Context, client *github.Client, prURL string, prNumber int, ch chan *PRExtraInfo) {
```
marianunez:
```suggestion
func fetchPRDetails(ctx context.Context, client *github.Client, prURL string, prNumber int, ch chan *PRExtraInfo) {
```

webapp/src/actions/index.js
```diff
@@ -86,6 +86,29 @@ export function getYourPrs() {
     };
 }
 
+export function getYourPrsExtraInfo(prList) {
```
marianunez:
```suggestion
export function getYourPrsDetails(prList) {
```

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -35,29 +49,97 @@ function GithubItems(props) {
...
+                } else {
+                    reviewName = 'reviews';
+                }
+                reviews = (<span>{approved} out of {totalReviewers} {reviewName} complete.</span>);
+            }
+
+            if (changesRequested > 0) {
+                changes = (<div style={{...iconStyle, fill: '#c11b28'}}><DotIcon/></div>);
+            }
+        }
```
marianunez: Let's extract this logic to a function here for readability

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -35,29 +49,97 @@ function GithubItems(props) {
...
                 style={style.container}
             >
                 <div>
                     <strong>
-                        {title}
+                        {title}{status}
                     </strong>
                 </div>
+                <div>
+                    <strong><i className='fa fa-code-fork'/> #{item.number}</strong> <span className='light'>{repoName}</span>
```
marianunez: Repo name should be in parenthesis. See original design screenshot.

webapp/src/images/icons/cross.jsx
```diff
@@ -0,0 +1,20 @@
...
+                role='img'
+            >
+                <path
+                    fillRule='evenodd'
+                    d='M7.48 8l3.75 3.75-1.48 1.48L6 9.48l-3.75 3.75-1.48-1.48L4.52 8 .77 4.25l1.48-1.48L6 6.52l3.75-3.75 1.48 1.48L7.48 8z'
+                />
+            </svg>
+        );
+    }
+}
```
marianunez: Missing new line at end of file here

webapp/src/images/icons/dot.jsx
```diff
@@ -0,0 +1,20 @@
...
+                role='img'
+            >
+                <path
+                    fillRule='evenodd'
+                    d='M0 8c0-2.2 1.8-4 4-4s4 1.8 4 4-1.8 4-4 4-4-1.8-4-4z'
+                />
+            </svg>
+        );
+    }
+}
```
marianunez: Missing new line at end of file here

webapp/src/images/icons/tick.jsx
```diff
@@ -0,0 +1,20 @@
...
+                role='img'
+            >
+                <path
+                    fillRule='evenodd'
+                    d='M12 5l-8 8-4-4 1.5-1.5L4 10l6.5-6.5L12 5z'
+                />
+            </svg>
+        );
+    }
+}
```
marianunez: Missing new line at end of file here

webapp/src/utils/date_utils.js
```diff
@@ -33,3 +33,20 @@ export function formatDate(date, useMilitaryTime = false) {
...
+    if (minutesSince < 60) {
+        return minutesSince + 'm';
+    }
+    const hoursSince = Math.trunc(minutesSince / 60);
+    if (hoursSince < 24) {
+        return hoursSince + 'h';
+    }
+    const daysSince = Math.trunc(hoursSince / 24);
+    return daysSince + 'd';
+}
```
marianunez: Missing new line at end of file here

webapp/src/components/sidebar_right/sidebar_right.jsx
```diff
@@ -57,7 +69,25 @@ export default class SidebarRight extends React.PureComponent {
...
+                    return pr;
+                }
+
+                return {
+                    ...pr,
+                    status: extra.status,
+                    reviewers: extra.reviewers,
+                    reviews: extra.reviews,
+                };
+            });
```
marianunez: Could we move this mapping logic to the reducer and have it ready when you map `yourPrsExtraInfo` for the state to the component? 

Also, these details would also be needed for the case of PR `RHSStates.REVIEWS` below. So this logic can be reused there for that case too.

server/api.go
larkox: The naming of the variables is ugly. I will change that.
In this function, yourPRs are passed as Post Form, since it is a http call. In order not to pass the full posts we only pass the significant information (repository_url and number). When we are unmarshalling, we use the same type in order to avoid creating more types.

larkox:
All requested changes should be addressed, latest mock version uploaded to the server, and screenshot of current look updated in the description.

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -42,22 +133,31 @@ function GithubItems(props) {
             >
                 <div>
                     <strong>
-                        {title}
+                        {title}{status}
                     </strong>
                 </div>
+                <div>
+                    {number} <span className='light'>{repoName}</span>
```
marianunez: Could we add here parenthesis to the repo name as the original design below?

66054464-05b5b300-e502-11e9-8f70-cac639ef5322

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -33,6 +51,79 @@ function GithubItems(props) {
...
+
+            const totalReviewers = item.reviewers + item.reviews.length;
+            if (totalReviewers > 0) {
+                let reviewName;
+                if (totalReviewers === 1) {
+                    reviewName = 'review';
+                } else {
+                    reviewName = 'reviews';
+                }
+                reviews = (<span>{approved} out of {totalReviewers} {reviewName} complete.</span>);
```
marianunez: Can we move this to a function `getReviewText` for readibility?

marianunez:
@larkox Your changes are looking great! Adding some comments regarding final UX details:

72522676-8a80ce80-385e-11ea-8d75-1373b1dd79b3

1. We should add a tad of padding between the when opened date and the milestone icon and text. Also, the icon doesn't look that great at that scale. Going to see if there are any better options for that. Thoughts @asaadmahmood?

2. What is the red dot next to the amount of reviews text (blue arrow below)? Couldn't see where in the code that is being handled.
3. I believe that 3 lines underneath the labels is getting a bit overcrowded. Also, not sure how much value the changes text bring (green box below) given that we do have the amount of reviews complete and status. I think we should remove it in favor of a simpler UI. Thoughts @mickmister?
4. Could you add a screenshot of how the `Unread Messages` case looks like now?

larkox:
Thanks for the comments @marianunez 

> 1. We should add a tad of padding between the when opened date and the milestone icon and text. Also, the icon doesn't look that great at that scale. Going to see if there are any better options for that. Thoughts @asaadmahmood?

I added some padding with style between the created at and the milestone. I will be waiting for the new icon to include it. Also, if you want to tweak the distance, I can change it.

> 2. What is the red dot next to the amount of reviews text (blue arrow below)? Couldn't see where in the code that is being handled.

This denotes when you have changes requested. This was proposed by @mickmister a few comments before, but I agree that it feels weird. We could change that.

> 3. I believe that 3 lines underneath the labels is getting a bit overcrowded. Also, not sure how much value the changes text bring (green box below) given that we do have the amount of reviews complete and status. I think we should remove it in favor of a simpler UI. Thoughts @mickmister?

This was my mistake populating the mockClient. The "reason" field is not set for PRs or Issues, only for messages. Therefore, they will not look as on the screenshot. The new screenshots are updated.

> 4. Could you add a screenshot of how the Unread Messages case looks like now?

Added

I also updated the mock client with data more similar to real data. Nevertheless, I have no requested reviews on my github account, so I am not sure 100% how the real data on those is.

asaadmahmood:
@marianunez @larkox The padding between the opened date and the milestone icon doesn't look bad, we can add 2px more in between them however, but not much I think.

We can also decrease the spacing between the `Opened` line and the `0 our of 2` line to be consistent with the spacing between the label and the `Opened` line.

Screenshot 2020-01-18 at 7 12 25 PM

larkox:
@asaadmahmood The distance between the 'opened at' and 'reviews' line should be now the same. There is still one case where it is not the same, that is when the "changes requested dot" is visible, but since that might change, we can wait for that to be decided.

marianunez:
> > 1. What is the red dot next to the amount of reviews text (blue arrow below)? Couldn't see where in the code that is being handled.
> 
> This denotes when you have changes requested. This was proposed by @mickmister a few comments before, but I agree that it feels weird. We could change that.

I'm 2/5 a dot at the top for status and then a second one on the bottom seems a bit overloaded. Thoughts @mickmister?

> I will be waiting for the new icon to include it.

@asaadmahmood Do you have the svg for the icon used in your original design that we can use here? I can't seem to find a free one in font awesome that would be good for that milestone one.

asaadmahmood:
Here it is @marianunez:
[Sign Icon _ Font Awesome-31.svg.zip](https://github.com/mattermost/mattermost-plugin-github/files/4120170/Sign.Icon._.Font.Awesome-31.svg.zip)

mickmister:
> I believe that 3 lines underneath the labels is getting a bit overcrowded. Also, not sure how much value the changes text bring (green box below) given that we do have the amount of reviews complete and status. I think we should remove it in favor of a simpler UI. Thoughts @mickmister?

@marianunez Sorry I missed this question. Removing the text such as "You commented on the thread" sounds good to me.

> I'm 2/5 a dot at the top for status and then a second one on the bottom seems a bit overloaded. Thoughts @mickmister?

What exactly does the dot at the top represent? The one next to the # of reviewers shows that at least one reviewer has pending requested changes, which I think is very useful.

marianunez:
> What exactly does the dot at the top represent? The one next to the # of reviewers shows that at least one reviewer has pending requested changes, which I think is very useful.

The dot in the top is when the build is in process, when completed it's actually a green check mark or red x. 

I'm OK wrapping up this PR with the current design and then iterate later in future improvements after we use it and get other peoples feedback.

larkox:
The dot at top (yellow dot) represents that the status of the latest commit is "pending". In other words, there are still some checks to be done on the commit, for example, the CI has not run the tests yet. The red cross means some checks failed (for example, the tests failed). Finally, the green tick means all checks passed successfully.

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -97,8 +183,78 @@ function GithubLabels(props) {
...
+        if (totalReviewers === 1) {
+            reviewName = 'review';
+        } else {
+            reviewName = 'reviews';
+        }
+        reviews = (<span>{approved} out of {totalReviewers} {reviewName} complete.</span>);
+    }
+
+    if (changesRequested > 0) {
+        changes = (<div style={{...style.icon, fill: '#c11b28'}}><DotIcon/></div>);
```
marianunez: Could we add a tooltip on this icon to indicate what it means? "Changes Requested"

larkox: Done!

mickmister: Is it possible to use the requested changes icon from GitHub? 
image

larkox: I can copy/paste the svg from mattermost site, if that is OK (similar to what I did with the tick, cross and dot icons)

mickmister: Do you mean GitHub? Getting it straight from there sounds fine to me.

larkox: Yeah, sorry. I meant GitHub. Tomorrow I will get that done first thing in the morning :D

larkox: Done
image
I feel the icon is still a bit thin, but apart of that I like the result.

mickmister: The requested changes icon seems like it should be more "in your face", as it represents an action item you should complete. But placing it next to the number of reviews seems to make sense.

@marianunez What do you think about putting it next to the build status icon instead, so it is more visible? I don't want to keep dragging this PR along, but I also want to get it right as we're making this final push.

Also, if the PR's name is much longer, I wonder how the RHS display looks in general.

marianunez: 0/5 I think I prefer on the bottom, two icons next to each other may seem confusing. 
However, deferring to @asaadmahmood expertise on this.

larkox: @mickmister Here I leave you a screenshot with real data. Sadly, I have no PR with changes requested at the moment, so the icon will not appear.

image

server/api.go
```diff
@@ -495,6 +505,116 @@ func (p *Plugin) getYourPrs(w http.ResponseWriter, r *http.Request) {
...
+
+	repoOwner, repoName := getRepoOwnerAndNameFromURL(prURL)
+
+	chReviews := make(chan []*github.PullRequestReview)
+
+	go fetchReviews(ctx, client, repoOwner, repoName, prNumber, chReviews)
+
+	prInfo, _, err := client.PullRequests.Get(ctx, repoOwner, repoName, prNumber)
+
+	if err != nil {
```
mickmister: I think we should delay the call to `fetchReviews` until after we know this `err` is nil.

server/api.go
```diff
@@ -495,6 +505,116 @@ func (p *Plugin) getYourPrs(w http.ResponseWriter, r *http.Request) {
...
+
+	var prDetails []*PRDetails
+
+	ch := make(chan *PRDetails)
+	for _, pr := range prList {
+		go fetchPRDetails(ctx, githubClient, pr.URL, pr.Number, ch)
+	}
+
+	for range prList {
+		pr := <-ch
```
mickmister: @lieut-data Do you mind taking a look at the the flow of goroutines and channels used in this PR? I'm concerned about the risk of this line blocking indefinitely if a given goroutine does not successfully send to the channel.

lieut-data: Happy to help! Taking a look now :)

server/api.go
```diff
@@ -495,6 +505,116 @@ func (p *Plugin) getYourPrs(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	var prList []*PRDetails
+	json.NewDecoder(r.Body).Decode(&prList)
+
+	var prDetails []*PRDetails
+
+	ch := make(chan *PRDetails)
+	for _, pr := range prList {
+		go fetchPRDetails(ctx, githubClient, pr.URL, pr.Number, ch)
```
lieut-data: To clarify the flow of data here, I propose that instead of passing a channel, just have `fetchPRDetails` return `(*PRDetails, error)`, and then coordinate here instead. This decouples `fetchPRDetails` from how a caller wants to use it,  and simplifies the return conditions of that code. 

On this side, the code then might look something like:

```go
prDetails = make([]*PRDetails, len(prList))
var wg sync.WaitGroup
for i, pr := range prList {
    i := i
    pr := pr
    wg.Add(1)
    go func() {
        defer wg.Done()
        prDetail, err := fetchPRDetails(ctx, githubClient, pr.URL, pr.Number)
        if err != nil {
            // log error ...
            prDetails[i] = ... // default value?
            return
        }
        
        prDetails[i] = prDetail
    }()
}
wg.Wait()

resp, _ := json.Marshal(prDetails)
w.Write(resp)
```

server/api.go
```diff
@@ -495,6 +505,116 @@ func (p *Plugin) getYourPrs(w http.ResponseWriter, r *http.Request) {
...
+func fetchPRDetails(ctx context.Context, client *github.Client, prURL string, prNumber int, ch chan *PRDetails) {
+	status := ""
+	requestedReviewers := []*string{}
+	var reviewsList []*github.PullRequestReview = nil
+
+	repoOwner, repoName := getRepoOwnerAndNameFromURL(prURL)
+
+	chReviews := make(chan []*github.PullRequestReview)
+
+	go fetchReviews(ctx, client, repoOwner, repoName, prNumber, chReviews)
```
lieut-data: Similar to the above, `fetchReviews` should just return the reviews list: there's no need for a channel at all in this case. Use the goroutine for parallelism, but just write directly a local variable and coordinate with a `sync.WaitGroup`.

server/api.go
```diff
@@ -495,6 +505,116 @@ func (p *Plugin) getYourPrs(w http.ResponseWriter, r *http.Request) {
...
+			Reviews:            reviewsList,
+		}
+		return
+	}
+
+	for _, v := range prInfo.RequestedReviewers {
+		requestedReviewers = append(requestedReviewers, v.Login)
+	}
+
+	statuses, _, err := client.Repositories.GetCombinedStatus(ctx, repoOwner, repoName, prInfo.GetHead().GetSHA(), nil)
```
lieut-data: As an aside, I agree with your concerns, @larkox, re: the number of extra fetches here. I wonder if GitHub's GraphQL endpoints might enable fewer, more performant queries.

server/api.go
larkox: The reason behind this is that fetchReviews will take some time, as getting the pull requests. Doing it in parallel would speed things up. Nevertheless, I see now that there might be problems of goroutines stopped on a channel in case this err occurs.

Would you like me to work with this order and solve this problem with the goroutines and channels, or do you prefer changing the order?

server/api.go
larkox: Thanks @lieut-data! I can see the improvement of doing things your way. I had understood that "the way to do things" in go was by using channels ("do not communicate by sharing memory, share memory by communication") but this simplify some things. Nevertheless, I do not see so much improvement on the fetchPRDetails conditions, so I am not sure if I am missing something. Nevertheless, I just pushed my changes for when you have time to take a look at them.

lieut-data: That maxim is indeed idiomatic, but its thrust is to avoid writing to shared memory from multiple processes which can be fraught with race conditions. The specific advice that applies here comes from https://github.com/golang/go/wiki/CodeReviewComments#synchronous-functions:
> Prefer synchronous functions - functions which return their results directly or finish any callbacks or channel ops before returning - over asynchronous ones.

> Synchronous functions keep goroutines localized within a call, making it easier to reason about their lifetimes and avoid leaks and data races. They're also easier to test: the caller can pass an input and check the output without the need for polling or synchronization.

> If callers need more concurrency, they can add it easily by calling the function from a separate goroutine. But it is quite difficult - sometimes impossible - to remove unnecessary concurrency at the caller side.

Thanks for iterating on this -- I'll take a look at your changes soon :)

server/api.go
```diff
@@ -516,98 +517,109 @@ func (p *Plugin) getPrsDetails(w http.ResponseWriter, r *http.Request) {
...
+			mlog.Error(err.Error())
+		}
+	}()
 
 	prInfo, _, err := client.PullRequests.Get(ctx, repoOwner, repoName, prNumber)
 
 	if err != nil {
-		mlog.Error(err.Error())
-		ch <- &PRDetails{
+		wg.Wait()
```
lieut-data: As a matter of style, can we put the `client.PullRequests.Get(ctx, repoOwner, repoName, prNumber)` and the `GetCombinedStatus` into a new goroutine and synchronize with the existing `sync.WaitGroup`? That leaves the "main thread" to focus on invoking the goroutines with a single `wg.Wait()` instead of having to remember to wait here as well.

It should also reduce the number of `return &PRDetails{...}` blocks as well.

server/api.go
```diff
@@ -516,98 +517,109 @@ func (p *Plugin) getPrsDetails(w http.ResponseWriter, r *http.Request) {
...
-		return
+		}, err
 	}
 
 	status = *statuses.State
 
-	reviewsList = <-chReviews
-
-	ch <- &PRDetails{
+	return &PRDetails{
```
lieut-data: In the spirit of reducing the number of `return &PRDetails{...}` blocks, if we rely on the additional goroutine above, then we should be able to reduce this down to just a single return akin to:
```go
	return &PRDetails{
		URL:                prURL,
		Number:             prNumber,
		Status:             status,
		RequestedReviewers: requestedReviewers,
		Reviews:            reviewsList,
	}, err
```

server/api.go
```diff
@@ -556,11 +556,13 @@ func (p *Plugin) getPrsDetails(w http.ResponseWriter, r *http.Request) {
 func fetchPRDetails(ctx context.Context, client *github.Client, prURL string, prNumber int) (*PRDetails, error) {
```
lieut-data: It looks like we never return a non-nil `error` -- should we omit that return value altogether for now?

server/api.go
```diff
@@ -556,11 +556,13 @@ func (p *Plugin) getPrsDetails(w http.ResponseWriter, r *http.Request) {
...
 	status := ""
 	requestedReviewers := []*string{}
-	var reviewsList []*github.PullRequestReview = nil
+	var reviewsList []*github.PullRequestReview = []*github.PullRequestReview{}
```
lieut-data: Would I be correct in assuming you're initializing this so that the resulting JSON serialization returns an empty array `[]` vs the `null` it would otherwise encode when there are no reviews? If that's the case, I'd recommend the more idiomatic:
```
        // Initialize to a non-nil slice to simplify JSON handling semantics.
	reviewsList := []*github.PullRequestReview{}
```

server/api.go
larkox: You are completely right. We suppose the result always comes the best it can. There is nothing to handle if there is any error either way (the call was just logging the error).

server/api.go
larkox: It was more a hunch than anything else. I did not like that sometimes reviewsList was a nil, sometimes was a list. I preferred it to be something more stable.

DHaussermann:
Hi @larkox well done! This UX improvement looks great!

I found 1 very small issue. On the PR Needing Review Section, when the last record at bottom has changes requested from the reviewer(s), the help text for the "+/-" still opens downward so the text "Changes Requested" is not visible.
Changes

Is it possible to fix it such the it opens upward when there is not enough vertical space for it below?

DHaussermann:
Hi @larkox as mentioned above, I see only one minor issue (possibly CSS related) is this is something easy enough to resolve?

larkox:
I have been checking, and there seems that there will not be an "easy" solution.
I tried using the "auto" placement, but it seems it is not supported. Any other solution I can think about requires calculating the position of the objects in relation to the window, which may be troublesome specially because it would have to react to the scroll.

The only easy solution will be always showing it on top. There is always space on top, but may be annoying since will cover part of the text of the issue you are looking at.

DHaussermann:
@asaadmahmood Is this issue something you be able to provide some guidance on?

Alternatively, this PR has been waiting a while and this minor CSS issue is an edge case. @aaronrothschild would it makes sense to merge this and create a follow-up issue for the remain   CSS problem?

asaadmahmood:
@DHaussermann I may have a to look at it. We can take it in, so I can see it live in action.

marianunez:
> Alternatively, this PR has been waiting a while and this minor CSS issue is an edge case. @aaronrothschild would it makes sense to merge this and create a follow-up issue for the remain CSS problem?

I agree that we should open a JIRA ticket to fix this border case and merge this PR. It's a great improvement and would love to see it in community soon. 🎉

hanzei:
Thanks @larkox. This is such an awesome improvement. I can't wait to use it myself.

PR #173, opened 2020-01-08
warn about master pushes

### LINKED ISSUE #146: Alert if push to master?
We would like to have alert in mattermost channel if someone pushes to master (instead of using PRs and normal PR process). I think this cannot be done with current plugin?

zetaab:
#### Summary

This PR notifies channel only if non-whitelisted user will push to master. Whitelist is needed in order to use normal PR processes.

#### Ticket Link

Fixes #146

### DIFF
--- a/server/utils.go
+++ b/server/utils.go
@@ -310,3 +310,21 @@ func getCodeMarkdown(user, repo, repoPath, word, lines string, isTruncated bool)
 	final += "```\n"
 	return final
 }
+
+// stringValue returns string value from pointervalue
+func stringValue(s *string) string {
+	if s == nil {
+		return ""
+	}
+	return *s
+}
+
+// contains is checking does array contain single word
+func contains(array []string, word string) bool {
+	for _, item := range array {
+		if item == word {
+			return true
+		}
+	}
+	return false
+}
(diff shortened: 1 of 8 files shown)


zetaab:
looks like:

Screenshot 2020-01-08 at 11 29 40

hanzei:
Hi @zetaab,

Thanks for the PR. I'm wondering what exactly the use case of this feature is. GitHub already as a setting to disallow pushed to e.g. `master`
Screenshot from 2020-01-08 11-59-52

zetaab:
@hanzei we are using github enterprise and at least for us these rules does not work like should. Also these rules are somehow cleaned sometimes (in updates?). I configured these rules for one repository, but still I can push to master. So I would say that feature is useless.

hanzei:
@zetaab Is it possible that you didn't enable the rules for admins?
Screenshot from 2020-01-08 13-35-36

server/utils.go
```diff
@@ -310,3 +310,21 @@ func getCodeMarkdown(user, repo, repoPath, word, lines string, isTruncated bool)
 	final += "```\n"
 	return final
 }
+
+// stringValue returns string value from pointervalue
```
jwilander:
```suggestion
// stringValue returns the string value from a string pointer
```

server/utils.go
```diff
@@ -310,3 +310,21 @@ func getCodeMarkdown(user, repo, repoPath, word, lines string, isTruncated bool)
...
+func stringValue(s *string) string {
+	if s == nil {
+		return ""
+	}
+	return *s
+}
+
+// contains is checking does array contain single word
```
jwilander:
```suggestion
// contains checks if a string array contains a word
```

hanzei:
Hey @zetaab,

What do you thin about my feedback at https://github.com/mattermost/mattermost-plugin-github/pull/173#issuecomment-572029386? Does this setting help with your use case?

hanzei:
@zetaab given that this PR has been stale for some time and the use case is questionable, I'm going to close this PR. Still, thanks for your PR

Issue #176, opened 2020-01-10
Replace GitHub usernames with Mattermost usernames in notifications


lieut-data:
We already resolve the GitHub user to the Mattermost user when generating the template and posting to Mattermost -- e.g. `@saturninoabril` becomes `@saturnino.abril`, given our knowledge of the linking in the GitHub plugin:
![image](https://user-images.githubusercontent.com/1023171/72159116-15128b00-3392-11ea-965d-3b339220cd61.png)

but let's also consider doing this in the body of notifications as well, so that `@asaadmahmood`, `@marianunez` and `@lieut-data` become `@asaad.mahmood`, `@maria.nunez` and `@jesse.hallam` respectively:
![image](https://user-images.githubusercontent.com/1023171/72159102-0c21b980-3392-11ea-9a02-2aed5841b4d6.png)

Of course, if no known Mattermost user has that GitHub username, we can leave the GitHub user intact.

levb:
The Jira plugin does it already, see https://github.com/mattermost/mattermost-plugin-jira/blob/ec8f206261f3b96045f58378b70469333538aac7/server/webhook.go#L145

agarciamontoro:
I would like to work on this one.

agarciamontoro:
I am planning to solve this as follows:

1. Add a new function to `funcMap` in `template.go` that receives a string and returns that same string with all the GitHub usernames resolved to their assigned Mattermost usernames, if any. If no Mattermost username is found, the GitHub username will remain unchanged. This function will be really similar to the one @levb pointed to, [`replaceJiraAccountIds`](https://github.com/mattermost/mattermost-plugin-jira/blob/ec8f206261f3b96045f58378b70469333538aac7/server/utils.go#L101-L121).
2. Add unit tests for this function in `template_test.go` so we can make sure it resolves all mapped GitHub usernames.
3. Modify all templates that call the `.GetBody` function, piping that body into the new registered function. The templates to modify are:
    - `newPR`
    - `newIssue`
    - `issueComment`
    - `pullRequestReviewEvent`
    - `newReviewComment`
    - `commentMentionNotification`
    - `commentAuthorPullRequestNotification`
    - `pullRequestReviewNotification`
4. Adapt the tests in `template_test.go` so that we test the changes in the body both when there is a `gitHubToUsernameMappingCallback` registered and when there is not.

Some notes:

[...]

PR #177, opened 2020-01-20
GH-176 Replace GitHub usernames in body of notifications

agarciamontoro:
#### Summary

This PR adds a function to the template generation process that matches all Github usernames and replaces them with their assigned Mattermost usernames, if any.

The regular expression to match the GitHub usernames is the following:

```
(^|[^_\x60[:alnum:]])(@[[:alnum:]](-?[[:alnum:]]+)*)
\___________________/ \__________/\_______________/
         (1)             (2.1)           (2.2)
                     \_____________________________/
                                    (2)
```

- The first matching group, (1), matches the beginning of the line or a character that is allowed to precede a GitHub username; i.e., any character different than underscore, backtick or an alphanumeric character.
- The second matching group, (2), wraps the matched GitHub username. Notice that there is no limit on its maximum length (39 at the moment), but in the rare case that we would match a string with more than 39 characters, we would not have any mapping to a Mattermost user, so the string will remain unchanged.
    - The (2.1) part makes sure that the username begins with a letter and is at least 1 character long.
    - The (2.2) part allows dashes, if and only if they are immediately followed by a non-empty string of alphanumeric characters. This can be repeated as many times as needed.

The matched strings contain the preceding character, if any. As there is no `ReplaceAllStringFunc` that accepts a function receiving the slice of submatches (see https://github.com/golang/go/issues/5690),  I needed to split the matched string, in the `replaceAllGitHubUsernames` function, into the prefix and the username itself, which we then lookup using the `lookupMattermostUsername` function.

The `replaceAllGitHubUsernames` function is applied to the following templates, when writing the body of the notification:
  - `newPR`
  - `newIssue`
  - `issueComment`
  - `pullRequestReviewEvent`
  - `newReviewComment`
  - `commentMentionNotification`
  - `commentAuthorPullRequestNotification`
  - `pullRequestReviewNotification`

I have also changed the unit tests in `template_test.go`:
- I have adapted the tests for the corresponding templates, adding an additional `"with mentions"` test that contains the usernames in the `usernameMap` keys, that maps GitHub usernames to Mattermost ones. The string with the values of this map is then used as the expected value in the tests. I did it this way so if we need to add an edge case for a GitHub username in the future, we can just modify the map and let the tests themselves unchanged.
- I have also added a unit test for the GitHub regular expression.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-github/issues/176

### DIFF
--- a/server/template.go
+++ b/server/template.go
@@ -11,10 +11,10 @@ import (
 )
 
 const mdCommentRegexPattern string = `(<!--[\S\s]+?-->)`
-const gitHubRegexPattern string = `(^|[^_\x60[:alnum:]])(@[[:alnum:]](-?[[:alnum:]]+)*)`
+const gitHubUsernameRegexPattern string = `(^|[^_\x60[:alnum:]])(@[[:alnum:]](-?[[:alnum:]]+)*)`
 
 var mdCommentRegex = regexp.MustCompile(mdCommentRegexPattern)
-var gitHubRegex = regexp.MustCompile(gitHubRegexPattern)
+var gitHubUsernameRegex = regexp.MustCompile(gitHubUsernameRegexPattern)
 var masterTemplate *template.Template
 var gitHubToUsernameMappingCallback func(string) string
 
@@ -48,7 +48,7 @@ func init() {
 
 	// Replace any GitHub username with its corresponding Mattermost username, if any
 	funcMap["replaceAllGitHubUsernames"] = func(body string) string {
-		return gitHubRegex.ReplaceAllStringFunc(body, func(matched string) string {
+		return gitHubUsernameRegex.ReplaceAllStringFunc(body, func(matched string) string {
 			// The matched string contains the @ sign, and may contain a single
 			// character prepending the whole thing.
 			gitHubUsernameFirstCharIndex := strings.LastIndex(matched, "@") + 1
--- a/server/template.go
+++ b/server/template.go
@@ -11,6 +11,18 @@ import (
 )
 
 const mdCommentRegexPattern string = `(<!--[\S\s]+?-->)`
+
+// There is no public documentation of what constitutes a GitHub username, but
+// according to the error messages returned in https://github.com/join, it must:
+//   1. be between 1 and 39 characters long.
+//   2. contain only alphanumeric characters or non-adjacent hyphens.
+//   3. not begin or end with a hyphen.
+// When matching a valid GitHub username in the body of messages, it must:
+//   4. not be preceded by an underscore, a backtick (that cryptic \x60) or an
+//      alphanumeric character.
+// Ensuring the maximum length is not trivial without lookaheads, so this
+// regexp ensures only the minimum length, besides points 2, 3 and 4.
+// Note that the username, with the @ sign, is in the second capturing group.
 const gitHubUsernameRegexPattern string = `(^|[^_\x60[:alnum:]])(@[[:alnum:]](-?[[:alnum:]]+)*)`
 
 var mdCommentRegex = regexp.MustCompile(mdCommentRegexPattern)
--- a/server/template.go
+++ b/server/template.go
@@ -11,8 +11,10 @@ import (
 )
 
 const mdCommentRegexPattern string = `(<!--[\S\s]+?-->)`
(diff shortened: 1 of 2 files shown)


server/template.go
```diff
@@ -11,8 +11,10 @@ import (
 )
 
 const mdCommentRegexPattern string = `(<!--[\S\s]+?-->)`
+const gitHubRegexPattern string = `(^|[^_\x60[:alnum:]])(@[[:alnum:]](-?[[:alnum:]]+)*)`
```
lieut-data: Thank you for the detailed breakdown in the PR description. Can you also link out to the definition of what constitutes a valid GitHub username, should we need to refer back to that any point?

server/template.go
```diff
@@ -11,8 +11,10 @@ import (
...
 
 var mdCommentRegex = regexp.MustCompile(mdCommentRegexPattern)
+var gitHubRegex = regexp.MustCompile(gitHubRegexPattern)
```
lieut-data: should we call this `gitHubUsernameRegex` for clarity?

server/template.go
lieut-data: I note that this might match `test+@agarciamontoro.com`, which is otherwise a valid email address, but I suspect this would be rare compared to the flexibility of matching `let's ask @lieut-data+@agarciamontoro`, so no changes needed.

server/template_test.go
```diff
@@ -54,11 +69,64 @@ var issue = github.Issue{
...
 	HTMLURL: sToP("https://github.com/panda"),
 }
 
+// A map of known associations between Github users and Mattermost users
+var usernameMap = map[string]string{
+	"panda":          "pandabot",
+	"asaadmahmood":   "asaad.mahmood",
+	"marianunez":     "maria.nunez",
+	"lieut-data":     "jesse.hallam",
+	"sameusername":   "sameusername",
```
lieut-data: :+1:

server/template_test.go
```diff
@@ -772,6 +1027,64 @@ func TestPullRequestReviewNotification(t *testing.T) {
...
+				State:   sToP("approved"),
+				Body:    sToP("Excited to see git-get-head land!\n" + gitHubMentions),
+			},
+		})
+		require.NoError(t, err)
+		require.Equal(t, expected, actual)
+	}))
+}
+
+func TestGitHubUsernameRegex(t *testing.T) {
```
lieut-data: 👍

server/template.go
agarciamontoro: > Can you also link out to the definition of what constitutes a valid GitHub username, should we need to refer back to that any point?

There is no public definition of what constitutes a valid GitHub username. What I have used here are the error messages returned in https://github.com/join (inspired by https://github.com/shinnn/github-username-regex); namely:
  1. The username must be between 1 and 39 characters long.
  2. The username may only contain alphanumeric characters or single hyphens (i.e., not two adjacent hyphens, as in `@wrong--username`).
  3. The username cannot begin or end with a hyphen.

I can add that as a comment before the `gitHubRegexPattern` definition.

> I note that this might match test+@agarciamontoro.com

This matches indeed test+@agarciamontoro, but as you can see in this same comment, GitHub matches it as a username. I think preserving what GitHub does here is the best thing.

server/template.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-github/pull/177/commits/5db457eae041ea83f8f438228b523fcd6a600cec.

server/template.go
agarciamontoro: Done in https://github.com/mattermost/mattermost-plugin-github/pull/177/commits/c43784aa31f7a9352b578750d82c8b4999347dc2.

agarciamontoro:
Just noticed that I was creating the mocked pull request with mentions with the actual Mattermost usernames, not the GitHub ones.
Fixed it in https://github.com/mattermost/mattermost-plugin-github/pull/177/commits/635104fc5d6d85dbc5cdf75d981fba384b73d7aa.

DHaussermann:
@mickmister is there an easy way to get theses templates since I won't test on our production and don't have prod access to initiate these actions so I can't actually get the template

PR #183, opened 2020-02-04
Fix #165: Validate features before subscribing

### LINKED ISSUE #165: GitHub plugin has no validation if subscribing with an invalid feature
If you try and subscribe to a repo with an invalid feature, The ephemeral post tells you the subscription was successful but nothing will be delivered as the feature does not exist.

Can we validate when a feature does not exist to prevent user error

Steps:
1. type  `/github subscribe DHaussermann/hello-world create` (the 's' is missing)
**Observed** `Successfully subscribed to hello-world.`
**Expected**: `create` is not a valid feature

checkaayush:
#### Summary
Adds validation to prevent subscription to a repo with invalid feature(s)

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/165

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -36,6 +36,45 @@ const COMMAND_HELP = `* |/github connect| - Connect your Mattermost account to y
   * |setting| can be "notifications" or "reminders"
   * |value| can be "on" or "off"`
 
+var validFeatures = map[string]bool{
+	"issues":         true,
+	"pulls":          true,
+	"pushes":         true,
+	"creates":        true,
+	"deletes":        true,
+	"issue_comments": true,
+	"pull_reviews":   true,
+}
+
+// validateFeatures returns false when 1 or more given features
+// are invalid along with a list of the invalid features.
+func validateFeatures(features []string) (bool, []string) {
+	valid := true
+	invalidFeatures := []string{}
+	hasLabel := false
+	for _, f := range features {
+		if _, ok := validFeatures[f]; ok {
+			continue
+		}
+		if strings.HasPrefix(f, "label") {
+			hasLabel = true
+			continue
+		}
+		invalidFeatures = append(invalidFeatures, f)
+		valid = false
+	}
+	if valid && hasLabel {
+		// must have "pulls" or "issues" in features when using a label
+		for _, f := range features {
+			if f == "pulls" || f == "issues" {
+				return valid, invalidFeatures
+			}
+		}
+		valid = false
+	}
+	return valid, invalidFeatures
+}
+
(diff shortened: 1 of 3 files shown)


levb:
@checkaayush it's failing gofmt; you can either run `gofmt -w` ob that file yourself; or as we usually do, configure your editor to automatically apply the go formatting.

hanzei:
Glad to see you are still around @checkaayush :tada:

server/command.go
```diff
@@ -141,6 +161,10 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 				return &model.CommandResponse{}, nil
 			} else if len(optionList) == 1 {
 				features = optionList[0]
+				if !areFeaturesValid(features) {
```
jfrerich: 1/5 I would prefer this function call to return an array of bad features and print them out so the user.  However, given there are currently only a max of 7 specified features, maybe that is overkill.  I will defer to @marianunez to see if she thinks it's worth the change.

Otherwise, this methodology works for me!

aayush-plivo: Makes sense. I'm up for adding it. Waiting for @marianunez 's go-ahead.

checkaayush: Makes sense. I'm up for adding it if needed.

DHaussermann:
Hi @checkaayush thanks for this improvement! It really adds some usability to the feature. 

I think I may have found an issue with this change. I testes subscribing to all features and the validation. I'm not seeing the use of labels work as expected on your branch.    
For example 
`/github subscribe mattermost/mattermost-mobile issues,label:"ruby"` returns `Invalid feature(s) provided: label:"ruby"` 
I believe this exact syntax works on the master branch. Can you see if you can reproduce this issue?

checkaayush:
> Hi @checkaayush thanks for this improvement! It really adds some usability to the feature.
> 
> I think I may have found an issue with this change. I testes subscribing to all features and the validation. I'm not seeing the use of labels work as expected on your branch.
> For example
> `/github subscribe mattermost/mattermost-mobile issues,label:"ruby"` returns `Invalid feature(s) provided: label:"ruby"`
> I believe this exact syntax works on the master branch. Can you see if you can reproduce this issue?

Nice catch @DHaussermann! Have made the required changes and added relevant tests.

I've assumed validating labels, when present, from the GH API is out of scope of the current fix.

checkaayush:
Hey @marianunez, I've made some critical code changes after you approved the code review to fix bugs pointed out during QA. Could you have a quick look at the changes again?

PR #187, opened 2020-02-06
Fix template so `>` is not displayed if review body is empty

### LINKED ISSUE #185: Don't try to print review body if it's empty
https://github.com/mattermost/mattermost-plugin-github/pull/153 added support for including the review comment in the notification.

But if there wasn't any message within the review only `>` is shown. 

Screenshot from 2020-02-05 16-16-39

This issues is about not showing `>` if there is no message.

bolariin:
#### Summary
This PR makes it such that '>' is not displayed if the review body is empty

#### Ticket Link
Resolves #185

### DIFF
--- a/server/template.go
+++ b/server/template.go
@@ -261,8 +261,8 @@ func init() {
 {{- else if eq .GetReview.GetState "changes_requested" }} requested changes on your pull request
 {{- else if eq .GetReview.GetState "commented" }} commented on your pull request
 {{- end }} {{template "reviewRepoPullRequestWithTitle" .}}
->{{.Review.GetBody | replaceAllGitHubUsernames}}
-`))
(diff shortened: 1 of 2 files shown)


server/template.go
```diff
@@ -261,8 +261,8 @@ func init() {
 {{- else if eq .GetReview.GetState "changes_requested" }} requested changes on your pull request
 {{- else if eq .GetReview.GetState "commented" }} commented on your pull request
 {{- end }} {{template "reviewRepoPullRequestWithTitle" .}}
->{{.Review.GetBody | replaceAllGitHubUsernames}}
-`))
+{{if .GetReview.GetBody}}>{{.Review.GetBody | replaceAllGitHubUsernames}}
+{{else}}{{end}}`))
```
bolariin: I added this to deal with the double newlines that appeared when the review body is empty

Issue #179, opened 2020-01-24
Commit notifications trigger unexpected link preview


lieut-data:
I recently received a notification in the [message of a commit](https://github.com/mattermost/mattermost-sa-code/commit/528462af4dca50e41011fe1076c76edb00327b5b):

![image](https://user-images.githubusercontent.com/1023171/73068914-ca5d3c80-3e82-11ea-9a13-39edad97c9ab.png)

In my summary, this then resulted in a perhaps unexpected link preview compared to all the other links in the message:

![image](https://user-images.githubusercontent.com/1023171/73068947-df39d000-3e82-11ea-8ccb-25bf3b8fd795.png)

levb:
(at a superficial glance) - review and add tests for https://github.com/mattermost/mattermost-plugin-github/blob/23f77f082659e78f1d519eb3aacb0ceb94a3cc21/server/plugin.go#L57?

vespian:
I'll like to take this one.

lieut-data:
Thanks, @vespian!

vespian:
Short update - the problem occurs due to other users doing @mentions of the given user in the commit message. We would need to adjust the regex as pointed out in [2] and adjust the current code [1] to handle rewriting the raw link to the commit (i.e. do not always assume a file/blob) to a nice markdown link.

Will try to prepare a patch.

[1] https://github.com/mattermost/mattermost-plugin-github/blob/ab7832c593851ce7b55882a1325e6e64bf933bc1/server/permalinks.go#L113-L117
[2] https://github.com/mattermost/mattermost-plugin-github/issues/179#issuecomment-578401262

Etcaxofjut:
@lieut-data  

I was playing with the code rewriting the raw link to the commit [4] and realized that I may not have all the data yet.

The `ToDo` text you linked in the Jira description one can trigger directly to bot (`/github todo`) but from what I can see/understand this posts an `ephemeral post`, which in turn does not trigger `MessageWillBePosted` hook (see my question in to ~developers channel [2], I already have enabled `Enable Code Previews` in system console and did some printf-debugging to confirm). 

Another option is while connecting/disconnecting to github[5] the todo list also gets displayed, but I guess this is not the case here. 

The third option is the API call via ServeHTTP-`/api/v1/todo`, but I can't find it being called anywhere.

I wonder if I misconfigured somehow my dev instance. How can I recreate it reliably?

[...]

PR #191, opened 2020-02-11
[GH-179] Passing of sender's UserId in BotMessage through Props

vespian:
#### Summary

I was playing with recreating https://github.com/mattermost/mattermost-plugin-github/issues/179 and in a process observed that `MessageWillBePosted` hook has issues fetching GitHub from KV store. 

The problem can be summarized as a mismatch between the UserId used to store GH credentials and the UserID used to retrieve them:
* Github login data is stored with the UserID of the user using the plugin: https://github.com/mattermost/mattermost-plugin-github/blob/6e7a3ef8df706cbaf39c579e42d6452b0cbfe968/server/api.go#L151-L168
* but the ToDo message is being sent with the UserID of the bot:
  * https://github.com/mattermost/mattermost-plugin-github/blob/6e7a3ef8df706cbaf39c579e42d6452b0cbfe968/server/api.go#L743-L745
  * https://github.com/mattermost/mattermost-plugin-github/blob/6e7a3ef8df706cbaf39c579e42d6452b0cbfe968/server/plugin.go#L289
* then, once the hook fires, the code tries to fetch the GitHub login data for the bot (i.e. `github`):
  * https://github.com/mattermost/mattermost-plugin-github/blob/6e7a3ef8df706cbaf39c579e42d6452b0cbfe968/server/plugin.go#L136-L140

The idea is to pass the `real UserId` in the Props. I am not sure how secure it is, and whether we want to harden it somehow. Can a different user spoof such a message? Do we need to secure the message somehow?

It is not clear to me how it works/worked for existing users, maybe there were some existing GH credentials already stored in the KVStore before the upgrade introduced a bug. Maybe nobody really noticed? Dunno.

Kudos to @gabrieljackson for helping me in understanding the issue.

#### Ticket Link

  Relates to https://github.com/mattermost/mattermost-plugin-github/issues/179 but does not solve it.

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -128,12 +128,18 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
 		return nil, ""
 	}
 
-	if post.UserId == "" {
+	userId := post.UserId
+	// Nil-map behaves like an empty map
+	if value, ok := post.Props["UserID"]; ok {
+		userId = value.(string)
+	}
+
+	if userId == "" {
 		return nil, ""
 	}
 
 	msg := post.Message
-	info, err := p.getGitHubUserInfo(post.UserId)
+	info, err := p.getGitHubUserInfo(userId)
 	if err != nil {
 		p.API.LogError("error in getting user info", "error", err.Message)
 		return nil, ""
@@ -290,6 +296,10 @@ func (p *Plugin) CreateBotDMPost(userID, message, postType string) *model.AppErr
 		ChannelId: channel.Id,
 		Message:   message,
 		Type:      postType,
+
+		Props: model.StringInterface{
+			"UserID": userID,
+		},
 	}
 
 	if _, err := p.API.CreatePost(post); err != nil {


vespian:
@lieut-data

>  I think the current behaviour is by design, in that it uses the credentials of the person making the post to query and do the replacements. 

The `MessageWillBePosted` will always be called with the UserId of the bot (see the PR description for details). The github credentials are stored by the plugin with the UserId that connected their account to GH. How does one configure the GH credentials for the bot? Without them the substitutions will not work in the current setup/codebase.

> This means that any user without a linked GitHub account won't have this behaviour, including -- as you note -- the GitHub bot itself.

Makes sense, but I would like to suggest that we document this behaviour and add log message with some warning. For me, as a user trying to configure this plugin properly for the first time and without any knowledge about MM APIs and the plugin itself, this behaviour was not imediatelly clear and caused confusion. I can prepare the patch.

>  If we want to change this approach, I think we need a more generic solution than passing the /target/ UserID that the bot happens to use in a DM, since this could just as easily happen in a public channel to which the plugin is configured to post. I wonder if it's feasible to fall back to the default GitHub credentials instead? Though there may be security concerns there.

GH bot posts ephemeral messages, is there a use-case when they are visible to users othen than the one which e.g. executed the command or got the reminder?



I was also following up the conversation in https://github.com/mattermost/mattermost-plugin-autolink/issues/94 between you and @lveb. I wonder if we should not postpone the discussion about this PR untill @lveb and you decide on the target artchiteture as the code affected by this PR can still change a lot. 

Personally, I would like to trow my 2cents into the direction of https://github.com/mattermost/mattermost-plugin-autolink/issues/94#issuecomment-587485209 as we may not be able to fully utilize the autolinks plugin without the GH-provided data. I know very little about the big picture though, so I may be totally wrong here.


Thanks in advance for the reply :)

lieut-data:
> The MessageWillBePosted will always be called with the UserId of the bot (see the PR description for details). 

I believe this is only true for the "bot DM post" flow, but the substitution behaviour is generic to any post. Here's me linking to a permalink before and after unlinking my GitHub account. Notice that the substitution does not occur in the latter case either:
image

If we want this functionality without requiring the posting user to link their GitHub account (including, of course, your specific case of the bot account posting itself), we need some kind of "top-level" account to do this. This gets tricky when it comes to security concerns, i.e. that account having access to a closed-source repository, and possibly being used to exfiltrate information on a public Mattermost server. I'm in favour of pursuing something like this, but propose tackling in a new issue.

vespian:
> I believe this is only true for the "bot DM post" flow, but the substitution behavior is generic to any post. 

I see. Do you think this is something worthy of an entry in the `Frequently Asked Questions` section of `README.md` ?

>  I'm in favour of pursuing something like this, but propose tackling in a new issue.

Agreed. Let's move the discussion about this feature to a separate issue and maybe gather some thoughts from a Product Manager as well :)

I am closing this PR.

Issue #186, opened 2020-02-05
Result for /github subscribe list should show flags


hanzei:
Currently only features are shown in the list. There should also be all flags listed e.g. `--exclude-org-member`.

aayush-plivo:
I'd like to give this up.

checkaayush:
I'd like to take this up.

hanzei:
Awesome, thanks @checkaayush

checkaayush:
Currently, `--exclude-org-member` is the only supported flag. Assuming there'll be more flags supported in the future, here's what I'd like to propose:

- When 1 or more flags are set:
  * `mattermost-server` - pulls,issues,creates,deletes - `--exclude-org-member,--another-flag`

- When no flag is set:
  * `mattermost-server` - pulls,issues,creates,deletes

hanzei:
That is a great suggestion. Looking forward to the PR :+1:

checkaayush:
> That is a great suggestion. Looking forward to the PR 👍

Here you go: https://github.com/mattermost/mattermost-plugin-github/pull/197!

PR #197, opened 2020-02-23
Result for /github subscribe list now shows flags

checkaayush:
#### Summary
Result for /github subscribe list now shows flags

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/186

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -163,7 +163,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 				subFlags := fmt.Sprintf("%s", sub.Flags.String())
 				txt += fmt.Sprintf("* `%s` - %s", strings.Trim(sub.Repository, "/"), sub.Features)
 				if subFlags != "" {
-					txt += fmt.Sprintf(" - %s", subFlags)
+					txt += fmt.Sprintf(" %s", subFlags)
 				}
 				txt += "\n"
 			}
--- a/server/command.go
+++ b/server/command.go
@@ -28,7 +28,7 @@ const COMMAND_HELP = `* |/github connect| - Connect your Mattermost account to y
 	* label:"<labelname>" - must include "pulls" or "issues" in feature list when using a label
 	Defaults to "pulls,issues,creates,deletes"
   * |flags| currently supported:
-    * --exclude-org-member - events triggered by organization members will not be delivered (the Github organization config
+    * --exclude-org-member - events triggered by organization members will not be delivered (the GitHub organization config
 		should be set, otherwise this flag has not effect)
 * |/github unsubscribe owner/repo| - Unsubscribe the current channel from a repository
 * |/github me| - Display the connected GitHub account
@@ -78,8 +78,8 @@ func validateFeatures(features []string) (bool, []string) {
 func getCommand() *model.Command {
 	return &model.Command{
 		Trigger:          "github",
-		DisplayName:      "Github",
-		Description:      "Integration with Github.",
+		DisplayName:      "GitHub",
+		Description:      "Integration with GitHub.",
 		AutoComplete:     true,
 		AutoCompleteDesc: "Available commands: connect, disconnect, todo, me, settings, subscribe, unsubscribe, help",
 		AutoCompleteHint: "[command]",
--- a/server/command.go
+++ b/server/command.go
@@ -160,7 +160,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 				txt = "### Subscriptions in this channel\n"
 			}
 			for _, sub := range subs {
-				subFlags := fmt.Sprintf("%s", sub.Flags.String())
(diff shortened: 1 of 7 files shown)


server/subscriptions.go
```diff
@@ -28,6 +29,27 @@ func (s *SubscriptionFlags) AddFlag(flag string) {
 	}
 }
 
+func (s SubscriptionFlags) String() string {
+	flags := []string{}
+	v := reflect.ValueOf(s)
+	typeOfS := reflect.TypeOf(s)
+
+	for i := 0; i < v.NumField(); i++ {
+		if !v.Field(i).Bool() {
```
hanzei: Do we _need_ to use `reflect` here? Could this be done without it?

server/subscriptions.go
```diff
@@ -28,6 +29,27 @@ func (s *SubscriptionFlags) AddFlag(flag string) {
...
+			// Flag is unset
+			continue
+		}
+
+		fName := typeOfS.Field(i).Name
+		switch fName {
+		case "ExcludeOrgMembers":
+			flags = append(flags, "--exclude-org-member")
```
hanzei: Nit: Use `EXCLUDE_ORG_MEMBER_FLAG`

server/command.go
```diff
@@ -160,7 +160,12 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 				txt = "### Subscriptions in this channel\n"
 			}
 			for _, sub := range subs {
-				txt += fmt.Sprintf("* `%s` - %s\n", strings.Trim(sub.Repository, "/"), sub.Features)
+				subFlags := fmt.Sprintf("%s", sub.Flags)
```
hanzei: This is clever, but hard to read. Calling `String()` directly would make it more clear to me what the code does.

server/command.go
```diff
@@ -160,7 +160,12 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+				if subFlags == "" {
+					txt += fmt.Sprintf("* `%s` - %s\n", strings.Trim(sub.Repository, "/"), sub.Features)
+				} else {
+					txt += fmt.Sprintf("* `%s` - %s - %s\n", strings.Trim(sub.Repository, "/"), sub.Features, subFlags)
```
hanzei: Nit: You could build the first part `"* `%s` - %s` in every case and only append the second part `- %s` if flags are present. This would help keeping the code DRY.

server/subscriptions.go
checkaayush: I thought about it, but this approach felt the cleanest, assuming we have to retain scope for more flags without changing the core `SubscriptionFlags` struct. Let me know if the assumption can be done away with or if you have a better approach.

server/command.go
```diff
@@ -160,7 +160,12 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 				txt = "### Subscriptions in this channel\n"
 			}
 			for _, sub := range subs {
-				txt += fmt.Sprintf("* `%s` - %s\n", strings.Trim(sub.Repository, "/"), sub.Features)
+				subFlags := fmt.Sprintf("%s", sub.Flags.String())
+				txt += fmt.Sprintf("* `%s` - %s", strings.Trim(sub.Repository, "/"), sub.Features)
+				if subFlags != "" {
+					txt += fmt.Sprintf(" - %s", subFlags)
```
jfrerich: I propose removing the single dash prefixing `subFlags`.  See below.  @hanzei, do you agree?

image

checkaayush: Done.

server/command.go
```diff
@@ -160,7 +160,12 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 				txt = "### Subscriptions in this channel\n"
 			}
 			for _, sub := range subs {
-				txt += fmt.Sprintf("* `%s` - %s\n", strings.Trim(sub.Repository, "/"), sub.Features)
+				subFlags := fmt.Sprintf("%s", sub.Flags.String())
```
hanzei:
```suggestion
				subFlags := sub.Flags.String()
```

server/subscriptions.go
hanzei: Using the [Go Proverbs](https://go-proverbs.github.io/) here I don't think using `reflect` is clear.

I had something like this in mind:

```golang
func (s SubscriptionFlags) String() string {
	flags := []string{}

	if s.ExcludeOrgMembers {
		flag := "--" + EXCLUDE_ORG_MEMBER_FLAG
		flags = append(flags, flag)
	}

	return strings.Join(flags, ",")
}
```

server/command.go
checkaayush: Nice catch! Thanks.

hanzei:
Great work  @checkaayush :tada:

Issue #199, opened 2020-03-02
Subscriptions with excluded org members something still shows org members


hanzei:
This e.g. happend with a new pull request form `@streamer45` (who is a member of the `mattermost` organisation). He created a [pull request](https://github.com/mattermost/mattermost-load-test-ng/pull/127), which triggered a notification.

This ticket is about investigating, why this happend and fixing the code if needed.

hanzei:
This might be caused by organisation memberships being private.

PR #200, opened 2020-03-04
Require read:org scope for OAuth connect

hanzei:
#### Summary
https://github.com/mattermost/mattermost-plugin-github/pull/154 did add a feature flag `--exclude-org-member` to subscriptions, which allows to ignore events from organization members. Currently, the oauth scope doesn't include information about the membership of the user. This PR fixes it by broadening the scope.

To apply this fix to a given subscription the user who created it has to disconnect and reconnect the MM account to GitHub.

I've tested my changes with ngrok and a test organisation.

#### QA test steps
1. Create two users and connect them both to the github via `github connect`
2. Create an organization ORG and let both users join it
3. Make sure organization visibility is set to "private" for user B
4. As user A create a subscription to ORG using `--exclude-org-member`
5. As user B create an issue to trigger the subscription
6. Confirm no notification is shown in MM

@DHaussermann If you need help testing this, just let me know.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/199

### DIFF
--- a/server/plugin.go
+++ b/server/plugin.go
@@ -159,16 +159,16 @@ func (p *Plugin) getOAuthConfig() *oauth2.Config {
 	authURL.Path = path.Join(authURL.Path, "login", "oauth", "authorize")
 	tokenURL.Path = path.Join(tokenURL.Path, "login", "oauth", "access_token")
 
-	repo := "public_repo"
+	repo := github.ScopePublicRepo
 	if config.EnablePrivateRepo {
-		// means that asks scope for privaterepositories
-		repo = "repo"
+		// means that asks scope for private repositories
+		repo = github.ScopeRepo
 	}
 
 	return &oauth2.Config{
 		ClientID:     config.GitHubOAuthClientID,
 		ClientSecret: config.GitHubOAuthClientSecret,
-		Scopes:       []string{repo, string(github.ScopeNotifications), string(github.ScopeReadOrg)},
+		Scopes:       []string{string(repo), string(github.ScopeNotifications), string(github.ScopeReadOrg)},
 		Endpoint: oauth2.Endpoint{
 			AuthURL:  authURL.String(),
 			TokenURL: tokenURL.String(),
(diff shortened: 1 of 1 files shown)


DHaussermann:
@hanzei I'm not quite understanding this problem.  When you say 

> Create two users and connect them both to the github oauth app

Is the issue that users who log into GutHub and make changes while not connected in Mattermost are not affected by the `--exclude-org-member` functionality?

hanzei:
@DHaussermann Does
> Create two users and connect them both to the github via `github connect`

make it more clear?

DHaussermann:
@hanzei No, to clairify my comment above, the problem is not that I don't understand the instructions. I don't understand which users are affected by the issue and which one are not. 

I can do this and test. But, my concern is that I don't see how this differs form the testing done before that did not catch this issue. (I always connect the test users in this way.) Maybe a test user was authenticated in GitHub and not mapped in Mattermost at the time when I tested initially?

DHaussermann:
@hanzei you can disregards this question. ☝️   I'm not having any issues reproducing it. In fact, I'm having a hard time making the flag work now to exclude anyone. 
I'm looking into it.

DHaussermann:
@hanzei I can't make this exclude flag work at all. Maybe we can follow up?

DHaussermann:
Hi @hanzei as we discuss, I'm not seeing the `--exclude-org-member` I have tried to isolate this with new data. In did the folllwing;
- Setup new MM server
- Build GitHub plugin form this branch and install
- Create a new application in my Organization
- Configure the plugin to run with the new App
- Login as the owner, connect to GitHub and Grant access to the app    
- Create a subscription that includes the `--exclude-org-member` flag
As several users who are publicly members of the ORG, perform actions that trigger events to be delivered 
**Observed** Events are still delivered  

I feel like you where on the correct track by seeing what set of permissions are set when you `Grant` but I'm not seeing how to verify this.

hanzei:
@DHaussermann Could you double check that you have set a organisation in settings? Otherwise `--exclude-org-member` has to effect.
Screenshot from 2020-03-19 13-49-46

DHaussermann:
> Could you double check that you have set a organisation in settings? Otherwise --exclude-org-member has to effect.

Thanks @hanzei this was the problem! (on both servers I tried) I did add an issue https://github.com/mattermost/mattermost-plugin-github/issues/214 to validate against this.

Issue #180, opened 2020-01-25
Results from /github subscribe list should be ordered


hanzei:
Currently, the results returned from `/github subscribe list` get returned in a random order. The should be ordered by the repo name instead.

bakurits:
I would like to work on this

hanzei:
Thanks @bakurits :+1:

hanzei:
@bakurits How are you doing with this ticket? Do you have any questions?

bakurits:
> @bakurits How are you doing with this ticket? Do you have any questions?

Yes, I am. I'm done I think, but can't test it. Do you have any suggestions?

hanzei:
You could move your logic for creating a the list of subscriptions into from https://github.com/mattermost/mattermost-plugin-github/blob/9a40ca24e3d9ea9edd2c20079cfe1a9a10208f2a/server/command.go#L157-L164 to `server/subscriptions.go` and then set up a unit test at `server/subscriptions_test.go`. Could this work for you?

bakurits:
I'm on it

bakurits:
@hanzei 
I think I have to mock GetSubscriptions() method here. [https://github.com/mattermost/mattermost-plugin-github/blob/9a40ca24e3d9ea9edd2c20079cfe1a9a10208f2a/server/subscriptions.go#L148](url)
Is there any other way?

PR #201, opened 2020-03-05
Subscribe list is sorted by the repo name

bakurits:
#### Summary
 I've changed the implementation of GetSubscriptionsByChannel, added sorting by repository name, also added some basic unit tests for checking behavior and resolved small typos 

#### Ticket Link
  Fixes https://github.com/mattermost/mattermost-plugin-github/issues/180

### DIFF
--- a/server/subscriptions_test.go
+++ b/server/subscriptions_test.go
@@ -7,6 +7,14 @@ import (
 	"testing"
 )
 
+func CheckError(t *testing.T, wantErr bool, err error) {
+	message := "should return no error"
+	if wantErr {
+		message = "should return error"
+	}
+	assert.Equal(t, wantErr, err != nil, message)
+}
+
 // pluginWithMockedSubs returns mocked plugin for given subscriptions
 func pluginWithMockedSubs(subscriptions []*Subscription) *Plugin {
 	p := NewPlugin()
@@ -100,7 +108,7 @@ func TestPlugin_GetSubscriptionsByChannel(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			got, err := tt.plugin.GetSubscriptionsByChannel(tt.args.channelID)
 
-			assert.Equal(t, tt.wantErr, err != nil, "they should be same")
+			CheckError(t, tt.wantErr, err)
 
 			assert.Equal(t, tt.want, got, "they should be same")
 		})
--- a/server/subscriptions_test.go
+++ b/server/subscriptions_test.go
@@ -3,12 +3,12 @@ package main
 import (
 	"encoding/json"
 	"github.com/mattermost/mattermost-server/v5/plugin/plugintest"
-	"sort"
+	"github.com/stretchr/testify/assert"
 	"testing"
 )
 
-// pluginFromRepoNames returns mocked plugin for given repository names
-func pluginFromRepoNames(subscriptions []*Subscription) *Plugin {
+// pluginWithMockedSubs returns mocked plugin for given subscriptions
+func pluginWithMockedSubs(subscriptions []*Subscription) *Plugin {
 	p := NewPlugin()
 	mockPluginAPI := &plugintest.API{}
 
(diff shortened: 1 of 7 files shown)


server/subscriptions_test.go
```diff
@@ -0,0 +1,95 @@
...
+	for i := 0; i < len(sub1); i++ {
+		if sub1[i].Repository != sub2[i].Repository {
+			return false
+		}
+	}
+	return true
+}
+
+func TestPlugin_GetSubscriptionsByChannel(t *testing.T) {
+
```
hanzei: Nit: drop empty line
```suggestion
func TestPlugin_GetSubscriptionsByChannel(t *testing.T) {
```

server/subscriptions_test.go
```diff
@@ -0,0 +1,95 @@
...
+import (
+	"encoding/json"
+	"github.com/mattermost/mattermost-server/v5/plugin/plugintest"
+	"sort"
+	"testing"
+)
+
+// pluginFromRepoNames returns mocked plugin for given repository names
+func pluginFromRepoNames(repoNames []string, chanelID string) *Plugin {
```
hanzei: Maybe we simplify this to
```suggestion
func pluginFromRepoNames(s subscriptions) *Plugin {
```

and let the caller build the `subscriptions`. This would add more flexibility to the caller.

server/subscriptions_test.go
```diff
@@ -0,0 +1,95 @@
...
+		subs = append(subs, &Subscription{
+			ChannelID:  chanelID,
+			Repository: st,
+		})
+	}
+	return subs
+}
+
+// compareSubsByRepoName checks if two subscription list is identical by repository names
+func compareSubsByRepoName(sub1, sub2 []*Subscription) bool {
```
hanzei: I've used `github.com/stretchr/testify` in the past and find the very handy to compare structs. Would this also work here?

server/subscriptions_test.go
```diff
@@ -0,0 +1,95 @@
...
+		name    string
+		plugin  *Plugin
+		args    args
+		want    []*Subscription
+		wantErr bool
+	}{
+		{
+			name:    "basic test",
+			args:    args{channelID: "1"},
+			plugin:  pluginFromRepoNames([]string{"asd", "123", ""}, "1"),
```
hanzei: Could you add a test where the subscriptions are not already ordered?

server/subscriptions_test.go
```diff
@@ -0,0 +1,95 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := tt.plugin.GetSubscriptionsByChannel(tt.args.channelID)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("GetSubscriptionsByChannel() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !compareSubsByRepoName(got, tt.want) {
+				t.Errorf("GetSubscriptionsByChannel() got = %v, want %v", got, tt.want)
+			}
```
hanzei: `github.com/stretchr/testify` would IMO help making this more readable.

hanzei:
To continue our conversation from https://github.com/mattermost/mattermost-plugin-github/issues/180#issuecomment-593951706:

My suggestion was to let `GetSubscriptionsByChannel()` return an un ordered list and then add a `SortSubscription([]*Subscription) []*Subscription` method and unit test this method. But your integrations tests are also fine :+1:

hanzei:
@bakurits Heads up that you need to merge `master` into your branch

server/subscriptions_test.go
```diff
@@ -1,41 +1,108 @@
...
+			}),
+			want:    wantedSubscriptions([]string{"a", "ab", "b", "c"}, "1"),
+			wantErr: false,
 		},
 	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := tt.plugin.GetSubscriptionsByChannel(tt.args.channelID)
+
+			assert.Equal(t, tt.wantErr, err != nil, "they should be same")
```
larkox: nit: Shouldn't the message be something more in the lines of "should return no error"?

bakurits: Yes, I think that would be better

larkox: Sorry, just realized it will not. Returning or not an error depends on the wantErr variable, so you would have to check that variable to create the error message. I think it would be more readable for the test user if you make the exact message. What about something like:
```
message := "should return no error"
if tt.wantErr {
  message := "should return error"
}
asser.Equal(..., message)
```

hanzei:
Nice work @bakurits. Thanks for the contribution  :tada:

PR #210, opened 2020-03-17
[MM-22640] Make access to private repos opt-in

cpoile:
#### Summary
- With this change we can turn on "Enable Private Repositories" in community.
- Only people who connect with `/github connect private` will be giving the github bot read access to their private repos. People who connect with `/github connect` will be giving access to public_repos only.

#### NOTE
- I was not able to test this with the MM org repos on my dev server. But it works with my personal private repos. Needs testing to make sure it works with MM org repos.

#### Ticket Link
- [MM-22640: Investigate allowing GitHub Plugin to show private requested reviews](https://mattermost.atlassian.net/browse/MM-22640)


server/api.go
```diff
@@ -22,6 +22,12 @@ const (
 	API_ERROR_ID_NOT_CONNECTED = "not_connected"
 )
 
+type OauthState struct {
```
jwilander:
```suggestion
type OAuthState struct {
```
If you accept, will need to update the other uses of it as well

cpoile: Changed, thanks!

server/api.go
```diff
@@ -95,13 +101,34 @@ func (p *Plugin) connectUserToGitHub(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	conf := p.getOAuthConfig()
+	privateAllowed := false
+	pVal := r.URL.Query().Get("private")
+	if pVal == "ok" {
```
lieut-data: Would it make sense to rely on [`strconv.ParseBool`](https://golang.org/pkg/strconv/#ParseBool) for something like this instead of a string value of `ok` for a boolean query string parameter?

server/api.go
```diff
@@ -95,13 +101,34 @@ func (p *Plugin) connectUserToGitHub(w http.ResponseWriter, r *http.Request) {
...
-	url := conf.AuthCodeURL(state, oauth2.AccessTypeOffline)
+	stateBytes, err := json.Marshal(state)
+	if err != nil {
+		http.Error(w, "json marshal failed", http.StatusInternalServerError)
+		return
+	}
+
+	appErr := p.API.KVSet(state.Token, stateBytes)
+	if appErr != nil {
+		fmt.Println(appErr.Error())
```
lieut-data: Intentional to log error to STDOUT? (here, and existing cases below)

server/api.go
```diff
@@ -95,13 +101,34 @@ func (p *Plugin) connectUserToGitHub(w http.ResponseWriter, r *http.Request) {
...
+	}
 
-	url := conf.AuthCodeURL(state, oauth2.AccessTypeOffline)
+	stateBytes, err := json.Marshal(state)
+	if err != nil {
+		http.Error(w, "json marshal failed", http.StatusInternalServerError)
+		return
+	}
+
+	appErr := p.API.KVSet(state.Token, stateBytes)
```
lieut-data: I know this already existed, but this surprises me: I would have expected the flow to be stateless, with the full state being returned back to the plugin signed so as to be trusted.

lieut-data: As another side, should we `KVSetWithExpiry` to ensure these key values are cleaned up even if the oAuth flow never terminates?

server/api.go
```diff
@@ -113,37 +140,46 @@ func (p *Plugin) completeConnectUserToGitHub(w http.ResponseWriter, r *http.Requ
...
+	if appErr != nil {
+		fmt.Println(appErr.Error())
+		http.Error(w, "error deleting stored state", http.StatusBadRequest)
 		return
 	}
 
-	userID := strings.Split(state, "_")[1]
+	var state OAuthState
+	if err := json.Unmarshal(storedState, &state); err != nil {
+		http.Error(w, "json unmarshal failed", http.StatusInternalServerError)
```
lieut-data: Should this be a `StatusBadRequest`?

server/api.go
cpoile: KVSetWithExpiry is a great idea (I should have thought of that) -- added.

This is the way we did it in Jira, and the way the Github plugin was doing it -- but regardless of "how we've always done it" (a bad argument by itself, imho), I think storing the one-time token is the right way to do it. From the oauth2 docs: https://godoc.org/golang.org/x/oauth2#Config.AuthCodeURL 

```
State is a token to protect the user from CSRF attacks. You must always provide a non-empty string and validate that it matches the state query parameter on your redirect callback. See http://tools.ietf.org/html/rfc6749#section-10.12 for more info.
```

If that's not what we're doing here, _please_ let me know (because I've been doing it wrong).

server/api.go
cpoile: I removed this one, since it's in response to the user's click. The ones below are in response to the OAuth flow, and so any returned http error will disappear and never be surfaced for a sysadmin. That's why I'm assuming they're intentional. :)

server/api.go
lieut-data: I think your approach is correct -- I was vaguely recalling a SAML implementation from a lifetime ago, but realized that the key value store here replaces the cookies that otherwise might be used to do the same thing. Thanks for the link and the chance to learn something :)

server/api.go
```diff
@@ -95,13 +104,33 @@ func (p *Plugin) connectUserToGitHub(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	conf := p.getOAuthConfig()
+	privateAllowed := false
+	pValBool, err := strconv.ParseBool(r.URL.Query().Get("private"))
```
hanzei: Any objections on logging `err` if it's `!= nil`?

cpoile: No objection, per se, but why? That would mean the user intentionally edited the connect link to put in something non-true or non-false. That's not really an error, and it should default to false in that case. 
Sound good?

lieut-data: I'd go one step furhter and just write this as:
```
	pValBool, _ := strconv.ParseBool(r.URL.Query().Get("private"))
```

hanzei: While I have to agree in this specific case that ignoring the error is fine, but in general it's a non idiomatic pattern. In the future there might be other sources that hit this endpoint besides the connect link and then it would be handy to know if an error happens.

hanzei: To have a bit more background: I've run into a couple of cases in the past with this plugin and others where I tried debugging a failure, but the error was not logged. Because of this I'm general advocating for logging as many errors as possible.

DHaussermann:
Testing of this ongoing. I have tested with repos on my account but I still need to test with an Organization. 

1. @cpoile - I have found what I think is an issue. Receiving DMs is only dependent on if private repos are enabled in Mattermost. Even when a user uses `connect` instead of `connect private` mentions from with the private repository are received. 
Steps:
- Install plugin from this branch
- Enable Private repos
- Locate or create a private repo (and webhook)
- Invite target user and grant them access
- As target user dis-connect from GitHub (isolation step)        
- As target user connect to github using `connect` and **not** `connect private`
- As a different user in GitHub, open an issue in the private repo and at mention the targhet user
**Observed**: user receives a DM
Can you repro This? 

2. I have an impediment. Because I have a free account, I cannot create a private repo with in my Organization. I need some options for this. Possibly we could temporarily add a webhook on one of the MM private repos and point it to my cloud server? @crspeller any thoughts?

cpoile:
@DHaussermann For 1, I can't repro this.  Here's my steps:
1. Open private repo as UserA (connects webhook), invite UserB.
1. UserB connects without private.
1. UserA mentions UserB, UserB doesn't receive message.
1. UserB disconnects, connects with private.
1. UserA mentions UserB, UserB receives message.
1. UserB disconnects, connects without private.
1. UserA mentions UserB, UserB doesn't receive message.

What do you think the difference is? I'll try it again if you spot a difference.

For 2, I also have that problem. I was thinking we might have to actually test on community? Or maybe on ci-extensions? @crspeller any idea how we can test this?

crspeller:
@DHaussermann I can help. I will DM you.

DHaussermann:
@cpoile I've have taken I closer look at this. 
0/5 My best guess is that the authentication to GitHub is the session state is interfering when you try to change your connection between Public and Private connect. I can also reproduce the inverse when connecting using public first then disconnecting and using private. I don't get the private repo DMs 

So to clarify the issue it seems as simple as: User's cannot switch between `connect` and `connect private` unless they open a new browser session and re-authenticate to GitHub. Using the authentication link generated by the plugin will just use the authentication made on the first mapping. Is this plausible based on the design?

cpoile:
@DHaussermann What browser are you using? Because I can switch from private->public->private and DMs are received as expected. I'm using chrome, and not changing sessions or reloading the chrisp.ngrok.io page. Maybe try connecting to my MM and try it out there?

cpoile:
@DHaussermann try revoking the oauth authorizations you've given to the Mattermost GH Plugin (there should be two for each server you've done this public/private connection with)

cpoile:
@DHaussermann Oh - to answer your question -- there is no code in the plugin that deals with sessions or cookies or anything. It's storing a struct in the db (keyed to the MM UserId), which holds the token the plugin uses for any future github requests. That token will be the one that has the public-only, or private scoped permissions. That token will change when you disconnect and reconnect with a different scope (or will be replaced with the same token if you disconnect and reconnect with the same scope). 

When we receive a webhook, to decide whether or not to show that to a user, the plugin makes a request (using the user's stored token for authentication) and if the github plugin can see the repo, then that means the user can see the webhook event.

So I don't know why you're seeing what you're seeing... Once you revoke the permissions and authorize again, make sure the permissions say "public" when using `/github connect` and `public and private` when using `/github connect private`.

The private one will look like:
image

If you ask for private first, then it won't ask you again when you connect using a lesser scope (public). But the lesser scope will be in effect from then on (until you reconnect with a different scope). I just tested it and it works on chrisp.ngrok.io on latest commit to this PR.

hanzei:
Thanks for adding this PR @cpoile Looking forward to using it on Community :tada:

PR #213, opened 2020-03-18
Refactor command handlers

bakurits:
#### Summary

- Split ExecuteCommand into separate command handlers.
- Made map of [command][handler]
- Resolved some unhandled error warnings

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-github/issues/211

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -309,7 +309,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 		return &model.CommandResponse{}, nil
 	}
 
-	if f, ok := p.CommandHandlers[command]; ok {
+	if f, ok := p.CommandHandlers[action]; ok {
 		message := f(c, args, parameters, info)
 		p.postCommandResponse(args, message)
 		return &model.CommandResponse{}, nil
--- a/server/command.go
+++ b/server/command.go
@@ -97,9 +97,7 @@ func (p *Plugin) postCommandResponse(args *model.CommandArgs, text string) {
 }
 
 func (p *Plugin) getGithubClient(userInfo *GitHubUserInfo) *github.Client {
-	var githubClient *github.Client
-	githubClient = p.githubConnect(*userInfo.Token)
-	return githubClient
+	return p.githubConnect(*userInfo.Token)
 }
 
 func (p *Plugin) handleSubscribe(_ *plugin.Context, args *model.CommandArgs, parameters []string, userInfo *GitHubUserInfo) string {
@@ -150,7 +148,7 @@ func (p *Plugin) handleSubscribe(_ *plugin.Context, args *model.CommandArgs, par
 			if !ok {
 				msg := fmt.Sprintf("Invalid feature(s) provided: %s", strings.Join(ifs, ","))
 				if len(ifs) == 0 {
-					msg = fmt.Sprintf("Feature list must have \"pulls\" or \"issues\" when using a label.")
+					msg = "Feature list must have \"pulls\" or \"issues\" when using a label."
 				}
 				return msg
 			}
--- a/server/command.go
+++ b/server/command.go
@@ -50,7 +50,7 @@ var validFeatures = map[string]bool{
 // are invalid along with a list of the invalid features.
 func validateFeatures(features []string) (bool, []string) {
 	valid := true
-	var invalidFeatures []string
+	invalidFeatures := []string{}
 	hasLabel := false
 	for _, f := range features {
 		if _, ok := validFeatures[f]; ok {
@@ -159,7 +159,7 @@ func (p *Plugin) handleSubscribe(_ *plugin.Context, args *model.CommandArgs, par
 	ctx := context.Background()
 	githubClient := p.getGithubClient(userInfo)
 
-	_, owner, repo := parseOwnerAndRepo(parameters[0], config.EnterpriseBaseURL)
+	owner, repo := parseOwnerAndRepo(parameters[0], config.EnterpriseBaseURL)
 	if repo == "" {
 		if err := p.SubscribeOrg(ctx, githubClient, args.UserId, owner, args.ChannelId, features, flags); err != nil {
 			return err.Error(), nil
--- a/server/command.go
+++ b/server/command.go
@@ -250,9 +250,15 @@ func handleSettings(p *Plugin, _ *plugin.Context, args *model.CommandArgs, param
 
 	if setting == SETTING_NOTIFICATIONS {
 		if value {
-			p.storeGitHubToUserIDMapping(userInfo.GitHubUsername, userInfo.UserID)
+			err := p.storeGitHubToUserIDMapping(userInfo.GitHubUsername, userInfo.UserID)
+			if err != nil {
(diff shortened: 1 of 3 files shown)


bakurits:
I think additional tests are needed. I don't know much about commands behavior, so this should be separate issued

server/command.go
```diff
@@ -95,10 +95,205 @@ func (p *Plugin) postCommandResponse(args *model.CommandArgs, text string) {
...
+	text := "###### Mattermost GitHub Plugin - Slash Command Help\n" + strings.Replace(COMMAND_HELP, "|", "`", -1)
+	p.postCommandResponse(args, text)
+	return &model.CommandResponse{}, nil
+}
+func handleEmpty(p *Plugin, _ *plugin.Context, args *model.CommandArgs, _ []string, _ *github.Client, _ context.Context, _ *GitHubUserInfo) (*model.CommandResponse, *model.AppError) {
+	text := "###### Mattermost GitHub Plugin - Slash Command Help\n" + strings.Replace(COMMAND_HELP, "|", "`", -1)
+	p.postCommandResponse(args, text)
+	return &model.CommandResponse{}, nil
+}
+func handleSettings(p *Plugin, _ *plugin.Context, args *model.CommandArgs, parameters []string, _ *github.Client, _ context.Context, userInfo *GitHubUserInfo) (*model.CommandResponse, *model.AppError) {
```
hanzei: Why is this a function and not a method of `*Plugin`?

server/command.go
```diff
@@ -95,10 +95,205 @@ func (p *Plugin) postCommandResponse(args *model.CommandArgs, text string) {
...
+	err := p.storeGitHubUserInfo(userInfo)
+	if err != nil {
+		mlog.Error(err.Error())
+	}
+
+	p.postCommandResponse(args, "Settings updated.")
+	return &model.CommandResponse{}, nil
+}
+
+type CommandHandleFunc func(p *Plugin, c *plugin.Context, args *model.CommandArgs, parameters []string, githubClient *github.Client, ctx context.Context, userInfo *GitHubUserInfo) (*model.CommandResponse, *model.AppError)
```
hanzei: Why is the same `ctx` for the commands needed? Can they use there own?

hanzei: Is it worth to not pass `*github.Client` every time and instead let he handlers create it themself?

hanzei: Every handler method returns `&model.CommandResponse{}, nil`. Can we unity the code?

server/command.go
```diff
@@ -95,10 +95,205 @@ func (p *Plugin) postCommandResponse(args *model.CommandArgs, text string) {
...
+	} else if setting == SETTING_REMINDERS {
+		userInfo.Settings.DailyReminder = value
+	}
+
+	err := p.storeGitHubUserInfo(userInfo)
+	if err != nil {
+		mlog.Error(err.Error())
+	}
+
+	p.postCommandResponse(args, "Settings updated.")
```
hanzei: Most methods call `p.postCommandResponse` to send a method to the user. Could this message string make sense as a return value for the handlers?

server/command.go
bakurits: In that case, `commandHandlers` map should be inside `Plugin` and initialize with it. Something like that
```go
p.commandHandlers = map[string]CommandHandleFunc{
		"subscribe":   p.handleSubscribe,
		"unsubscribe": p.handleUnSubscribe,
		"disconnect": p.handleDisconnect,
		"todo":        p.handleTodo,
		"me":          p.handleMe,
		"help":       p.handleHelp,
		"":            p.handleEmpty,
		"settings":    p.handleSettings,
	}
```

server/command.go
bakurits: This code should be a separate function then.
```go
        var githubClient *github.Client

	info, apiErr := p.getGitHubUserInfo(args.UserId)
	if apiErr != nil {
		text := "Unknown error."
		if apiErr.ID == API_ERROR_ID_NOT_CONNECTED {
			text = "You must connect your account to GitHub first. Either click on the GitHub logo in the bottom left of the screen or enter `/github connect`."
		}
		p.postCommandResponse(args, text)
	}

	githubClient = p.githubConnect(*info.Token)
```

server/command.go
bakurits: so what you are saying is we should return message from handlers and no more
```go
p.postCommandResponse(args, message)```
lines inside handlers, we will do that after returning message

**Comment by hanzei** [comment]
@bakurits FYI: The last commit caused a build failure:
```
./command.go:162:17: assignment mismatch: 3 variables but parseOwnerAndRepo returns 2 values
make: *** [Makefile:71: server] Error 2
```

**Comment by bakurits** [comment]
> @bakurits FYI: The last commit caused a build failure:
> 
> ```
> ./command.go:162:17: assignment mismatch: 3 variables but parseOwnerAndRepo returns 2 values
> make: *** [Makefile:71: server] Error 2
> ```

wow I think my utils.go file is out of date parseOwnerAndRepo is returning 3 variables in my version

File: server/command.go [hunk 4]
```diff
@@ -135,174 +305,12 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
-		} else if setting == SETTING_REMINDERS {
-			info.Settings.DailyReminder = value
-		}
-
-		p.storeGitHubUserInfo(info)
-
-		p.postCommandResponse(args, "Settings updated.")
-		return &model.CommandResponse{}, nil
+	if f, ok := p.CommandHandlers[command]; ok {
+		message, err := f(c, args, parameters, info)
```
hanzei: Is there any place where `err != nil`? If not, can we drop the error return value.

bakurits: `err` is always nil. I can remove it

mickmister:
@DHaussermann This will need a regression test for each slash command/subcommand

DHaussermann:
@bakurits When I build and deploy this I'm seeing that the commands that were moved no longer work. No changes where made to how the commands must be used correct?

For example `/github help`  is no longer recognized. 

GitHub-Command

Can you see if you can repro this? As far as I can tell I have the latest commit so, I'm not sure what's happening.

bakurits:
@DHaussermann I think I've found the bug. Checks don't work, don't know why.

I think we should add unit tests for command

server/command.go
```diff
@@ -96,10 +96,181 @@ func (p *Plugin) postCommandResponse(args *model.CommandArgs, text string) {
...
+		return fmt.Sprintf("Successfully subscribed to organization %s.", owner)
+	}
+
+	if err := p.Subscribe(ctx, githubClient, args.UserId, owner, repo, args.ChannelId, features, flags); err != nil {
+		return err.Error()
+	}
+
+	return fmt.Sprintf("Successfully subscribed to %s.", repo)
+}
+func (p *Plugin) handleUnSubscribe(_ *plugin.Context, args *model.CommandArgs, parameters []string, _ *GitHubUserInfo) string {
```
larkox, line 176: nit: handleUnSubscribe -> handleUnsubscribe

hanzei:
Great work @bakurits :tada:

PR #221, opened 2020-03-20
fix linting error

jfrerich:
This fixes the linting error from `make check-style`


levb:
Why do we have QA Review on feature-branch PRs. I don't think we should? cc @DHaussermann

PR #230, opened 2020-03-26
go mod tidy

hanzei:
#### Summary
Run `go mod tidỳ`. In the near feature `go mod tidy` will be enforced by CI.


jfrerich:
Are we omitting QA on purpose?

hanzei:
@jfrerich Yes, I think we can do this here. This doesn't effect the build at all.

PR #123, opened 2019-09-20
Adding create issue modal

dhadiseputro:
Most of the work here is from #120 and I just modified some parts to fit the requirements to create issue. I am new to react redux, so I am sorry if there are some stupid mistakes. Any kind of feedback is very appreciated, as it is important for my learning :)
Thanks!
ps. I am still stuck trying to bring up the webapp, so this code hasn't been run yet. I'll update it asap when I get the webapp portion to run (if there are any mistakes).

PR #239, opened 2020-04-14
Merge FEATURE-create-issue-modal into master

jfrerich:
#### Summary
This PR is a request to merge the `FEATURE-create-issue-modal` into the `master` branch. 

NOTE TO PR REVIEWERS:

All of the code has been approved in previous PRs that were incrementally merged into the feature branch.  Please review with this in mind (ie. you've likely already given in-depth reviews) 

QA: To expedite the turnaround time to Feature Complete, we omitted QA from the incremental PRs, once I took over the code from the original author.  @DHaussermann, you should see fixes for all items you found in the original PR for this feature.  

Here is the comment from that original PR.  https://github.com/mattermost/mattermost-plugin-github/pull/123#issuecomment-601809037


server/api.go
```diff
@@ -931,3 +942,124 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+	info, err := p.getGitHubUserInfo(userID)
+	if err != nil {
+		writeAPIError(w, err)
+		return
+	}
+
+	ctx := context.Background()
+	githubClient := p.githubConnect(*info.Token)
+
+	repositories, _, apiErr := githubClient.Repositories.List(ctx, "", &github.RepositoryListOptions{})
```
hanzei, line 962: `RepositoryListOptions` used `Page` and `PerPage` for paging. Do you catch the case where there are more then repos then fit into a response? See also https://pkg.go.dev/github.com/google/go-github/v31/github?tab=doc#hdr-Pagination

server/api.go
```diff
@@ -931,3 +942,124 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+	if apiErr != nil {
+		mlog.Error(apiErr.Error())
+		return
+	}
+
+	resp, _ := json.Marshal(repositories)
```
hanzei, line 968: How big does the response to the client get if a user has access to a lot (>400) repos?

webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -0,0 +1,205 @@
...
+                            onClick={this.handleClose}
+                        />
+                        <FormButton
+                            type='submit'
+                            btnClass='btn btn-primary'
+                            saving={submitting}
+                            defaultMessage='Submit'
+                            savingMessage='Submitting'
+                        >
+                            {'Submit'}
```
hanzei, line 185: The icon doesn't render for me in chromium 80
Screenshot from 2020-04-16 14-16-10

server/api.go
```diff
@@ -931,3 +942,124 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+
+	githubClient := p.githubConnect(*info.Token)
+
+	// call Create(ctx context.Context, owner string, repo string, issue *IssueRequest)
+	result, resp, api_err := githubClient.Issues.Create(ctx, info.GitHubUsername, issue.Repo, gh_issue)
+	if api_err != nil {
+		mlog.Error(api_err.Error())
+	}
+
+	if resp.Response.StatusCode == http.StatusGone {
```
hanzei, line 1028: Do we have to catch the case where `resp.Response.StatusCode != http.StatusOK`?

server/api.go
```diff
@@ -931,3 +942,124 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+	info, err := p.getGitHubUserInfo(userID)
+	if err != nil {
+		writeAPIError(w, err)
+		return
+	}
+
+	githubClient := p.githubConnect(*info.Token)
+
+	// call Create(ctx context.Context, owner string, repo string, issue *IssueRequest)
+	result, resp, api_err := githubClient.Issues.Create(ctx, info.GitHubUsername, issue.Repo, gh_issue)
```
hanzei, line 1023: Why does the issue get created under the org `info.GitHubUsername`? What happens if the issue should be created in an org repo?

server/api.go
```diff
@@ -931,3 +942,124 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+	if api_err != nil {
+		mlog.Error(api_err.Error())
```
hanzei, line 1025: Why does the code continue if there is an error?

webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -0,0 +1,205 @@
...
+                    maxLength={65}
+                    value={this.state.issueTitle}
+                    onChange={this.handleIssueTitleChange}
+                />
+                {issueTitleValidationError}
+                <Input
+                    label='Description for the GitHub Issue'
+                    type='textarea'
+                    value={this.props.post.message}
+                    disabled={true}
```
hanzei, line 141: Why can't the user edit the description?

server/api.go
```diff
@@ -931,3 +942,124 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	if issue.Repo == "" {
+		writeAPIError(w, &APIErrorResponse{ID: "", Message: "Please provide a valid repo name.", StatusCode: http.StatusBadRequest})
+		return
+	}
+	ctx := context.Background()
+	gh_issue := &github.IssueRequest{
+		Title: &issue.Title,
+		Body:  &issue.Body,
```
hanzei, line 1010: Maybe we do something similar to the JIRA plugin and make clear that an issue was created from Mattermost:
https://github.com/mattermost/mattermost-plugin-jira/blob/3dea7716f4e3ba04d8d894a0329fe590fb25ae29/server/issue.go#L151

server/api.go
```diff
@@ -931,3 +942,124 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+		writeAPIError(w, &APIErrorResponse{ID: "", Message: "Please provide a valid issue description.", StatusCode: http.StatusBadRequest})
+		return
+	}
+
+	if issue.Repo == "" {
+		writeAPIError(w, &APIErrorResponse{ID: "", Message: "Please provide a valid repo name.", StatusCode: http.StatusBadRequest})
+		return
+	}
+	ctx := context.Background()
+	gh_issue := &github.IssueRequest{
```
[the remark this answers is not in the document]
jfrerich: Only because we missed it in reviews :)

server/api.go
```diff
@@ -931,3 +942,124 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+	userID := r.Header.Get("Mattermost-User-ID")
+	if userID == "" {
+		http.Error(w, "Not authorized", http.StatusUnauthorized)
+		return
+	}
+
+	// get data for the issue from the request body and fill IssueRequest object
+	issue := &IssueRequest{}
+	dec := json.NewDecoder(r.Body)
+	if err := dec.Decode(&issue); err != nil {
```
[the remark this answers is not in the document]
jfrerich: Done.

server/api.go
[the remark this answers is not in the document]
jfrerich: Done

server/api.go
```diff
@@ -931,3 +942,124 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+	if api_err != nil {
+		mlog.Error(api_err.Error())
+	}
+
+	if resp.Response.StatusCode == http.StatusGone {
+		writeAPIError(w, &APIErrorResponse{ID: "", Message: "Issues are disabled on this repository.", StatusCode: http.StatusMethodNotAllowed})
+		return
+	}
+
+	api := p.API
```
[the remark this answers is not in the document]
jfrerich: done

server/api.go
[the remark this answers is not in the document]
jfrerich: Passing the empty string will list the repos for the authenticated user. 

image

webapp/src/components/modals/create_issue/create_issue.jsx
jfrerich: Done

server/api.go
```diff
@@ -931,3 +942,124 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+
+	resp, _ := json.Marshal(repositories)
+	w.Write(resp)
+}
+
+func (p *Plugin) createIssue(w http.ResponseWriter, r *http.Request) {
+	if r.Method != http.MethodPost {
+		http.Error(w, fmt.Sprintf("Request: %s is not allowed, must be POST", r.Method), http.StatusMethodNotAllowed)
+		return
+	}
```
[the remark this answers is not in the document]
jfrerich: @hanzei, would you mind creating an GH issue for this?

server/api.go
jfrerich: I see.  I committed a change to the form_button css classes and believe this is now fixed.  @asaadmahmood, does this seem correct with this new commit?  https://github.com/mattermost/mattermost-plugin-github/pull/239/commits/4e7e89e54e53d926c66f7aecfbc280b3a29cc089

webapp/src/components/modals/create_issue/create_issue.jsx
jfrerich: I see.  I committed a change to the form_button css classes and believe this is now fixed.  @asaadmahmood, does this seem correct with this new commit?  https://github.com/mattermost/mattermost-plugin-github/pull/239/commits/4e7e89e54e53d926c66f7aecfbc280b3a29cc089

server/api.go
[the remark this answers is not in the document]
hanzei: Got it, thanks for the explanation :+1:

server/api.go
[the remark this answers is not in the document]
hanzei: https://github.com/mattermost/mattermost-plugin-github/issues/246

server/api.go
jfrerich: Done.

server/api.go
jfrerich: First off, good catch.  I've been thinking about this here's some of the issues.  I'd like to discuss this before moving on to implementation.

1) The repo selector currently list repos that I have available but doesn't include the Mattermost Org.  I will need to address this first.  There is a `GithubOrg` setting in the plugin settings.  I am curious if this is being set to `mattermost` in the community plugin?  I believe the implementation should by default get the repos you are a member of and if `GitHubOrg` is set, the repos should show only mattermost org repos.  

2) In the code above, you are correct and the `owner` should come from the `owner/repo` that will be specified in the create-issue modal. This could be a user repo, or an org.  My suggestion is to fix the item in 1) first.  Doing so will get us the correct `owner`/`repo` values to add to teh `githubClient.Issues.Create()` command

jfrerich: 3) if we use the org mode and only show `mattermost` org repos, that will reduce restrict us to create issues in personal repos.  For instance.  We have the `wrangler` channel and could not create an issue in the repo because it is under Gabe, not Mattermost

hanzei: 1, Yes, `GithubOrg` is set to `mattermost` on the community server. Agreed with your conclusion.
2. :+1: 
3. While is see that there might be some rough edge cases for us, not showing "non org" repos still seams consistent to me. E.g. "non org" notifications are omitted.

server/api.go
```diff
@@ -931,3 +935,139 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+
+	ctx := context.Background()
+	ghIssue := &github.IssueRequest{
+		Title: &issue.Title,
+		Body:  &issue.Body,
+	}
+
+	siteUrl := p.API.GetConfig().ServiceSettings.SiteURL
+	permalink := getPermaLink(*siteUrl, issue.PostId, issue.CurrentTeam)
```
hanzei, line 1016: Side note: Providing the team is not needed to create a perma link. `getPermaLink` can just be:
```go
func getPermaLink(siteUrl, postId string) string {
	return fmt.Sprintf("%v/_redirect/pl/%v", siteUrl, postId)
}
```

jfrerich: Thanks, @hanzei.  I changed the function and occurrences of the function call.

server/api.go
jfrerich: Please see https://github.com/mattermost/mattermost-plugin-github/pull/239/commits/d2a20760d413d4b4020afeff3edcce728a2dedf6 for the fix.  Basically, the repo value sent back from the client is now <owner_or_org>/<reponame>.  This is split by the server-side code to correctly determine the issue repo destination.

server/api.go
jfrerich: Any apiErr from the client now handled with writeAPIError. https://github.com/mattermost/mattermost-plugin-github/pull/239/commits/d2a20760d413d4b4020afeff3edcce728a2dedf6

server/api.go
```diff
@@ -931,3 +935,130 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+	if apiErr != nil {
+		mlog.Error(apiErr.Error())
+	}
+
+	if resp.Response.StatusCode == http.StatusGone {
+		writeAPIError(w, &APIErrorResponse{ID: "", Message: "Issues are disabled on this repository.", StatusCode: http.StatusMethodNotAllowed})
+		return
+	}
+
+	post, appErr := p.API.GetPost(issue.PostId)
```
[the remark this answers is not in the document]
jfrerich: I think I agree with you.  At first, I was thinking that `appErr` would catch a non-existent post, which it does, but not until after the issue is created.  I've added the check on `issue.PostId` with the other checks.  https://github.com/mattermost/mattermost-plugin-github/pull/239/commits/ce678a8405d99c844be42450f3b4924cf26f534c

webapp/src/components/react_select_setting.jsx
```diff
@@ -0,0 +1,157 @@
...
+        if (this.props.resetInvalidOnChange) {
+            this.setState({invalid: false});
+        }
+    };
+
+    // Standard search term matching plus reducing to < 100 items
+    filterOptions = (input) => {
+        let options = this.props.options;
+        if (input) {
+            options = options.filter((x) => x.label.toUpperCase().includes(input.toUpperCase()));
```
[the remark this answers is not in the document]
jfrerich: This was arbitrary and added by the original author.  I check the occurrences in mattermost-webapp and you are correct (toLowerCase() seems more popular).

image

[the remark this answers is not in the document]
jfrerich: https://github.com/mattermost/mattermost-plugin-github/pull/239/commits/f8f4fdd1c8d1eaf058024c2bdf201478436857cf

webapp/src/components/react_select_setting.jsx
```diff
@@ -0,0 +1,157 @@
...
+        if (this.props.required && this.state.invalid) {
+            validationError = (
+                <p className='help-text error-text'>
+                    <span>{requiredMsg}</span>
+                </p>
+            );
+        }
+
+        let selectComponent = null;
+        if (this.props.limitOptions && this.props.options.size > MAX_NUM_OPTIONS) {
```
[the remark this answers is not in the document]
jfrerich: I would say yes, this is a better option, and just set the default in this component if MAX_NUM_OPTIONS is not set by the parent.  This code was stolen from `jira` and I tried minimizing changes to the code.  Hopefully we can compare the two and create a generic shared component because this one is so useful.

webapp/src/components/react_select_setting.jsx
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    componentDidUpdate(prevProps, prevState) {
+        if (prevState.invalid && (this.props.value && this.props.value.value) !== (prevProps.value && prevProps.value.value)) {
+            this.setState({invalid: false}); //eslint-disable-line react/no-did-update-set-state
+        }
+    }
+
+    handleChange = (value) => {
```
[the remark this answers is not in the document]
jfrerich: how about `optionValue`??  I like `value` as that seems a common input to a component.  I'm open to changing and 0/5 on naming preference.

server/api.go
```diff
@@ -931,3 +942,124 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request) {
...
+	githubClient := p.githubConnect(*info.Token)
+
+	// call Create(ctx context.Context, owner string, repo string, issue *IssueRequest)
+	result, resp, api_err := githubClient.Issues.Create(ctx, info.GitHubUsername, issue.Repo, gh_issue)
+	if api_err != nil {
+		mlog.Error(api_err.Error())
+	}
+
+	if resp.Response.StatusCode == http.StatusGone {
+		writeAPIError(w, &APIErrorResponse{ID: "", Message: "Issues are disabled on this repository.", StatusCode: http.StatusMethodNotAllowed})
```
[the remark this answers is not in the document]
jfrerich: No, it doesn't.  What it does do is cause the modal to hang.  I'll have to get a fix in for this!~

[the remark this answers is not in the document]
jfrerich: I was able to bubble up the `writeAPIError()` messages to the UI.  

A couple of notes.  Validation is handled in the UI so checks on issue.Repo and issue.Title won't ever not be set.  For the others errors, the user will see them now.  

image

Addressed in this commit https://github.com/mattermost/mattermost-plugin-github/pull/239/commits/0cabd9524af13464b78852f98dc3a6cb63820c1d

server/api.go
```diff
@@ -805,7 +805,7 @@ func (p *Plugin) createIssueComment(w http.ResponseWriter, r *http.Request) {
 
 	siteUrl := api.GetConfig().ServiceSettings.SiteURL
 
-	permalink := getPermaLink(*siteUrl, req.PostId, req.CurrentTeam)
```
larkox, line 804: Since we do not use it here, do we need passing around the current team at all?

hanzei: Good catch. Done in 8dad1cd5d89a93d96aca6a44e7d7f02291a00fba

webapp/src/components/modals/create_issue/create_issue.jsx
hanzei: Works now :+1:

server/api.go
hanzei: fixed in f4a1cc3

hanzei:
@jfrerich I did fix the fetching of repositories and cleaned up some code pieces.

hanzei:
@mickmister This is ready for a final review.

@levb @larkox I did some changes since you last looked at the code. Feel free to re-review.

jfrerich:
@hanzei, I reviewed all of your changes and they look great.  Thanks for contributing and helping on this while I'm on SET!

server/api.go
```diff
@@ -853,3 +847,163 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request, userID s
...
+		return
+	}
+
+	rootId := issue.PostId
+	if post.RootId != "" {
+		rootId = post.RootId
+	}
+
+	reply := &model.Post{
+		Message:   fmt.Sprintf("Created GitHub issue [#%v](%v)", *result.Number, *result.HTMLURL),
```
hanzei, line 994: @jfrerich Maybe we can make it more clear _what_ message what used to create the issue. Also is it needed to mention that it's a GitHub issue?
```suggestion
		Message:   fmt.Sprintf("Created [#%v](%v) from a [message](%s)", *result.Number, *result.HTMLURL, permalink),
```

server/api.go
```diff
@@ -853,3 +847,163 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request, userID s
...
+	}
+
+	ghIssue := &github.IssueRequest{
+		Title: &issue.Title,
+		Body:  &issue.Body,
+	}
+
+	permalink := p.getPermaLink(issue.PostId)
+
+	mmMessage := fmt.Sprintf("_Issue created from a [message in Mattermost](%v)_.", permalink)
```
hanzei, line 936: @jfrerich Mentioning the author of the message would help make the context of the message more clear. But I'm failing a bit on the wording.  

My initial suggestion was `Issue created from a message in Mattermost from @hanzei`, but it feels weird to have `from` two times in the sentence.

server/api.go
jfrerich: @hanzei, I think it might be useful to recognize this is a GitHub issue.  Although a user can decipher it from the link and will know that they created the issue in Github, I consider it subtle advertising for users scanning the posts in a channel

server/api.go
jfrerich: Hmmm.. I thought about several other options, but believe yours is quite simple and correct.  How about the following?

`Issue created from a [Mattermost message] by @hanzei`

server/api.go
jfrerich: Also, as more plugins are added, that create issues, it might be useful to acknowledge in the message where the issue went..  (Examples: github, jira, gitlab, bitbucket, .etc)

hanzei: Got, makes sense to keep mentioning that it's about a GitHub issue. What do you think about the message link?

jfrerich: I think it is ok to exclude the message link also.  The post is a reply to the original post, creating a thread.  I'm 0/5 though
image

server/api.go
mickmister: @jfrerich Do you have any comments on this?

server/api.go
mickmister: 1. You can test the org functionality by making your own org on GitHub and scoping the plugin setting to that org @jfrerich 

If the org is set, I think we should restrict to only creating issues in the org's repos. That's currently how notifications operate in the plugin.

webapp/src/components/react_select_setting.jsx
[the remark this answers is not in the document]
mickmister: @larkox It's a sane default to minimize the number of DOM elements present at any given time. Someone had 5000+ Jira projects which led to us making this component work this way :)

The limit only applies to what is currently selectable in the component. If you start typing, options that were not visible at first (e.g. their first letter is fairly late in the alphabet) will show after the results have been filtered.

server/api.go
jfrerich: I'm not sure how to measure this or what the commonly used solution is to this situation. Is it typical to limit the number of items in a request and if so, how do we determine the limit?

mickmister: It's probably fine since we are doing a one-time fetch. We wouldn't want to skip out on results that are supposed to be there, so I think we shouldn't limit.

What do the results look like the plugin is not set up to be scoped to an org? Do I get only repos that I interact with on GitHub?

webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -0,0 +1,221 @@
...
+        if (this.state.showErrors && !this.state.issueTitleValid) {
+            issueTitleValidationError = (
+                <p className='help-text error-text'>
+                    <span>{requiredMsg}</span>
+                </p>
+            );
+        }
+
+        if (!visible) {
+            return null;
```
mickmister, line 120: I think it makes more sense to have this at the top of the function

webapp/src/components/react_select_setting.jsx
```diff
@@ -0,0 +1,157 @@
...
+
+    componentWillUnmount() {
+        if (this.props.removeValidate && this.props.name) {
+            this.props.removeValidate(this.props.name);
+        }
+    }
+
+    componentDidUpdate(prevProps, prevState) {
+        if (prevState.invalid && (this.props.value && this.props.value.value) !== (prevProps.value && prevProps.value.value)) {
+            this.setState({invalid: false}); //eslint-disable-line react/no-did-update-set-state
```
mickmister, line 56: Same comment about [getDerivedStateFromProps](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops) here. Non-blocking suggestion

webapp/src/components/react_select_setting.jsx
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    handleChange = (value) => {
+        if (this.props.onChange) {
+            if (Array.isArray(value)) {
+                this.props.onChange(this.props.name, value.map((x) => x.value));
+            } else {
+                const newValue = value ? value.label : null;
+                this.props.onChange(this.props.name, newValue);
```
mickmister, line 66: Why are we using `value.label` to pass to `this.props.onChange`? That's certainly not what `ReactSelectSetting` should be doing, and is not what it does in the Jira plugin either. 4/5 We should be using `value.value`.

server/api.go
```diff
@@ -853,3 +851,182 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request, userID s
...
+	type IssueRequest struct {
+		Title  string `json:"title"`
+		Body   string `json:"body"`
+		Repo   string `json:"repo"`
+		PostId string `json:"post_id"`
+	}
+
+	// get data for the issue from the request body and fill IssueRequest object
+	issue := &IssueRequest{}
+	if err := json.NewDecoder(r.Body).Decode(&issue); err != nil {
```
mickmister, line 924: Can we add this italics text at the end of the issue description:

> _Created in Mattermost from [this post](somewhere.com)_

There is similar functionality in the Jira plugin's feature

server/api.go
hanzei: I did the link given that it doesn't have to be the root post that was used for the issue.

server/api.go
```diff
@@ -643,8 +648,10 @@ func (p *Plugin) searchIssues(w http.ResponseWriter, r *http.Request, userID str
 	w.Write(resp)
 }
 
-func getPermaLink(siteUrl string, postId string, currentTeam string) string {
-	return fmt.Sprintf("%v/%v/pl/%v", siteUrl, currentTeam, postId)
+func (p *Plugin) getPermaLink(postId string) string {
+	siteUrl := *p.API.GetConfig().ServiceSettings.SiteURL
```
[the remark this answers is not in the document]
hanzei: While there is a theoretical possibility for `SiteURL` to be `nil` the server does ensure that there are no `nil` values. I don't think it's worth the error handling here.

server/api.go
```diff
@@ -853,3 +851,182 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request, userID s
...
+
+	var allRepos []*github.Repository
+	opt := github.ListOptions{PerPage: 50}
+
+	if org == "" {
+		for {
+			repos, resp, err := githubClient.Repositories.List(ctx, "", &github.RepositoryListOptions{ListOptions: opt})
+			if err != nil {
+				mlog.Error(err.Error())
+				return
```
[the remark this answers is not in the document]
hanzei: There is no consistent error handling in the code. Returning there just lead to an empty response, which is not wrong. There is room for improvement, but  I need to think about it and there are more important issues.

server/api.go
```diff
@@ -853,3 +851,182 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request, userID s
...
+		return
+	}
+
+	rootId := issue.PostId
+	if post.RootId != "" {
+		rootId = post.RootId
+	}
+
+	reply := &model.Post{
+		Message:   fmt.Sprintf("Created GitHub issue [#%v](%v) from a [message](%s)", *result.Number, *result.HTMLURL, permalink),
```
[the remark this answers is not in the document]
hanzei: While I see that this is useful, I'm hesitant to introduce a style for rendering links to GitHub. There is no autolink functionality in the plugin right now. Once that gets added, I'm happy to add it also here.

server/api.go
```diff
@@ -853,3 +851,182 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request, userID s
...
+	}
+
+	ghIssue := &github.IssueRequest{
+		Title: &issue.Title,
+		Body:  &issue.Body,
+	}
+
+	permalink := p.getPermaLink(issue.PostId)
+
+	mmMessage := fmt.Sprintf("_Issue created from a [Mattermost message](%v) by *from %s*._", permalink, username)
```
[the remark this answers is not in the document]
hanzei: `mmMessage` is used to post post the GitHub issue. Hence, the GitHub username ist used to ensure, the user get's notified. 
Screenshot from 2020-05-09 01-29-04

webapp/src/actions/index.js
```diff
@@ -88,6 +88,29 @@ export function getReviewsDetails(prList) {
...
+export function getRepos() {
+    return async (dispatch, getState) => {
+        let data;
+        try {
+            data = await Client.getRepositories();
+        } catch (error) {
+            return {error: data};
+        }
+
+        const connected = await checkAndHandleNotConnected(data)(dispatch, getState);
```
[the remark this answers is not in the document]
hanzei: I can't say why, but every other api call has the same pattern. Another ticket to investigate?

webapp/src/components/github_repo_selector/github_repo_selector.jsx
```diff
@@ -0,0 +1,65 @@
...
+                    limitOptions={true}
+                    required={true}
+                    onChange={this.onChange}
+                    options={repoOptions}
+                    isMulti={false}
+                    key={'repo'}
+                    theme={this.props.theme}
+                    addValidate={this.props.addValidate}
+                    removeValidate={this.props.removeValidate}
+                    value={repoOptions.find((option) => option.label === this.props.value)}
```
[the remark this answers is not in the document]
hanzei: @jfrerich :po

webapp/src/components/github_repo_selector/index.js
```diff
@@ -0,0 +1,25 @@
...
+import {connect} from 'react-redux';
+import {bindActionCreators} from 'redux';
+
+import {getRepos} from '../../actions';
+
+import GithubRepoSelector from './github_repo_selector.jsx';
+
+function mapStateToProps(state) {
+    return {
+        yourRepos: state['plugins-github'].yourRepos,
```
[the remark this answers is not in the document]
hanzei: Fixed in fb1609e

server/api.go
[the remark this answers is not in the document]
hanzei: I would consider that out of scope of the PR for now.

webapp/src/components/post_menu_action/create_issue/index.js
```diff
@@ -0,0 +1,28 @@
...
+
+import CreateIssuePostMenuAction from './create_issue';
+
+const mapStateToProps = (state, ownProps) => {
+    const post = getPost(state, ownProps.postId);
+    const systemMessage = post ? isSystemMessage(post) : true;
+
+    return {
+        isSystemMessage: systemMessage,
+        connected: state[`plugins-${pluginId}`].connected,
```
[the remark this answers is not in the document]
hanzei: Good suggestion. Added in 40378f2.

webapp/src/reducers/index.js
```diff
@@ -93,6 +93,15 @@ function yourPrs(state = [], action) {
     }
 }
 
+function yourRepos(state = [], action) {
+    switch (action.type) {
+    case ActionTypes.RECEIVED_REPOSITORIES:
+        return action.data;
```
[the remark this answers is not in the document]
hanzei: On my local environment it takes 1-2 seconds to fetch ~50 repos. Hence, it think there is value in having it cached.

webapp/src/components/modals/create_issue/create_issue.jsx
hanzei: Fixed in eab01f5

webapp/src/components/react_select_setting.jsx
hanzei: @jfrerich :point_up:

server/api.go
hanzei: That is already the case

Screenshot from 2020-05-09 01-29-04

hanzei:
@jfrerich I did deffer some of the webapp comments towards you given that I don't understand the code well.

@mickmister Please let me know which of the webapp comments are blocking.

webapp/src/components/react_select_setting.jsx
mickmister: @hanzei I'll make this into a ticket. It's not causing any functional issues because the label and value are the same in the case of the GitHub repo options, but it is certainly not how we should be using this component.

webapp/src/actions/index.js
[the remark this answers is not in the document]
mickmister: @hanzei I think it's safe to say it is there for a valid reason. I'll make a ticket to check it out in case we can improve the error handling.

server/api.go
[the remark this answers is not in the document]
mickmister: > the server does ensure that there are no nil values

@hanzei is this check for nil values by the server new functionality? We have experienced panics for this same reason in the past and have since guarded against it in most plugins.

server/api.go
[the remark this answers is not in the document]
mickmister: @hanzei I want to know why this is the case. If we don't call `Write` or `WriteHeader` (or an equivalent), how does the response get closed? Is it something that the server does automatically for us? Maybe it's my mis-understanding of how http requests are handled in Go.

mickmister:
@hanzei After looking through my requested changes, there are none that are blocking. I'm making a ticket for the `option.label` vs `option.value` situation, and investigating why we need to do the `checkAndHandleNotConnected` call.

server/api.go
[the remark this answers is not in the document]
hanzei: I was under the impression that this should not be an issue for every server version. Do you have a link to a ticket/conversation where this issue came up?

server/api.go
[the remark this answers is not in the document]
hanzei: That is part of the `http` package. Once `ServeHTTP` terminates and not header has been written, 200 gets send as a response.

Mini example:
```golang
	http.HandleFunc("/", func(w http.ResponseWriter, req *http.Request) {})
	log.Fatal(http.ListenAndServe(":8080", nil))
```

server/api.go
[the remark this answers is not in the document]
mickmister: @hanzei Here's the PR that fixed the issue in the Jira plugin https://github.com/mattermost/mattermost-plugin-jira/pull/316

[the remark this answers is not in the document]
hanzei: Thanks for the hint. I added a check to `OnActivate`.

DHaussermann:
There is a bit of an issue remaining with attach functionality. The business logic is not consistent between Create and Attach when the plugin is not locked to any Organization.

Attach will allow attaching to any public repo where you're able to create issues

Create will allow you to create an issue in any repo where you are a member or contributor. Public repos that allow any users to create an issue are not returned in seaach

@aaronrothschild - I don't feel too strongly that the attach functionality is more correct (it's a bit awkward to use). It's not necessarily an issue that I can do something from the Web UI but not via the plugin. But, having inconsistent business logic between the 2 features is more of an issue. I would suggest that the logic of which repos you can create in match the Attach functionality Thoughts?

mickmister:
> Attach will allow attaching to any public repo where you're able to add an issue comment.

@DHaussermann I always thought this was a bit overkill. My opinion is to change the attach functionality to match create, and only include repos that you are a member/contributor of. If I want to create an issue or comment on a repo that I am not a member of, I wouldn't think this plugin would help me do that. It should be a more tailored experience for the user imo.

DHaussermann:
> My opinion is to change the attach functionality to match create, and only include repos that you are a member/contributor of.

I'm inclined to agree with @mickmister  I don't think being able to comment in any public repo adds too much value and it makes the usability worse. One thing to conciser is that we already have attach out in production. We would need to clearly release note that we're making this change.

mickmister:
It seems to me that this is ready to merge. I've created a [ticket](https://github.com/mattermost/mattermost-plugin-github/issues/273) to change the "Attach to Issue" functionality as it seems obvious that the UX should be tailored to the user. I'd like to get this merged so we can address the remaining issues in separate PRs.

@DHaussermann I think we can merge once we have your blessing on this. Regarding the followup change for "Attach to Issue", I agree with "We would need to clearly release note that we're making this change."

hanzei:
:+1: for continuing the discussion about "Attach to Issue" in https://github.com/mattermost/mattermost-plugin-github/issues/273

Issue #238, opened 2020-04-13
Update webapp code to use `id` from manifest.js


jfrerich:
Currently, the pluginId value is hardcoded in many places, using `state['plugins-github']`.

This ticket request is to replace all occurrences with `plugins-${id}`

bakurits:
I would like to work on that

jfrerich:
Wow! 2 minutes after posting :) Sure, it is yours, @bakurits.

bakurits:
@jfrerich wouldn't it be better to call this variable `pluginId` and not `id`?
for example here
```javascript
const id = ownProps.user ? ownProps.user.id : '';
const user = state['plugins-github'].githubUsers[id] || {};
```
id is already declared

jfrerich:
@bakurits, I agree that pluginId would be more appropriate.  Please use that naming convention!

PR #240, opened 2020-04-14
Replaced hard coded plugin id with manifested variable

bakurits:
#### Summary

Replaced hard coded plugin id with manifested variable

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/238


hanzei:
@DHaussermann A smoke test on the webapp part should be enough to test this PR.

Issue #255, opened 2020-05-02
Add tooltip for pull request links


hanzei:
Similar on how mousing over a link on github.com works there should be a tooltip for plugin to pull requests in the webapp. 

It should show:
- The repo
- The date where the PR was opened
- Title
- Number
- The first three labels

This is how a tooltip looks like on GitHub. Please note that some information that GitHub shows should not be part of the initial implementations to reduce complexity. 

![Screenshot from 2020-05-01 11-01-01](https://user-images.githubusercontent.com/16541325/80863706-5b139780-8c7e-11ea-9284-9530a44fa36d.png)

For more information on how to add tooltips to the webapp see https://github.com/mattermost/mattermost-webapp/pull/2386, which introduced support for it.

hanzei:
@asaadmahmood Let me know if you have any concerns about copying the style from GitHub.

asaadmahmood:
@hanzei Not much, some of the tooltips and alignments look off, but we can fix that when a PR is in.

fedealconada:
I can take this one @hanzei

hanzei:
That would be awesome, @fedealconada :tada:

PR #252, opened 2020-04-25
Enable more linters

hanzei:
#### Summary
This PR sync the `golangci-lint` config with the one from other repos. Hence, a lot of new linter got enabled. The config file is synced with https://github.com/mattermost/mattermost-plugin-starter-template/pull/90.

Given the high number of changed lines, I tried to break the changes into multiple commits.

While most changes are just stylistic, I did comment on the more tricky one.

#### Ticket Link
Follow up on https://github.com/mattermost/mattermost-plugin-autolink/pull/108

PR #258, opened 2020-05-05
[GH-255] Add tooltip for pull request and issue links

fedealconada:
#### Summary
This PR adds a tooltip for pull request and issue links

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/255

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -889,7 +889,11 @@ func (p *Plugin) getPrByNumber(w http.ResponseWriter, r *http.Request, userID st
 	owner := r.FormValue("owner")
 	repo := r.FormValue("repo")
 	number := r.FormValue("number")
-	numberInt, _ := strconv.Atoi(number)
+
+	numberInt, err := strconv.Atoi(number)
+	if err != nil {
+		mlog.Error(err.Error())
+	}
 
 	if info, err := p.getGitHubUserInfo(userID); err != nil {
 		writeAPIError(w, err)
--- a/server/api.go
+++ b/server/api.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/pkg/errors"
 	"net/http"
 	"strconv"
 	"strings"
@@ -604,7 +605,7 @@ func fetchReviews(ctx context.Context, client *github.Client, repoOwner string, 
 	reviewsList, _, err := client.PullRequests.ListReviews(ctx, repoOwner, repoName, number, nil)
 
 	if err != nil {
-		return []*github.PullRequestReview{}, err
+		return []*github.PullRequestReview{}, errors.Wrap(err, "could not list reviews")
 	}
 
 	return reviewsList, nil
@@ -729,23 +730,37 @@ func (p *Plugin) createIssueComment(w http.ResponseWriter, r *http.Request, user
 		return
 	}
 
-	commentUser, appErr := api.GetUser(post.UserId)
-	if appErr != nil {
-		writeAPIError(w, &APIErrorResponse{ID: "", Message: "failed to load post.UserID " + post.UserId + ": not found", StatusCode: http.StatusInternalServerError})
-		return
+	commentUsername := ""
+
+	if info, apiEr := p.getGitHubUserInfo(post.UserId); apiEr != nil {
+		if apiEr.ID == API_ERROR_ID_NOT_CONNECTED {
+			commentUser, appEr := api.GetUser(post.UserId)
+			if appEr != nil {
+				writeAPIError(w, &APIErrorResponse{ID: "", Message: "failed to load post.UserID " + post.UserId + ": not found", StatusCode: http.StatusInternalServerError})
+				return
+			}
+			commentUsername = commentUser.Username
+		} else {
+			writeAPIError(w, apiEr)
+			return
+		}
+	} else {
+		commentUsername = info.GitHubUsername
 	}
 
-	currentUser, appErr := api.GetUser(userID)
-	if appErr != nil {
-		writeAPIError(w, &APIErrorResponse{ID: "", Message: "failed to load current user", StatusCode: http.StatusInternalServerError})
+	currentUsername := ""
+	if info, apiEr := p.getGitHubUserInfo(userID); apiEr != nil {
+		writeAPIError(w, apiEr)
 		return
+	} else {
+		currentUsername = info.GitHubUsername
 	}
 
 	siteUrl := api.GetConfig().ServiceSettings.SiteURL
 
 	permalink := getPermaLink(*siteUrl, req.PostId, req.CurrentTeam)
 
-	permalinkMessage := fmt.Sprintf("*@%s attached a* [message](%s) *from @%s*\n", currentUser.Username, permalink, commentUser.Username)
+	permalinkMessage := fmt.Sprintf("*@%s attached a* [message](%s) *from @%s*\n", currentUsername, permalink, commentUsername)
 
 	req.Comment = permalinkMessage + req.Comment
 	comment := &github.IssueComment{
@@ -754,7 +769,7 @@ func (p *Plugin) createIssueComment(w http.ResponseWriter, r *http.Request, user
 
 	result, rawResponse, err := githubClient.Issues.CreateComment(ctx, req.Owner, req.Repo, req.Number, comment)
 	if err != nil {
-		writeAPIError(w, &APIErrorResponse{ID: "", Message: "failed to create an issue comment: " + getFailReason(rawResponse.StatusCode, req.Repo, currentUser.Username), StatusCode: rawResponse.StatusCode})
+		writeAPIError(w, &APIErrorResponse{ID: "", Message: "failed to create an issue comment: " + getFailReason(rawResponse.StatusCode, req.Repo, currentUsername), StatusCode: rawResponse.StatusCode})
 		return
 	}
 	rootId := req.PostId
@@ -763,8 +778,9 @@ func (p *Plugin) createIssueComment(w http.ResponseWriter, r *http.Request, user
 		rootId = post.RootId
 	}
 
+	permalinkReplyMessage := fmt.Sprintf("Message attached to [#%v](https://github.com/%v/%v/issues/%v)", req.Number, req.Owner, req.Repo, req.Number)
 	reply := &model.Post{
-		Message:   fmt.Sprintf("Message attached to [#%v](https://github.com/%v/%v/issues/%v)", req.Number, req.Owner, req.Repo, req.Number),
+		Message:   permalinkReplyMessage,
 		ChannelId: post.ChannelId,
 		RootId:    rootId,
 		ParentId:  rootId,
--- a/server/api.go
+++ b/server/api.go
@@ -77,6 +77,8 @@ func (p *Plugin) initialiseAPI() {
 	apiRouter.HandleFunc("/unreads", p.extractUserMiddleWare(p.getUnreads, false)).Methods("GET")
 	apiRouter.HandleFunc("/settings", p.extractUserMiddleWare(p.updateSettings, false)).Methods("POST")
 	apiRouter.HandleFunc("/user", p.extractUserMiddleWare(p.getGitHubUser, true)).Methods("POST")
+	apiRouter.HandleFunc("/issue", p.extractUserMiddleWare(p.getIssueByNumber, false)).Methods("GET")
+	apiRouter.HandleFunc("/pr", p.extractUserMiddleWare(p.getPrByNumber, false)).Methods("GET")
 }
 
 func (p *Plugin) extractUserMiddleWare(handler HTTPHandlerFuncWithUser, jsonResponse bool) http.HandlerFunc {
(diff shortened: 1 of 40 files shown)


server/api.go
```diff
@@ -853,3 +855,54 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request, userID s
...
 }
+
+func (p *Plugin) getIssueByNumber(w http.ResponseWriter, r *http.Request, userID string) {
+
+	ctx := context.Background()
+
+	var githubClient *github.Client
+	owner := r.FormValue("owner")
+	repo := r.FormValue("repo")
+	number := r.FormValue("number")
```
hanzei, line 866: Other API methods are using a request body instead.  (e.g. https://github.com/mattermost/mattermost-plugin-github/blob/119ff1807a40eca316e5c1644589f056baa71936/server/api.go#L284-L291)

I would like to converge on that pattern for consistency despite there not being a functional difference.

server/api.go
```diff
@@ -853,3 +855,54 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request, userID s
...
+	numberInt, _ := strconv.Atoi(number)
```
hanzei, line 867: Given that this is a user input, the error should be properly checked and `Debug` logged.

server/api.go
```diff
@@ -853,3 +855,54 @@ func (p *Plugin) updateSettings(w http.ResponseWriter, r *http.Request, userID s
...
+	if info, err := p.getGitHubUserInfo(userID); err != nil {
+		writeAPIError(w, err)
+		return
+	} else {
+		githubClient = p.githubConnect(*info.Token)
+	}
+
+	result, _, err := githubClient.Issues.Get(ctx, owner, repo, numberInt)
+	if err != nil {
+		mlog.Error(err.Error())
```
hanzei, line 878: Why not `return` here?

webapp/src/components/link_tooltip/tooltip.css
```diff
@@ -0,0 +1,116 @@
...
+    font-size: 14px;
+    display: inline;
+}
+
+.github-tooltip .tooltip-info .markdown-text {
+    -ms-overflow-style: none;
+    -webkit-hyphens: auto;
+    -ms-hyphens: auto;
+    hyphens: auto;
+    max-height: 150px;
```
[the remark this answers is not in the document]
fedealconada: Well... actually, I gave this a though and decided to do it like that. The thing is, the scrolling is enabled, so you should be able to scroll down there but I have hidden the scrollbar because I thought it was more "elegant" not to show it, but it might seem a bit weird.
Either I could show the scrollbar or I could play around with the 'react-markdown' so it only renders an amount of HTML tags... There are several options in which this could be done... I could also just convert the whole text to one only <p> with no format and add ellipsis on overflow, i don't know.... 
I would say that, because of the scope of this issue (which was not actually including the pr/issue description on the tooltip), I would rather decide between keeping it like this or un-hiding the scrollbar. Thoughts?

webapp/src/components/link_tooltip/link_tooltip.jsx
```diff
@@ -0,0 +1,169 @@
...
+                        verticalAlign='middle'
+                    />
+                </span>
+            );
+            break;
+        }
+        return icon;
+    };
+
+    const getLabelFontColor = (hexcolor) => {
```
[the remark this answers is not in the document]
fedealconada: Yep, assign it to me

webapp/src/components/link_tooltip/link_tooltip.jsx
```diff
@@ -0,0 +1,169 @@
...
+        let iconType;
+        switch (data.type) {
+        case 'pull':
+            color = '#28a745';
+            iconType = GitPullRequest;
+            if (data.state === 'closed' && data.merged) {
+                color = '#6f42c1';
+                iconType = GitMerge;
+            } else {
+                color = '#cb2431';
```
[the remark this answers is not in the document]
fedealconada: Good catch, I'll take a look at it :)

server/api.go
fedealconada: Well... I would disagree on this... `/createissuecomment` is a POST endpoint which calls the Github POST endpoint, so it makes sense that it receives a body. But in the case of `/issue` and `/pr` are GET endpoints so it makes sense to receive the values as query params.
What do you think?

server/api.go
fedealconada: Makes sense!

server/api.go
fedealconada: That's the way it's done on other parts of the code... For instance, [here](https://github.com/mattermost/mattermost-plugin-github/blob/master/server/api.go#L799)

webapp/src/components/link_tooltip/tooltip.css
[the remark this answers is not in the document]
hanzei: I'm wondering how much a scrollbar would be actually used. The tooltip is mostly "for a quick glance" and if a user wants for information, the link to the issue/pr is right there. 


@asaadmahmood I would like to hear your toughs on this. Is the current solution acceptable to you?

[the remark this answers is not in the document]
fedealconada: I feel it could be useful... Even though it's a quick glance, sometimes we get a bit lazy about having to open the link, wait the page to load, etc etc. Imagine that maybe you know there's a useful link inside the description that you know is placed at the bottom of it... just scrolling will help you out... But yeah, up to others opinions :)

hanzei:
@asaadmahmood A couple of screenshots:
Screenshot from 2020-05-05 21-43-55
Screenshot from 2020-05-05 21-44-02
Screenshot from 2020-05-05 21-44-11
Screenshot from 2020-05-05 21-44-18
Screenshot from 2020-05-05 21-44-24

webapp/src/components/link_tooltip/link_tooltip.jsx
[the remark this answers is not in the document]
fedealconada: Pushed change to fix that @hanzei :)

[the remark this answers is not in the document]
hanzei: Great, thanks @fedealconada :+1:

server/api.go
hanzei: That is a good point. While this is consistent, it might cause unexpected behaviour if `result` is `nil`. https://github.com/mattermost/mattermost-plugin-github/pull/252 will address all issue in other parts of the code.

webapp/src/components/link_tooltip/link_tooltip.jsx
```diff
@@ -0,0 +1,157 @@
+import React, {useEffect, useState} from 'react';
+import PropTypes from 'prop-types';
+import './tooltip.css';
+import Octicon, {GitMerge, GitPullRequest, IssueClosed, IssueOpened} from '@primer/octicons-react';
+import ReactMarkdown from 'react-markdown';
+
+import Client from 'client';
+import {getLabelFontColor} from '../../utils/styles';
+
+export const LinkTooltip = ({href}) => {
```
hanzei, line 10: FYI: CI is failing because of this line:
```
  10:30  error  'href' is missing in props validation  react/prop-types
 ```

File: server/api.go (hunk 1 above)
Comment (hanzei) [reply server/api.go:866]: Good point, I completely missed that `/issue` and `/pr` are `POST` endpoints.

File: server/api.go (hunk 3 above)
Comment (fedealconada) [reply server/api.go:878]: So, what do you suggest? Because, currently, if an error occurs `result` would just be `nil`, and then this [line](https://github.com/mattermost/mattermost-plugin-github/pull/258/files/a9fc584c315325868877a4969ba89b624ada6097#diff-0d5a44e9b96f048fa52b477ce6be6253R881) is executed, which ignores any possible error as well.
If I handle that error, there should be an error handler if the marshal occurs... Thoughts?

File: webapp/src/components/link_tooltip/link_tooltip.jsx (hunk 7 above)
Comment (fedealconada) [reply webapp/src/components/link_tooltip/link_tooltip.jsx:10]: I'll take a loot at it!

Comment (fedealconada) [reply webapp/src/components/link_tooltip/link_tooltip.jsx:10]: Fixed, I had a typo.

File: server/api.go (hunk 3 above)
Comment (hanzei) [reply server/api.go:878]: If the code just `return`s here, it's made sure that no invalid response is send. I'm aware that still `200` is returned, but that could be separate ticket.

File: webapp/src/components/link_tooltip/tooltip.css (hunk 4 above)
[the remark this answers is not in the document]
Comment (asaadmahmood) [reply webapp/src/components/link_tooltip/tooltip.css:71]: @hanzei Sorry for replying late guys. I don't think a scrollbar should be needed here, we can just have a max-height or max number or elements, without it cutting off content.

I would also be willing to add a "See more" link above the labels which would just open the PR (similar to the title link) to be very obvious.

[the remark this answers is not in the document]
Comment (fedealconada) [reply webapp/src/components/link_tooltip/tooltip.css:71]: it currently has a max-height, which cuts the text off and that's why I've added the ability to scroll (just that I've hidden the scrollbar, but you are allowed to scroll) @asaadmahmood I'm not so sure which is the solution you propose.
Regarding the "See more" link, no problem!

File: webapp/src/components/link_tooltip/link_tooltip.jsx [hunk 8]
```diff
@@ -0,0 +1,165 @@
...
+    const [data, setData] = useState(null);
+
+    useEffect(() => {
+        const init = async () => {
```
larkox, line 14: Correct me if I am wrong, but this piece of code will:
1. Execute on component mount
2. Fetch the PR
3. Update the status
4. Render the element with updated status
5. Execute again after the render
6. Fetch the PR one more time
7. Update the status
8. Will not render since the status will be the same as previous.

Maybe a check like `if data != null return` would avoid the second fetch?

fedealconada: Hey @larkox, thanks for the feedback. I'm not sure if I follow you... Are you referring to this [check](https://github.com/mattermost/mattermost-plugin-github/pull/258/files#diff-030a588562eadd14c868ae7644412561R80)?

larkox: That check will only render if there is some data, which makes sure the tooltip does not render before you have retrieved the data from the server, and not render the tooltip if the link is not a github link.

My problem is that the effect you are registering with useEffect, if I am not mistaken, will be executed after every render. I think (correct me if I am wrong) that the body of the effect should only be called once (you only want to fetch the PR once during the lifetime of the tooltip). Having some kind of check to make sure it is only executed once would be nice to have.

fedealconada: I see! Great catch. I've just added the line that fixes that :) @larkox

DHaussermann:
@hanzei i If you have some time, could you please peer-test this PR using the peer testing process documented here: https://mattermost.atlassian.net/wiki/spaces/INT/pages/619937892/Peer+Testing+Process

webapp/src/components/link_tooltip/link_tooltip.jsx
```diff
@@ -0,0 +1,168 @@
...
+                                </div>
+                            )}
+
+                            <div className='see-more mt-1'>
+                                <a
+                                    href={href}
+                                    target='_blank'
+                                    rel='noopener noreferrer'
+                                >See more</a>
+                            </div>
```
hanzei, line 140: 2/5: It's quite common to have this kind of links put at the end of a unit. Hence I'm wondering if the link should be put after the labels.

fedealconada: hey @hanzei, I put the link there because @asaadmahmood told me to do it like that. Check his comment [https://github.com/mattermost/mattermost-plugin-github/pull/258#discussion_r423559012](here). But if you prefer me to put it below the labes, I can do it! Let me know your thoughts!

fedealconada:
@hanzei I've just fixed the warning issue, check the changes.

webapp/src/components/link_tooltip/link_tooltip.jsx
hanzei: Good point, I missed that comment. Then I'm fine with how it's is.

asaadmahmood:
@hanzei @fedealconada Not going to go much into the code, but can we have screenshots of how it looks so I can approve/comment on it from a UX perspective?

fedealconada:
@hanzei my bad, I forgot to run the linter. Changes are pushed now :) @asaadmahmood see below the screenshot, the only thing that has changed has been the addition of the "See more" button. Clicking on it opens up a new tab.

image

Hovering over the link looks like this:

image

asaadmahmood:
@fedealconada Would it not be possible for us to just add an elipsis after x number of characters and then on a new line add See more?
I don't like it cutting the text like that.

asaadmahmood:
@fedealconada Can we also vertically middle align the arrow here.
Screenshot 2020-05-17 at 7 22 27 PM

fedealconada:
Hey @asaadmahmood, I agree, I don't like the text being cut off. I've been digging a bit into that. What I've managed to do is, instead of adding ellipsis, just hide the elements (check image below).

image

Thoughts?

asaadmahmood:
@fedealconada This looks good to me, but are we sure that solution would work with different PR descriptions?

asaadmahmood:
Discussed it with @fedealconada, we're going with the overflow trick for now, and may create an improvement issue later.

fedealconada:
So, after talking with @asaadmahmood we’ve agreed that we will keep the text cut off so we can get this feature released and maybe on a later iteration we can adjust some minor termination style details

Issue #254, opened 2020-05-02
Properly wrap errors


hanzei:
Generally errors should be wrapped using [`Wrap`](https://pkg.go.dev/github.com/pkg/errors?tab=doc#Wrap) and [`Wrapf`](https://pkg.go.dev/github.com/pkg/errors?tab=doc#Wrapf). This ticket is about going trough the code base and adding error wrapping everywhere where it's appropriate.

_Originally posted by @bakurits in https://github.com/mattermost/mattermost-plugin-github/pull/252_

dbejanishvili:
I would like to work on that

hanzei:
Thanks @dbejanishvili. Let me know, if you have any questions.

dbejanishvili:
https://github.com/mattermost/mattermost-plugin-github/blob/119ff1807a40eca316e5c1644589f056baa71936/server/subscriptions.go#L240

It seems that the error returned from KVSet is ignored here, is this intentional? If not, should I fix it?

hanzei:
No, that error should not be ignore. Please fix it.

bakurits:
> No, that error should not be ignore. Please fix it.

@hanzei don't you already resolved this in https://github.com/mattermost/mattermost-plugin-github/pull/252 ?
https://github.com/mattermost/mattermost-plugin-github/blob/0953348ee0956cd2478bfdbc97220f14a6261bc3/server/subscriptions.go#L243

hanzei:
You have eagle eyes @bakurits! That is correct, but I hope #260 gets merged before my PR and I can easily resolve the merge conflict. :wink:

PR #260, opened 2020-05-05
Errors properly wrapped

dbejanishvili:
#### Summary
Every error returned from a function is now wrapped with an appropriate message.
Fixed KVSet return error being ignored

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-github/issues/254

### DIFF
--- a/server/utils.go
+++ b/server/utils.go
@@ -205,7 +205,10 @@ func filterLines(s string, start, end int) (string, error) {
 		buf.WriteByte(byte('\n'))
 	}
 
-	return buf.String(), errors.Wrap(scanner.Err(), "scanner error occurred")
+	if err := scanner.Err(); err != nil {
+		return "", errors.Wrap(err, "scanner error occurred")
+	}
+	return buf.String(), nil
 }
 
 // getLineNumbers return the start and end lines from an anchor tag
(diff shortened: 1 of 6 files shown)


server/utils.go
```diff
@@ -205,7 +205,7 @@ func filterLines(s string, start, end int) (string, error) {
 		buf.WriteByte(byte('\n'))
 	}
 
-	return buf.String(), scanner.Err()
+	return buf.String(), errors.Wrap(scanner.Err(), "scanner error occurred")
```
hanzei, line 208: That error handling is really weird. Given that you are touching the code anyway, would you mind fixing the error handling and doing a standard `if err := scanner.Err(); err != nil { ...`

dbejanishvili: sure

dbejanishvili: Since I am already working with errors, I took it upon myself to handle some more unhandled errors. Also, there are some parts of the code (particularly in plugin.go) where an error can occur but the function doesn't return errors so it can't be handled, I figured I would at least log those errors.  Is it ok if I make these changes?

DHaussermann:
@hanzei If you have some time, could you please peer-test this PR using the  peer testing process documented here: https://mattermost.atlassian.net/wiki/spaces/INT/pages/619937892/Peer+Testing+Process

Issue #278, opened 2020-05-19
JS error in login screen


hanzei:
![Screenshot from 2020-05-19 18-02-41](https://user-images.githubusercontent.com/16541325/82364270-58030000-9a0f-11ea-8b67-dd5cad8734fa.png)

hanzei:
That is a regression caused by https://github.com/mattermost/mattermost-plugin-github/pull/248

PR #279, opened 2020-05-19
/api/v1/connected should be publicly accessible

hanzei:
#### Summary
`/api/v1/connected` should be publicly accessible. Otherwise, the webapp can't  figure out if a user is connected or not in the login screen.

#### QA Steps

1. Deploy PR
2. Open Mattermost in an incognito  window
3. Confirm no js error message is shown

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/278


server/api.go
```diff
@@ -351,6 +351,8 @@ func (p *Plugin) getConnected(w http.ResponseWriter, r *http.Request, userID str
 		Organization:      config.GitHubOrg,
 	}
 
+	userID := r.Header.Get("Mattermost-User-ID")
+
 	info, _ := p.getGitHubUserInfo(userID)
```
larkox, line 356: Nit: Not sure if necessary, but to make more clear this, I would only call getGitHubUserInfo if the userID is different to the empty string. This function will be clearer in its intent, and we will do one less call to the KVStore.

hanzei: Good point on the call to KV Store. I tried implementing it that way, but the code didn't got cleaner. By increasing the indenting, it's harder to read what the code does.

larkox: Not sure if clearer or not, but what about deferring the marshaling and writing of resp? Something like this:
```
resp := ...
defer func () {
    b, _ := json.Marshal(resp)
    w.Write(b)
}

userID := r.Header.Get("Mattermost-User-ID")
if userID == "" {
    return
}

info, _ := p.getGitHubUserInfo(userID)
if info == nil || info.Token == nil {
   return
}
...
```

larkox: Another solution would be to declare such function inside the function and use it before the returns.
```
resp := ...
writeResp := func () {
    b, _ := json.Marshal(resp)
    w.Write(b)
}

userID := r.Header.Get("Mattermost-User-ID")
if userID == "" {
    writeResp()
    return
}

info, _ := p.getGitHubUserInfo(userID)
if info == nil || info.Token == nil {
   writeResp()
   return
}
...
```

hanzei: Let's use the helper methods added in https://github.com/mattermost/mattermost-plugin-github/pull/252 to make the code cleaner. I will alter the code when either PR get's merged.

Issue #224, opened 2020-03-25
Error if user tries to connect private repos and "Enable Private Repositories" is disabled


hanzei:
https://github.com/mattermost/mattermost-plugin-github/pull/210 added a per user base opt in for private repositories using `/github connect private`. Connecting to private repos using `/github connect private` is only possible if "Enable Private Repositories" is enabled.

The current UX is confusing, if "Enable Private Repositories" is disabled:
- `/github help` still shows information about `/github connect private`. These information should be omitted.
- `/github connect private` works, but does not request access to private repos. Instead it should error, if "Enable Private Repositories" is disabled.

PR #288, opened 2020-05-26
Improve help text if private repos are disabled

hanzei:
#### Summary
If private repos are disabled, `/github help` should not inform about `/github connect private`.

I opted to use a template given that it seamed easier to implement to me then some string concatenation guarded by `if p.getConfiguration().EnablePrivateRepo`.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/224


server/template.go
```diff
@@ -263,6 +263,33 @@ func init() {
...
 {{if .GetReview.GetBody}}>{{.Review.GetBody | replaceAllGitHubUsernames}}
 {{else}}{{end}}`))
+
+	template.Must(masterTemplate.New("helpText").Parse("" +
+		"* `/github connect{{if .EnablePrivateRepo}} [private]{{end}}` - Connect your Mattermost account to your GitHub account.\n" +
+		"{{if .EnablePrivateRepo}}" +
+		"  * `private` is optional. If used, the GitHub bot will ask for read access to your private repositories. " +
+		"If these repositories send webhook events to this Mattermost server, you will be notified of changes to those repositories.\n" +
+		"{{end}}" +
+		"* `/github disconnect` - Disconnect your Mattermost account from your GitHub account\n" +
```
larkox, line 273: Might be an overkill, but should we extract from here the trigger name? (i.e. change all "/github" by a variable trigger name). It will probably not change, but if it were to change, we wouldn't have to update all the lines here.

hanzei: What does this add as value? Is the a use case for not using `/github` as trigger?

Issue #225, opened 2020-03-25
Indicate if a PR has a conflict


hanzei:
The new RHS contains a little check mark that indicates that all GH check are green. It would be handy if there is an indication for merge conflicts, so users can start resolving them soon.

@asaadmahmood @larkox Ideas on how this could look like?

larkox:
This is how GitHub show it
![Screenshot from 2020-03-25 13-12-24](https://user-images.githubusercontent.com/1933730/77535177-75e23780-6e9a-11ea-9c43-7ee158739637.png)
How about the gray warning sign with a tooltip "merge required"?

hanzei:
How would the gray warning sign look like?

larkox:
Something like this:
![Screenshot from 2020-03-25 15-07-13](https://user-images.githubusercontent.com/1933730/77545246-74b90680-6eaa-11ea-89aa-fddd6a8caa14.png)

PR #291, opened 2020-05-29
Indicate if a PR has a conflict

hanzei:
#### Summary
Indicate on RHS, if the PR as a conflict. Please note that the PR status is always `pending` if there is a conflict. But that issue was present before this changes and is unrelated.

Screenshot from 2020-05-29 07-34-48

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/225

### DIFF
--- a/server/api.go
+++ b/server/api.go
@@ -45,6 +45,7 @@ type PRDetails struct {
 	URL                string                      `json:"url"`
 	Number             int                         `json:"number"`
 	Status             string                      `json:"status"`
+	Mergeable          bool                        `json:"mergeable"`
 	RequestedReviewers []*string                   `json:"requestedReviewers"`
 	Reviews            []*github.PullRequestReview `json:"reviews"`
 }
@@ -569,7 +570,8 @@ func (p *Plugin) getPrsDetails(w http.ResponseWriter, r *http.Request, userID st
 }
 
 func fetchPRDetails(ctx context.Context, client *github.Client, prURL string, prNumber int) *PRDetails {
-	status := ""
+	var status string
+	var mergeable bool
 	// Initialize to a non-nil slice to simplify JSON handling semantics
 	requestedReviewers := []*string{}
 	var reviewsList []*github.PullRequestReview = []*github.PullRequestReview{}
@@ -599,6 +601,9 @@ func fetchPRDetails(ctx context.Context, client *github.Client, prURL string, pr
 			mlog.Error(err.Error())
 			return
 		}
+
+		mergeable = prInfo.GetMergeable()
+
 		for _, v := range prInfo.RequestedReviewers {
 			requestedReviewers = append(requestedReviewers, v.Login)
 		}
@@ -615,6 +620,7 @@ func fetchPRDetails(ctx context.Context, client *github.Client, prURL string, pr
 		URL:                prURL,
 		Number:             prNumber,
 		Status:             status,
+		Mergeable:          mergeable,
 		RequestedReviewers: requestedReviewers,
 		Reviews:            reviewsList,
 	}
--- a/webapp/src/components/sidebar_right/github_items.jsx
+++ b/webapp/src/components/sidebar_right/github_items.jsx
@@ -6,7 +6,6 @@ import PropTypes from 'prop-types';
 
 import {Badge, Tooltip, OverlayTrigger} from 'react-bootstrap';
 import {makeStyleFromTheme, changeOpacity} from 'mattermost-redux/utils/theme_utils';
-import Octicon, {Alert} from '@primer/octicons-react';
 
 import {formatTimeSince} from 'utils/date_utils';
 
@@ -116,7 +115,10 @@ function GithubItems(props) {
                         </Tooltip>
                     }
                 >
-                    <i className='icon icon-alert-outline'/>
+                    <i
+                        style={style.conflictIcon}
+                        className='icon icon-alert-outline'
+                    />
                 </OverlayTrigger>
             );
         }
@@ -180,10 +182,13 @@ const getStyle = makeStyleFromTheme((theme) => {
         icon: {
             top: 3,
             position: 'relative',
-            left: 6,
+            left: 3,
             height: 18,
             display: 'inline-flex',
             alignItems: 'center',
+        },
+        conflictIcon: {
+            color: theme.errorTextColor,
         },
         milestoneIcon: {
             top: 3,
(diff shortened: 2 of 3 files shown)


server/api.go
```diff
@@ -569,7 +570,8 @@ func (p *Plugin) getPrsDetails(w http.ResponseWriter, r *http.Request, userID st
 }
 
 func fetchPRDetails(ctx context.Context, client *github.Client, prURL string, prNumber int) *PRDetails {
-	status := ""
+	var status string
```
hanzei, line 573: It's idiomatic to use `var` if a variable should be assigned later on.

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -103,14 +104,37 @@ function GithubItems(props) {
             }
         }
 
+        let hasConflict = '';
+        if (item.mergeable != null && !item.mergeable) {
```
hanzei, line 108: Is there a better syntax for this check?

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -103,14 +104,37 @@ function GithubItems(props) {
...
+                    key='githubRHSPRMergeableIndicator'
+                    placement='top'
+                    overlay={
+                        <Tooltip id='githubRHSPRMergeableTooltip'>
+                            {'This pull request has conflicts that must be resolved'}
+                        </Tooltip>
+                    }
+                >
+                    <span>
+                        <Octicon
```
larkox, line 120: (0/5) The way this icon is used for conflicts, the color is not pure black, but (according to my browser) #6a737d. Maybe using that color would look better.

Screenshot from 2020-05-29 10-36-33

webapp/src/components/sidebar_right/github_items.jsx
larkox: A change on the variable name (and assignment logic) to "item.mustMerge"?

webapp/src/components/sidebar_right/github_items.jsx
larkox: Also the size looks a bit off to my eye, but I am not 100%. Maybe involve UX?

webapp/src/components/sidebar_right/github_items.jsx
hanzei: What makes `mustMerge` a better variable name then `mergeable'?

webapp/src/components/sidebar_right/github_items.jsx
hanzei: @asaadmahmood :point_up_2:

webapp/src/components/sidebar_right/github_items.jsx
larkox: It just makes the check simpler. Bet it null or false, you can skip creating the conflict icon by just checking `if (item.mustMerge) {`

hanzei: I not sure I get your suggestion. Do you mean renaming `PRDetails.Mergeable` to `PRDetails.MustMerge` in `server` and `webapp`?

larkox: Yes. Starting from `PRDetails.MustMerge := !prInfo.GetMergeable()` all the way to the front-end. It will simplify this check and does not introduce any more code than a negation. Semantic/readability-wise is pretty similar too.

hanzei: Hm, I see. `MustMerge` is such a clear name to me. Maybe `hasConflict`?

larkox: LGTM

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -6,6 +6,7 @@ import PropTypes from 'prop-types';
 
 import {Badge, Tooltip, OverlayTrigger} from 'react-bootstrap';
 import {makeStyleFromTheme, changeOpacity} from 'mattermost-redux/utils/theme_utils';
+import Octicon, {Alert} from '@primer/octicons-react';
```
asaadmahmood, line 9: @hanzei we should use our Mattermost icons.
`<i class='icon icon-alert-outline'/>`

webapp/src/components/sidebar_right/github_items.jsx
marianunez: Agreed on size and possibly also spacing. Could you update the screenshot once updated with the mattermost icon as suggested by Asaad?

webapp/src/components/sidebar_right/github_items.jsx
hanzei: I've updated the icon. Does this look good to you?
Screenshot from 2020-06-03 12-00-51

asaadmahmood: @hanzei Looks a lot better, however, I feel like the spacing is slightly inconsistent on the right/left of the icon.
Can we make it equal (or visually equal)?

asaadmahmood: Also, can we make the icon use our `errorTextColor` from the theme.

webapp/src/components/sidebar_right/github_items.jsx
hanzei: After thinking a bit more about this, I'm wondering if the API provided by the plugin should be similar to the on from GitHub. Hence, `Mergeable` would be the correct name, given that it's also the name GitHub uses in there API. Let me know what you think.

webapp/src/components/sidebar_right/github_items.jsx
hanzei: @asaadmahmood I'm not sure about the color. It seams a bit to much for a warning. 1/5
Screenshot from 2020-06-03 12-18-00

asaadmahmood: I think it looks okay. A conflict usually is a blocking action, and needs immediate attention if the PR is to be merged, so I'm okay it being the error text.
The design looks good to me now.

hanzei: Would you mind approving the PR?

webapp/src/components/sidebar_right/github_items.jsx
marianunez: +1 on `Mergeable` being the correct name here.

Issue #300, opened 2020-06-08
Links to the issue/PR in Link Tooltip do not work on desktop app


hanzei:
There are three links in the tooltip that all point to the linked issue/PR:

- The repo name at the top (which I would personally expect to link to the repo, not the issue/PR)
- The issue/PR title
- The "See More" link below the description

Only the last link described above works on the desktop app. It opens the issue/PR in a new tab in the user's default browser. The first two links do not have any effect when clicked.

In the browser, the first two listed above navigate to GitHub in the same browser tab, rather than opening a new tab. The "See More" link, however, opens a new tab. I think we should have all of them open in a new tab, just as all links in markdown function. So the user does not navigate away from Mattermost.

Issue #299, opened 2020-06-08
Heading are not shown in link tooltip


hanzei:
_Originally posted by @mickmister in https://github.com/mattermost/mattermost-plugin-github/issues/297#issuecomment-640407676_

Issue #298, opened 2020-06-08
Linking an issue or pr in a private repo causes js warning


hanzei:
![Screenshot from 2020-06-08 09-34-17](https://user-images.githubusercontent.com/16541325/84004046-40ae9700-a96b-11ea-98ef-f4018469698b.png)

PR #301, opened 2020-06-08
Various tooltip fixes

hanzei:
#### Summary
This PR fixes a lot of issues with the link tooltip. I did split up the changes into individual commits, which correspond with the issues linked below.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/300
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/299
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/298
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/296
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/297


webapp/src/components/link_tooltip/tooltip.css
```diff
@@ -115,8 +115,10 @@
...
     font: .75em/2 SFMono-Regular,Consolas,Liberation Mono,Menlo,monospace;
     color: var(--blue);
-    white-space: nowrap;
     background-color: var(--light-blue);
     border-radius: 3px;
-    maxWidth: 140px;
+    max-width: 140px;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
```
hanzei, line 123: @asaadmahmood This is how the long branch names look now. Please let me know if this is fine.

Screenshot from 2020-06-08 10-55-55

webapp/src/components/link_tooltip/link_tooltip.jsx
```diff
@@ -91,8 +91,12 @@ export const LinkTooltip = ({href, connected}) => {
                     <div className='header mb-1'>
                         <a
                             title={data.repo}
-                            href={href}
-                        >{data.repo}</a>
+                            href={data.base.repo.html_url}
```
hanzei, line 94: That is a bit of a hack, but I'm 90% sure it's works. `base` will always belong to the repo the PR was submitted against. Hence, it's fine to use it's `html_url`. Otherwise there would have to be another API call, which I want to avoid.

webapp/src/components/link_tooltip/tooltip.css
asaadmahmood: @hanzei Looks good. The description text however looks too light.

hanzei: @asaadmahmood sure. That issue is covered by https://github.com/mattermost/mattermost-plugin-github/issues/294

webapp/src/components/link_tooltip/tooltip.css
```diff
@@ -115,8 +115,10 @@
     padding: 0 5px;
     font: .75em/2 SFMono-Regular,Consolas,Liberation Mono,Menlo,monospace;
     color: var(--blue);
-    white-space: nowrap;
     background-color: var(--light-blue);
     border-radius: 3px;
-    maxWidth: 140px;
+    max-width: 140px;
+    text-overflow: ellipsis;
+    white-space: nowrap;
```
larkox, line 122: Any strong reason to move `white-space`? If not, maybe move it back to where it was, to have a cleaner diff.

hanzei: I like (1/5) having all rules related to the size and behaviour of the box grouped together.

DHaussermann:
@hanzei when I build and deploy this I see an issue where enabling the plugin is breaking the user side CSS of the app. 
```
react-dom.production.min.js:209 TypeError: Cannot read property 'repo' of undefined
    at lr (github_a546c25945e46fbc_bundle.js:22)
    at Vo (react-dom.production.min.js:153)
    at zs (react-dom.production.min.js:175)
    at pi (react-dom.production.min.js:263)
    at cu (react-dom.production.min.js:246)
    at iu (react-dom.production.min.js:246)
    at Zi (react-dom.production.min.js:239)
    at react-dom.production.min.js:123
    at t.unstable_runWithPriority (scheduler.production.min.js:19)
    at Ba (react-dom.production.min.js:122)
```
When I load the user side with the plugin running the CCS starts to render and then breaks if any of the channel content includes a GitHub link. 

CSS-Break

Disabling the plugin resolves the issue. Break does not occur when the channels contents does not include a GitHub plugin.
Can you see if this is something you can repo?

Steps:
- Deploy the plugin from this branch
- Open webapp to channel that does not contain GitHub link
- Enable the plugin
- On the user side post a link such as `https://github.com/mattermost/mattermost-plugin-agenda/pull/44`
**Observed** CSS breaks

hanzei:
Great observation @DHaussermann. My workaround to fetch the URL of the repository doesn't work for issues. I can think of three way to fix that:

1. Fetch the GitHub API for the repository.
	- Pro: Clean way
	- Con: 2x the API requests. There are already a lot.
2. String manipulation: The API does return a `repository_url` e.g. `https://api.github.com/repos/mattermosttest/mattermost-plugin-jira`. E.g. by splitting at `/` we could extract the URL. An alternative source is `html_url` e.g. `https://github.com/mattermosttest/mattermost-plugin-jira/issues/8"`.
	- Pro: No more API requests.
	- Con: Might break easily when GitHub changes there API
3. Remove link to repository from title. 
	- Pro: Easy implement. Use only what the API provides.
	- Con: Well, no more link to the repo, which users might needed.

@larkox @mickmister What do you think?

larkox:
I am not sure how often do you want to go to the repo instead of the issue itself. So unless there are a good amount of cases that I ignore where getting to the repo is important, I would just go with 3.

hanzei:
I don't think that happens often. Dropping the title link.

@DHaussermann PTAL

DHaussermann:
@hanzei  tt seems like this may only be a partial solution to this issue. Some links are now functional such as `https://github.com/mattermost/mattermost-plugin-github/pull/301` where as other links such as `https://github.com/mattermost/mattermost-plugin-ee-jira/pull/61` will break the CSS and make the webapp blank. Though the console now shows a different error. 
``` 
react-dom.production.min.js:198 TypeError: Cannot read property 'ref' of undefined
    at lr (github_89f690298ba0efab_bundle.js:22)
    at Ga (react-dom.production.min.js:150)
    at vo (react-dom.production.min.js:166)
    at ko (react-dom.production.min.js:179)
    at qi (react-dom.production.min.js:232)
    at Ui (react-dom.production.min.js:233)
    at Ts (react-dom.production.min.js:249)
    at vs (react-dom.production.min.js:248)
    at gs (react-dom.production.min.js:245)
    at Yi (react-dom.production.min.js:243)
ri @ main.9f98651ee0480b854c21.js:212
github_89f690298ba0efab_bundle.js:22 Uncaught (in promise) TypeError: Cannot read property 'ref' of undefined
    at lr (github_89f690298ba0efab_bundle.js:22)
    at Ga (react-dom.production.min.js:150)
    at vo (react-dom.production.min.js:166)
    at ko (react-dom.production.min.js:179)
    at qi (react-dom.production.min.js:232)
    at Ui (react-dom.production.min.js:233)
    at Ts (react-dom.production.min.js:249)
    at vs (react-dom.production.min.js:248)
    at gs (react-dom.production.min.js:245)
    at Yi (react-dom.production.min.js:243)
```
I realize that this link is to a private repo but other links to private repos don't break the CSS. The pop-over shows `Undefined` with a red symbol as the content (I can add en enhancement issue to improve this). 

Please let me know if this is something you're able to repro. Why are some of the private links causing the break?

mickmister:
@DHaussermann Are the other private repositories that are working owned by your GitHub profile? I think you need to specifically auth with Mattermost's GitHub org to access those repos, which can only be done with an OAuth app registered through a Mattermost GitHub admin.

If we find that the user cannot access the repo/issue, then I suggest we just don't show the tooltip.

DHaussermann:
@mickmister yes, it would make sense that it's hiding the data in some cases. But what I find confusing is that other links in the same repo don't break the CSS. See https://github.com/mattermost/mattermost-plugin-ee-jira/issues/49 for example. The app does not break when I post the link and I see the following:
UndefinedPop-over

This is a link to an issue and not a PR. But I wonder why the one link breaks the CSS and the other does not.

hanzei:
I did fix the issue with issue/PRs that are not found. @DHaussermann Please take another look.

DHaussermann:
@hanzei This seems resolved. I no longer see any links that cause the App to break. The linked issues appear resolved. Testing is ongoing. 
One question I have. It seems as thought when the user has no access such as a link to a private repo, there is no pop-over rather than the `Undefined` I was seeing before. This seems like a better UX. Please confirm if this is expected.

hanzei:
Yes, that is the expected behaviour.

hanzei:
@DHaussermann Indeed, that is the expected behaviour.

Issue #290, opened 2020-05-28
Add slash command autocomplete functionality


iomodo:
From 5.24 MM server has a slash command autocomplete functionality feature. An API can be found [here](https://github.com/mattermost/mattermost-server/blob/master/model/command_autocomplete.go#L28)
To implement the feature add `AutocompleteData` struct to the `Command` when registering a slash command.

Similar PRs:
For Todo plugin https://github.com/mattermost/mattermost-plugin-todo/pull/69
For Agenda plugin https://github.com/mattermost/mattermost-plugin-agenda/pull/40

NateWilliams2:
Hey I'm a first-time contributor and would love to take this issue!

hanzei:
Awesome, thanks for taking a shot @NateWilliams2 :+1: And welcome to the Mattermost Community. Please let me know, if you have any questions or need help.

NateWilliams2:
@hanzei thanks for the welcome :) 
I have a question- I'm just trying to set up my local mattermost server with the github plugin to test it, and I'm seeing I need to enable a public site url in order to authorize a Github OAuth client to use the plugin. 
Without this set up, I'm getting `Command with a trigger of '/github' not found`. Am I right in thinking I'll need to register a domain name just to test this plugin, or do you know of a way around this?

hanzei:
@NateWilliams2 indeed, you need a public domain for testing. I've been using https://ngrok.com/ for this. You might also want to give it a try.

NateWilliams2:
Awesome. Thanks!

PR #309, opened 2020-06-11
[GH-290] Add slash command autocomplete functionality

NateWilliams2:
#### Notes
- Optional command suggestions are hard to bypass as a user, as discussed in https://github.com/mattermost/mattermost-plugin-todo/pull/69
- I see no AutocompleteListItem analog where *multiple* of the list items can be chosen. Therefore this PR uses AddTextArgument type for manual entry of optional [features] for the subscribe command.
- `/github subscribe list` takes no arguments. `/github subscribe` takes three arguments, the first of which is a text argument. As discussed in https://github.com/mattermost/mattermost-plugin-todo/pull/69, "The current data model doesn't allow "" as Item". Therefore using `/github subscribe list` gives optional suggestions for additional arguments which are invalid for the list command. Solution could be replacing the `/github subscribe [options]` command with `/github subscribe add [options]` so we have a separate command from the list command. 

#### Ticket Link
  Fixes https://github.com/mattermost/mattermost-plugin-github/issues/290

### DIFF
--- a/server/command.go
+++ b/server/command.go
@@ -342,54 +342,20 @@ func getAutocompleteData() *model.AutocompleteData {
 	todo := model.NewAutocompleteData("todo", "", "Get a list of unread messages and pull requests awaiting your review")
 	github.AddCommand(todo)
 
-	// subscribe := model.NewAutocompleteData("subscribe", "[command/owner/repo] [features] [flags]", "Use command list or enter owner/repo name")
-	// subscribeList := model.NewAutocompleteData("list", "", "List the current channel subscriptions")
+	subscribe := model.NewAutocompleteData("subscribe", "[owner/repo] [features] [flags]", "Subscribe the current channel to receive notifications about opened pull requests and issues for an organization or repository. [features] and [flags] are optional arguments")
+	subscribe.AddTextArgument("Owner/repo to subscribe to", "[owner/repo]", "")
+	subscribe.AddTextArgument("(Optional) Comma-delimited list of one or more of: issues, pulls, pushes, creates, deletes, issue_comments, pull_reviews, label:\"<labelname>\". Detaults to pulls,issues,creates,deletes", "[features]", `/[^,-\s]+(,[^,-\s]+)*/`)
+	flags := []model.AutocompleteListItem{{
+		HelpText: "events triggered by organization members will not be delivered (the organization config should be set, otherwise this flag has not effect)",
+		Hint:     "(optional)",
+		Item:     "--exclude-org-member",
+	}}
+	subscribe.AddStaticListArgument("Currently supports --exclude-org-member", false, flags)
 
-	// subscribe.AddTextArgument("Subscribe the current channel to receive notifications about opened pull requests and issues for an organization or repository", "[owner/repo]", "")
-	// features := []model.AutocompleteListItem{{
-	// 	HelpText: "includes new and closed issues",
-	// 	Hint:     "(optional)",
-	// 	Item:     "issues",
-	// }, {
-	// 	HelpText: "includes new and closed pull requests",
-	// 	Hint:     "(optional)",
-	// 	Item:     "pulls",
-	// }, {
-	// 	HelpText: "includes pushes",
-	// 	Hint:     "(optional)",
-	// 	Item:     "pushes",
-	// }, {
-	// 	HelpText: "includes branch and tag creations",
-	// 	Hint:     "(optional)",
-	// 	Item:     "creates",
-	// }, {
-	// 	HelpText: "includes branch and tag deletions",
-	// 	Hint:     "(optional)",
-	// 	Item:     "deletes",
-	// }, {
-	// 	HelpText: "includes new issue comments",
-	// 	Hint:     "(optional)",
-	// 	Item:     "issue_comments",
-	// }, {
-	// 	HelpText: "includes pull request reviews",
-	// 	Hint:     "(optional)",
-	// 	Item:     "pull_reviews",
-	// }, {
-	// 	HelpText: "must include \"pulls\" or \"issues\" in feature list when using a label",
-	// 	Hint:     "(optional)",
-	// 	Item:     "label:\"<labelname>\"",
-	// }}
-	// subscribe.AddStaticListArgument("Defaults to \"pulls,issues,creates,deletes\"", false, features)
+	github.AddCommand(subscribe)
 
-	// flags := []model.AutocompleteListItem{{
-	// 	HelpText: "events triggered by organization members will not be delivered (the organization config should be set, otherwise this flag has not effect)",
-	// 	Hint:     "(optional)",
-	// 	Item:     "--exclude-org-member",
-	// }}
-	// subscribe.AddStaticListArgument("Currently supports --exclude-org-member", false, flags)
-	// subscribe.AddCommand(subscribeList)
-	//subscribe.AddCommand(subscribeOwner)
-	// github.AddCommand(subscribe)
+	subscribeList := model.NewAutocompleteData("subscribe list", "", "List the current channel subscriptions")
+	github.AddCommand(subscribeList)
 
 	unsubscribe := model.NewAutocompleteData("unsubscribe", "[owner/repo]", "Unsubscribe the current channel from an organization or repository")
 	unsubscribe.AddTextArgument("Owner/repo to unsubscribe from", "[owner/repo]", "")
--- a/server/command.go
+++ b/server/command.go
@@ -93,6 +93,7 @@ func getCommand() *model.Command {
 		AutoComplete:     true,
 		AutoCompleteDesc: "Available commands: connect, disconnect, todo, me, settings, subscribe, unsubscribe, help",
 		AutoCompleteHint: "[command]",
+		AutocompleteData: getAutocompleteData(),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


NateWilliams2:
The ci-linter is failing on 
```
golangci-lint run ./...
ERRO Running error: context loading failed: no go files to analyze 
make: *** [Makefile:47: golangci-lint] Error 5
```
Any ideas why?

server/command.go
```diff
@@ -323,3 +324,65 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+	unsubscribe.AddTextArgument("Owner/repo to unsubscribe from", "[owner/repo]", "")
+	github.AddCommand(unsubscribe)
+
+	me := model.NewAutocompleteData("me", "", "Display the connected GitHub account")
+	github.AddCommand(me)
+
+	settings := model.NewAutocompleteData("settings", "[setting] [value]", "Update your user settings")
+	setting := []model.AutocompleteListItem{{
+		HelpText: "Turn notifications on/off",
+		Hint:     "",
```
hanzei, line 367: Nit: Drop empty value
```suggestion
```

server/command.go
```diff
@@ -323,3 +324,65 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+		Item:     "notifications",
+	}, {
+		HelpText: "Turn reminders on/off",
+		Hint:     "",
```
hanzei, line 371: Nit: Same here
```suggestion
```

server/command.go
```diff
@@ -323,3 +324,65 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+	value := []model.AutocompleteListItem{{
+		HelpText: "Turn setting on",
+		Hint:     "",
+		Item:     "on",
+	}, {
+		HelpText: "Turn setting off",
+		Hint:     "",
+		Item:     "off",
+	}}
+	settings.AddStaticListArgument("New value", true, value)
```
hanzei, line 384: Where does the string `"New value` get shown in the UI? I can't find it.

server/command.go
```diff
@@ -323,3 +324,65 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+	flags := []model.AutocompleteListItem{{
+		HelpText: "events triggered by organization members will not be delivered (the organization config should be set, otherwise this flag has not effect)",
+		Hint:     "(optional)",
+		Item:     "--exclude-org-member",
+	}}
+	subscribe.AddStaticListArgument("Currently supports --exclude-org-member", false, flags)
+
+	github.AddCommand(subscribe)
+
+	subscribeList := model.NewAutocompleteData("subscribe list", "", "List the current channel subscriptions")
```
hanzei, line 354: When I type `/github subscribe list` and then press `enter`, `[feature]` is shown as suggested input.
Screenshot from 2020-06-15 11-24-14

@iomodo Could this be a bug in Mattermost?

server/command.go
```diff
@@ -323,3 +324,65 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+
+	subscribe := model.NewAutocompleteData("subscribe", "[owner/repo] [features] [flags]", "Subscribe the current channel to receive notifications about opened pull requests and issues for an organization or repository. [features] and [flags] are optional arguments")
+	subscribe.AddTextArgument("Owner/repo to subscribe to", "[owner/repo]", "")
+	subscribe.AddTextArgument("(Optional) Comma-delimited list of one or more of: issues, pulls, pushes, creates, deletes, issue_comments, pull_reviews, label:\"<labelname>\". Detaults to pulls,issues,creates,deletes", "[features]", `/[^,-\s]+(,[^,-\s]+)*/`)
+	flags := []model.AutocompleteListItem{{
+		HelpText: "events triggered by organization members will not be delivered (the organization config should be set, otherwise this flag has not effect)",
+		Hint:     "(optional)",
+		Item:     "--exclude-org-member",
+	}}
+	subscribe.AddStaticListArgument("Currently supports --exclude-org-member", false, flags)
```
hanzei, line 350: Does it make sense to show this option, if no organisation is configured?

server/command.go
```diff
@@ -323,3 +324,65 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+
+	todo := model.NewAutocompleteData("todo", "", "Get a list of unread messages and pull requests awaiting your review")
+	github.AddCommand(todo)
+
+	subscribe := model.NewAutocompleteData("subscribe", "[owner/repo] [features] [flags]", "Subscribe the current channel to receive notifications about opened pull requests and issues for an organization or repository. [features] and [flags] are optional arguments")
+	subscribe.AddTextArgument("Owner/repo to subscribe to", "[owner/repo]", "")
+	subscribe.AddTextArgument("(Optional) Comma-delimited list of one or more of: issues, pulls, pushes, creates, deletes, issue_comments, pull_reviews, label:\"<labelname>\". Detaults to pulls,issues,creates,deletes", "[features]", `/[^,-\s]+(,[^,-\s]+)*/`)
+	flags := []model.AutocompleteListItem{{
+		HelpText: "events triggered by organization members will not be delivered (the organization config should be set, otherwise this flag has not effect)",
+		Hint:     "(optional)",
```
hanzei, line 347: Here `optional` in the `Hint` and for the `Features` it's in the `HelpText`. What is the one  place where this word should be? @iomodo is there a guideline?

server/command.go
```diff
@@ -323,3 +324,65 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+
+	disconnect := model.NewAutocompleteData("disconnect", "", "Disconnect your Mattermost account from your GitHub account")
+	github.AddCommand(disconnect)
+
+	todo := model.NewAutocompleteData("todo", "", "Get a list of unread messages and pull requests awaiting your review")
+	github.AddCommand(todo)
+
+	subscribe := model.NewAutocompleteData("subscribe", "[owner/repo] [features] [flags]", "Subscribe the current channel to receive notifications about opened pull requests and issues for an organization or repository. [features] and [flags] are optional arguments")
+	subscribe.AddTextArgument("Owner/repo to subscribe to", "[owner/repo]", "")
+	subscribe.AddTextArgument("(Optional) Comma-delimited list of one or more of: issues, pulls, pushes, creates, deletes, issue_comments, pull_reviews, label:\"<labelname>\". Detaults to pulls,issues,creates,deletes", "[features]", `/[^,-\s]+(,[^,-\s]+)*/`)
```
hanzei, line 344: Maybe out of scope of this: I would be awesome if that list of feature could be autocompleted!

hanzei:
@NateWilliams2 The broken CI was caused by `go.sum` not being up to date: https://github.com/golangci/golangci-lint/issues/825

I did fix this myself.

server/command.go
NateWilliams2: I mention this in the PR notes- because `/github subscribe [owner/repo] [features]` is its own command, the autocomplete parses `/github subscribe list` as the subscribe command with `list` as the owner/repo field.

server/command.go
iomodo: No guidelines yet. I think we should see "in the wild" which will work best. And add guidelines later.

server/command.go
iomodo: Not a bug, for autocomplete there is no way to detect the difference. Clean fix will be to use `/github subscriptions` command to list them.

server/command.go
NateWilliams2: Good question.. I'm not aware of any way in the Autocomplete API to show a suggestion conditional on the current state of the plugin. Would you suggest I turn this field into a text field so users must manually type the flag they would like to use, and therefore this option will not be automatically filled?

server/command.go
NateWilliams2: Good point. I don't see any of the `helpText` strings in `AddStaticListArgument` showing up in the ui. It seems like this is the design on this type of field.

hanzei:
@NateWilliams2 Would you please update the PR based on the consensus in the GitLab plugin: https://github.com/mattermost/mattermost-plugin-gitlab/pull/182#issuecomment-654401432?

NateWilliams2:
sure thing @hanzei ! Would I need to figure out how to do this for the actual command itself, or do I only need to make the changes to the autocomplete functionality?

hanzei:
You also have to touch the code that handles the command itself. https://github.com/mattermost/mattermost-plugin-gitlab/pull/182/commits/6fb798cd5d83c8b6a6dea8894dee8a304ba1c394 is the commit that did that for GitLab. It might be a good starting point.

jasonblais:
@NateWilliams2 Thanks for your help on this PR! Let us know if you have any questions about the feedback above :)

hanzei:
@NateWilliams2 I would love to get this feature into the next release. Can I help you in any way? If you are short on time at the moment, would you be fine if someone else takes over?

NateWilliams2:
Hey @hanzei ! Apologies, I've been busy lately and haven't had time to work on this. Shoulda let you know earlier. I'd be fine if someone else takes over.

hanzei:
I've update the PR to include the changes from the GitLab plugin. Thanks for the groundwork @NateWilliams2 :+1: 

@iomodo @mickmister Please take a look.

NateWilliams2:
Yep! Thanks for continuing it @hanzei

DHaussermann:
@NateWilliams2 thanks for implementing this. It's very nice to use!
I have done some smoke testing and checked over the auto-complete
Just a couple small things... 
1. Seems like the `help` sub-command  does not seem to be included .
2. For `/github connect`  I think, the `private` option should have `(optional)` at the end to be consistent with this and other plugins. Thoughts?

hanzei:
@DHaussermann Thanks for the two suggestions :+1: PTAL

Issue #266, opened 2020-05-13
Don't try to autolink messages from bots


hanzei:
When a message gets posted by a bot and `EnableCodePreview` is enabled, the plugin tries to get the GitHub token for that user, which fails. Use [`ShouldProcessMessage`](https://developers.mattermost.com/extend/plugins/server/reference/#Helpers.ShouldProcessMessage) to skip posts from bots and system messages.

dbejanishvili:
I'll do this if it is still available

hanzei:
All your's @dbejanishvili :+1:

PR #314, opened 2020-06-20
Ignored bot and system messages in messageWillBePosted hook

dbejanishvili:
#### Summary
Used ShouldProcessMessage to skip posts from bots and system messages.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/266


server/plugin.go
```diff
@@ -162,6 +162,16 @@ func (p *Plugin) MessageWillBePosted(c *plugin.Context, post *model.Post) (*mode
 		return nil, ""
 	}
 
+	shouldProcessMessage, e := p.Helpers.ShouldProcessMessage(post)
```
hanzei, line 165: Nit pick: Using `e` as a name for an `error` is quite uncommon. I would suggest using `error` and renaming the `*model.Apperror` in L176 into `appError`.

dbejanishvili: yeah, I thought that would be better too, I just tried not to change any outside code

DHaussermann:
@larkox or @hanzei, If you have a bit of time, could one of you peer-test this PR using the  peer testing process documented here https://mattermost.atlassian.net/wiki/spaces/INT/pages/619937892/Peer+Testing+Process

hanzei:
@larkox Could you please take this one? I looking into the other two.

larkox:
Any message posted by a user not registered to github would also throw the error. Do we want to avoid that?

Also, I am not sure about the privacy of this feature. Since the poster token is used, potentially it may show to anyone content of private repositories. Do we want to avoid this here too? Or is this handled in a different part of the code?

larkox:
@hanzei :point_up:

hanzei:
> Any message posted by a user not registered to github would also throw the error. Do we want to avoid that?

While that is not a new behaviour, it defiantly spammy. An error should only be logged if `getGitHubUserInfo` returned an error different to `apiErrorIDNotConnected`. @dbejanishvili Would you be interested in also fixing that issue in this PR?

> Also, I am not sure about the privacy of this feature. Since the poster token is used, potentially it may show to anyone content of private repositories. Do we want to avoid this here too? Or is this handled in a different part of the code?

@larkox That is a very valid concern, but out of scope of this PR. Would you mind opening a new issue for that? We probably need also input from the security team.

dbejanishvili:
Sure

hanzei:
@larkox This is ready for another round of QA review

Issue #259, opened 2020-05-05
Add API route to allow other plugins to fetch the API key


hanzei:
https://github.com/mattermost/mattermost-server/pull/12829 added support for inter plugin http requests. This feature should be used to allow other plugins to fetch the oauth2 token for a given user. Using that token other plugins are able to integrate with the GitHub API.

This API endpoint should be accessible that `/api/v1/token` and only plugin are allowed to access it. If a request was made from a plugin can be checked by fetching the `Mattermost-Plugin-ID` http header.

A new method should be exposed that allows an easy integration of this feature into other plugin. A `client` package should be exposed with the following methods:
- `func NewClientPlugin(api PluginAPI) *Client`
- `func (c *Client) GetToken(userID string) (string, error)`
- [`func (c *Client) GetConfig() (*Configuration, error)`]((https://github.com/mattermost/mattermost-plugin-github/blob/119ff1807a40eca316e5c1644589f056baa71936/server/configuration.go#L21-L31))

Optionally, write a simple plugin that makes use of this feature to allow easy testing.

PR #315, opened 2020-06-24
Add API for other plugins to use

hanzei:
#### Summary
Expose two API endpoints for other plugins to use:
- GetConfig
- GetToken

For seamless integration into other plugins a new `client` package is exposed. In order to keep the code DRY. all plugin code expect `main.go` got moved into a separate package.

Also, a  `GetGitHubClient` method is added to avoid copy&pasting the same code in multiple plugins.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/259


server/plugin/api.go
```diff
@@ -1,4 +1,4 @@
-package main
+package plugin
```
hanzei, line 1: I opted for `plugin` as it's a general term that can be used in other plugins. Another consideration was `app`, as it's also used in the server.

server/plugin/api_test.go
```diff
@@ -0,0 +1,201 @@
...
+		Encoder: testutils.EncodeString,
+	}
+
+	for name, test := range map[string]struct {
+		httpTest         testutils.HTTPTest
+		request          testutils.Request
+		context          *plugin.Context
+		expectedResponse testutils.ExpectedResponse
+	}{
+		"not authorized": {
```
hanzei, line 97: Adding another tests turned out to be a PITA. So, there is only one test.

server/plugin/api.go
```diff
@@ -1109,3 +1126,25 @@ func (p *Plugin) createIssue(w http.ResponseWriter, r *http.Request, userID stri
 
 	p.writeJSON(w, result)
 }
+
+func (p *Plugin) getConfig(w http.ResponseWriter, r *http.Request) {
+	config := p.getConfiguration()
+
+	p.writeJSON(w, config)
```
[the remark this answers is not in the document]
hanzei: Both API endpoints can only be accessed by other plugins, as `plugin.Context.SourcePluginId` is only set, if the request is made by another plugin.

@larkox What is the attack vector we are talking about? A malicious plugin being deployment on the same server?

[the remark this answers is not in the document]
larkox: I was more thinking about someone external impersonating a plugin to get tokens or configuration. How reliable is the Context.SourcePluginId? It is something I can fake in any way?

levb:
Replaced myself with @crspeller for security review. I am generally ok with the mechanics of the implementation.

server/plugin/api.go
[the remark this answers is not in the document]
hanzei: `Context.SourcePluginId` is only set, if another plugins does an HTTP request using `api.PluginHTTP`. There is no way to fake it using e.g. `curl`.

server/client/client.go
```diff
@@ -0,0 +1,125 @@
...
+func (p *pluginAPIRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	resp := p.api.PluginHTTP(req)
+	if resp == nil {
+		return nil, errors.Errorf("Failed to make interplugin request")
+	}
+
+	return resp, nil
+}
+
+func NewClientPlugin(api PluginAPI) *Client {
```
mickmister, line 37: Does it make sense to call this `NewPluginClient`?

server/plugin/api.go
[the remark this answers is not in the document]
mickmister: @larkox Looks like it's set here https://github.com/mattermost/mattermost-server/blob/4c33b7a35d3129f8b63a86c35e738282dc2214a0/app/plugin_requests.go#L70 . This is similar how the plugins can trust that if the `Mattermost-User-Id` header is set, then the request was vetted with a valid user session.

server/client/client.go
hanzei: :+1: from me. @crspeller what do you think?

crspeller: Ya, that sounds better.

plugin.json
```diff
@@ -6,7 +6,7 @@
     "support_url": "https://github.com/mattermost/mattermost-plugin-github/issues",
     "icon_path": "assets/icon.svg",
     "version": "1.0.0",
-    "min_server_version": "5.12.0",
+    "min_server_version": "5.25.0",
```
crspeller, line 9: Why so high? `PluginHTTP` only needs 5.18.

hanzei: Because the bug fix in https://github.com/mattermost/mattermost-server/pull/14905 is needed for the plugin to work.

server/client/client.go
hanzei: Changed :+1:

hanzei:
@DHaussermann You can test this PR in the same way as https://github.com/mattermost/mattermost-server/pull/14905

DHaussermann:
@hanzei I'm seeing an issue on this one where the plugin crashes when deployed due to minimum version not being met.

I see 
```
{"level":"info","ts":1594395927.3588865,"caller":"cluster/cluster.go:332","msg":"This node attempted to join the cluster, but could not find any nodes."}
{"level":"error","ts":1594395930.4746335,"caller":"mlog/log.go:175","msg":"Unable to restart plugin on upgrade.","path":"/api/v4/plugins","request_id":"fnrezpboo38r5x67xqxob15bph","ip_addr":"10.160.99.229","user_id":"s1jrzjbtsfffzjdkkjeztfdyue","method":"POST","err_where":"installExtractedPlugin","http_code":500,"err_details":"failed to register commands: failed to register agenda command: invalid autocomplete data in command: An empty command name in the autocomplete data"}
```

The server is running `5.25.0-rc1` 

Any thoughts to help isolate this? Could this be am incorrect error?

hanzei:
@DHaussermann `v5.25.0-rc1` does incorrectly report it's own version as `5.24.0`. This bug was fixed in https://github.com/mattermost/mattermost-server/pull/14948 and will not longer be present in `v5.25.0-rc2`. Could you please test with  `v5.25.0-rc2` or `master`?

PR #316, opened 2020-06-26
Fix Github > GitHub capitalization error

jasonblais:
0/5, is it possible to have a build check of some kind that detects an incorrect capitalization in future?


hanzei:
> 0/5, is it possible to have a build check of some kind that detects an incorrect capitalization in future?

Detecting these error might cause a lot of false positives. 1/5 that it's more mana to sort them out as it's to just fix the error when someone finds them manally.

Issue #267, opened 2020-05-13
Post a notification if issue gets reopened


hanzei:
When a issue gets closed and then reopened, only a notification for closing the issue is send to a channel that subscribed to that repository. There should also be a notification saying that the issue got reopened. The wording should be:
`{{template "repo" .GetRepo}} Issue {{template "issue" .GetIssue}} reopened by {{template "user" .GetSender}}.`

checkaayush:
Isn't this what this part of the code (https://github.com/mattermost/mattermost-plugin-github/blob/master/server/webhook.go#L878) does?

hanzei:
@checkaayush That part of the code deals with the notification to a user, not to channels. Per user notifications for this event work fine.

PR #317, opened 2020-07-02
send notification when issue gets reopened

mustafayildirim:
#### Summary
Send notification when issue gets reopened

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/267


server/webhook.go
```diff
@@ -373,6 +373,12 @@ func (p *Plugin) postIssueEvent(event *github.IssuesEvent) {
 		return
 	}
 
+	reopenedIssueMessage, err := renderTemplate("reopenedIssue", event)
```
larkox, line 376: Outside the scope of this PR:
Why are we creating all the messages, if we are only going to use one? Wouldn't it make sense to check the action before we reach the for?

mustafayildirim: Hi @larkox 
I realized that before adding third one. I though that's the way.

I can refactor this code. 
I would be glad to do if it is suitable for you

larkox: I would like to hear the opinion of @hanzei before moving forward with that refactor.

hanzei: It totally makes sense to refactor the code :+1: Thanks for taking care of that @mustafayildirim

mustafayildirim: I've refactored template rendering. @larkox @hanzei

larkox: One more improvement on the refactor. Can we use the switch to decide just the template to use, and the rest of the repeated code outside of the switch (or in a function)?

mustafayildirim: I fixed

PR #319, opened 2020-07-03
Quote newlines within body

### LINKED ISSUE #281: Fix templates to properly wrap the whole body with quotes
Several templates are formed like this:

`>{{.GetPullRequest.GetBody | trimBody | replaceAllGitHubUsernames}}`

If the body had several line breaks, this may format the body as:
```
> A

B
```
This will make the first line be quoted but the second not. In order to avoid this, every line break should have a `>`. i.e.:
```
> A
>
> B
```

This changes should be covered by a lot of unit tests to ensure no edge cases are missed.

ofpiyush:
#### Summary

New lines in body used to break quoting, now it doesn't!

Tests updated to reflect.

#### Ticket Link

Fixes #281 

The `Difficulty/3:Hard` label is making me feel like I might have misunderstood the problem. If so, I'll correct it. Thanks!

### DIFF
--- a/server/template.go
+++ b/server/template.go
@@ -76,6 +76,11 @@ func init() {
 		})
 	}
 
+	// Quote the body
+	funcMap["quote"] = func(body string) string {
+		return ">" + strings.ReplaceAll(body, "\n", "\n>")
+	}
+
 	masterTemplate = template.Must(template.New("master").Funcs(funcMap).Parse(""))
 
 	// The user template links to the corresponding GitHub user. If the GitHub user is a known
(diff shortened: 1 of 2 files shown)


ofpiyush:
Do you need me to do something @hanzei?

DHaussermann:
@hanzei or @larkox, If you have a bit of time, could one of you peer-test this PR using the  peer testing process documented here https://mattermost.atlassian.net/wiki/spaces/INT/pages/619937892/Peer+Testing+Process

hanzei:
@DHaussermann :point_up:

ofpiyush:
Thanks!

PR #320, opened 2020-07-05
refactored extractUserMiddleWare

### LINKED ISSUE #275: Replace bool with new type in extractUserMiddleWare to make code more readable
I think it would be better if `jsonResponse bool` has its separate type something like `ResponseType.Json`
https://github.com/mattermost/mattermost-plugin-github/blob/f3ed30562256da481badc929f6fbda50172c49fb/server/api.go#L90
because these lines of code seem hard too read
https://github.com/mattermost/mattermost-plugin-github/blob/f3ed30562256da481badc929f6fbda50172c49fb/server/api.go#L74

Thoughts?

bakurits:
#### Summary
Refactored extractUserMiddleWare in api.go so it no longer accepts boolean and has a separate type for content type

#### Ticket Link
Fixes #275


server/api.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const (
 	apiErrorIDNotConnected = "not_connected"
-	// the OAuth token expiry in seconds
+	//TokenTTL the OAuth token expiry in seconds
```
hanzei, line 24:
```suggestion
	// TokenTTL is the OAuth token expiry duration in seconds
```

server/api.go
```diff
@@ -50,8 +50,19 @@ type PRDetails struct {
...
 type HTTPHandlerFuncWithUser func(w http.ResponseWriter, r *http.Request, userID string)
 
+// ResponseType indicates type of response returned by api
+type ResponseType string
+
+const (
+	// ResponseTypeJSON indicates that response type is json
+	ResponseTypeJSON ResponseType = "JSON_RESPONSE"
+	// ResponseTypePlain indicates that response type is text plain
+	ResponseTypePlain = "TEXT_RESPONSE"
```
hanzei, line 63: Should this be 
```suggestion
	ResponseTypePlain ResponseType = "TEXT_RESPONSE"
```
?

server/api.go
```diff
@@ -93,34 +104,35 @@ func (p *Plugin) initializeAPI() {
...
 	return func(w http.ResponseWriter, r *http.Request) {
 		userID := r.Header.Get("Mattermost-User-ID")
 		if userID == "" {
-			if jsonResponse {
+			switch responseType {
+			case ResponseTypeJSON:
 				p.writeAPIError(w, &APIErrorResponse{ID: "", Message: "Not authorized.", StatusCode: http.StatusUnauthorized})
-			} else {
+			case ResponseTypePlain:
 				http.Error(w, "Not authorized", http.StatusUnauthorized)
```
hanzei, line 136: Does adding a `default` case make sense to catch a developers misusing the responseType?

bakurits: How can we do that? Just log something in default case or you have a better way

hanzei: Logging an error is totally fine.

server/plugin/api.go
```diff
@@ -21,7 +21,7 @@ import (
 
 const (
 	apiErrorIDNotConnected = "not_connected"
-	// the OAuth token expiry in seconds
+	//TokenTTL  is the OAuth token expiry duration in seconds
```
larkox, line 24: Please, put an space between the `//` and the first word.
Also, there are two spaces between `TokenTTL` and `is`. Remove one.

server/plugin/api.go
```diff
@@ -93,38 +104,41 @@ func (p *Plugin) initializeAPI() {
...
 		if userID == "" {
-			if jsonResponse {
+			switch responseType {
+			case ResponseTypeJSON:
 				p.writeAPIError(w, &APIErrorResponse{ID: "", Message: "Not authorized.", StatusCode: http.StatusUnauthorized})
-			} else {
+			case ResponseTypePlain:
 				http.Error(w, "Not authorized", http.StatusUnauthorized)
+			default:
+				mlog.Error("Unknown ResponseType detected")
```
larkox, line 141: Do you mind using `p.API.LogError(...)` instead? The file is not very consistent on how to show errors, but `p.API.Log...` should be the standard.

DHaussermann:
@hanzei or @larkox  If you have a bit of time, could one of you peer-test this PR using the  peer testing process documented here https://mattermost.atlassian.net/wiki/spaces/INT/pages/619937892/Peer+Testing+Process

larkox:
@hanzei If you want, I can take this.

larkox:
@DHaussermann Peer testing done. Results above.

PR #335, opened 2020-07-27
[GH-257] Display labels and assignee on new issues and PRs

Szymongib:
#### Summary
Display labels and assignee when creating new Issue or PR.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-github/issues/257

### DIFF
--- a/server/plugin/template.go
+++ b/server/plugin/template.go
@@ -2,6 +2,7 @@ package plugin
 
 import (
 	"bytes"
+	"net/url"
 	"regexp"
 	"strings"
 	"text/template"
@@ -81,6 +82,27 @@ func init() {
 		return ">" + strings.ReplaceAll(body, "\n", "\n>")
 	}
 
+	// Escape characters not allowed in URL path
+	funcMap["pathEscape"] = func(s string) string {
+		return url.PathEscape(s)
+	}
+
+	// Transform multiple variables to dictionary
+	funcMap["dict"] = func(values ...interface{}) (map[string]interface{}, error) {
+		if len(values)%2 != 0 {
+			return nil, errors.New("invalid dict call, exactly one value is required for every key")
+		}
+		dict := make(map[string]interface{}, len(values)/2)
+		for i := 0; i < len(values); i+=2 {
+			key, ok := values[i].(string)
+			if !ok {
+				return nil, errors.New("dict keys must be strings")
+			}
+			dict[key] = values[i+1]
+		}
+		return dict, nil
+	}
+
 	masterTemplate = template.Must(template.New("master").Funcs(funcMap).Parse(""))
 
 	// The user template links to the corresponding GitHub user. If the GitHub user is a known
@@ -151,7 +173,7 @@ func init() {
 
 	template.Must(masterTemplate.New("labels").Funcs(funcMap).Parse(`
 {{- if .Labels }}
-Labels: {{range .Labels -}}` + "`[{{ .Name }}]({{ .URL }})` " + `{{end -}}
+Labels: {{range .Labels -}}` + "[`{{ .Name }}`]({{ $.RepositoryURL }}/labels/{{ .Name | pathEscape }}) " + `{{end -}}
 {{ end -}}
 `))
 
@@ -165,7 +187,7 @@ Assignee: [{{ .GetAssignee.GetLogin }}]({{ .GetAssignee.GetHTMLURL }})
 #### {{.GetPullRequest.GetTitle}}
 ##### {{template "eventRepoPullRequest" .}}
 #new-pull-request by {{template "user" .GetSender}}
-{{- template "labels" .GetPullRequest }}
+{{- template "labels" dict "Labels" .GetPullRequest.Labels "RepositoryURL" .GetRepo.GetHTMLURL  }}
 {{- template "assignee" .GetPullRequest }}
 
 {{.GetPullRequest.GetBody | removeComments | replaceAllGitHubUsernames}}
@@ -192,7 +214,7 @@ Assignee: [{{ .GetAssignee.GetLogin }}]({{ .GetAssignee.GetHTMLURL }})
 #### {{.GetIssue.GetTitle}}
 ##### {{template "eventRepoIssue" .}}
 #new-issue by {{template "user" .GetSender}}
-{{- template "labels" .GetIssue }}
+{{- template "labels" dict "Labels" .GetIssue.Labels "RepositoryURL" .GetRepo.GetHTMLURL  }}
 {{- template "assignee" .GetIssue }}
 
 {{.GetIssue.GetBody | removeComments | replaceAllGitHubUsernames}}
--- a/server/plugin/template.go
+++ b/server/plugin/template.go
@@ -83,9 +83,7 @@ func init() {
 	}
 
 	// Escape characters not allowed in URL path
-	funcMap["pathEscape"] = func(s string) string {
-		return url.PathEscape(s)
-	}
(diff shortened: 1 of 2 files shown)


Szymongib:
Hi @mickmister, 
I have a couple of follow up questions to finalize PR:
- Do you have any more suggestions regarding how labels and assignee should be displayed?
Currently, it looks like that:
image
- Do you think those should be displayed in case of any other event as well?

mickmister:
@Szymongib A quick pass through the code looks good :+1:

> Do you have any more suggestions regarding how labels and assignee should be displayed?

Do you mind posting some screenshots of the entire post, one with lots of summary/description text, and with a small amount?

> Do you think those should be displayed in case of any other event as well?

I think the additions would make sense on new comment posts as well, that are sent to a subscribed channel. As opposed to a DM notification from a comment.
@asaadmahmood What do you think?

What do you think about showing the other fields that can be set on an issue/PR? Such as Projects, Milestones, and Linked issues/PRs?

mickmister:
I'm thinking this has a lot of overlap with the plugin tooltip display, but this PR's functionality would show on mobile. Also, a post by the GitHub bot is more significant than a single link to an issue/PR, which is the scenario for the tooltip.

We may want to keep it simple with just labels and assignees. @Szymongib I'm wondering how it looks in the mobile app with more fields showing

asaadmahmood:
@mickmister @Szymongib Let's get a full screenshot of the channel/center channel to see how it looks.
Also @mickmister please help me explain this a bit more. I'm assuming this information will be displayed as a message to the channel (if subscribed) or DM.

What would be the case where this appears as a comment?

mickmister:
@asaadmahmood 
> Also @mickmister please help me explain this a bit more. I'm assuming this information will be displayed as a message to the channel (if subscribed) or DM.

Right now, it is showing for new PRs and new issues. If a channel is subscribed to pull request or issue creation, there will be posts in that channel and the labels will show up with this PRs functionality.

> What would be the case where this appears as a comment?

It's just a thought of having them show for comments as well. Channels can subscribe to GitHub comment events, and the labels could be shown with those posts. DM notifications could also have them, but I think that will be bloated for the GitHub DM feed.

asaadmahmood:
@mickmister I think it makes sense for issue creation and Pull requests, but I think for every comment, it may not be the best for us, as we may get a duplication on that most of the time.

I feel like we can have a separate event (if we don't have that right now) for label updates or assignee updates that can be used if people want to record those activities as well.

mickmister:
@asaadmahmood Currently there is no event for label or assignee changes. Oftentimes, labels are added by the reporter  after issue creation. So having an event for that would be nice. It may be within scope of task since it is directly related to this use case.

Szymongib:
@mickmister I think showing additional information like Project or Milestone could be added later as an enhancement to first see feedback from adding labels and assignee.

Here are some screenshots:

long issue


short issue


PR:

PR


and from the mobile app (IOS):

long issue


short issue


PR

mickmister:
@asaadmahmood Here are some screenshots ^

@Szymongib Placement of the new items looks good to me

mickmister:
@Szymongib Maybe place commas between each label? Just a thought, not sure if it is right to do so

Szymongib:
> Maybe place commas between each label? Just a thought, not sure if it is right to do so

Sure @mickmister, I think that makes sense, especially for mobile version.

Here are some screenshots for comparison:

image

mobile

Szymongib:
@mickmister, @asaadmahmood
Regarding the events for label updated I am not sure how it would suppose to work as I believe the GitHub would emit an event for every label added so if the 3 labels are added to the issue it would result in 3 posted messages. I feel like it could get really spammy quite quickly.
What are your thoughts about it?

asaadmahmood:
@Szymongib If they emit 3 events, then yes, we can post 3 events. But I think when a person is setting multiple labels together, it may just emit a single event.
So just something to look at.

The designs high level look good. On mobile, I think we have more spacing between the labels and the new issue text, but if not possible, its fine as is.

Szymongib:
@asaadmahmood You are right if multiple labels are assigned at once only one event is emitted.

Sorry, I do not have enough knowledge about the plugin, so correct me if I am wrong. As I understand the functionality would post a message to the subscribed channel when the issue/PR is labeled or assigned?

If that is the case I am not totally convinced if that is a good move, as it may result in a lot of messages and I am not sure if those would bring much value. Maybe making it somehow optional would make more sense?

asaadmahmood:
Most definitely, I'm fine if we can make that optional in the subscription.

Szymongib:
I have found that there already is the functionality of posting messages when the issue is labeled. By subscribing to feature label:"[LABEL_NAME]" the message is posted to the channel when the issue/PR is labeled with `[LABEL_NAME]`.
I feel like the new functionality would kind of conflict/duplicate already existing feature.

I have reached out to @mickmister to align how to proceed with this PR.

server/plugin/template.go
```diff
@@ -149,10 +169,24 @@ func init() {
...
 
+	template.Must(masterTemplate.New("labels").Funcs(funcMap).Parse(`
+{{- if .Labels }}
+Labels: {{range $i, $el := .Labels -}}` + "{{- if $i}}, {{end}}[`{{ $el.Name }}`]({{ $.RepositoryURL }}/labels/{{ $el.Name | pathEscape }})" + `{{end -}}
+{{ end -}}
+`))
+
+	template.Must(masterTemplate.New("assignee").Funcs(funcMap).Parse(`
+{{- if .GetAssignee }}
+Assignee: [{{ .GetAssignee.GetLogin }}]({{ .GetAssignee.GetHTMLURL }})
```
mickmister, line 180: It's possible that an issue/PR has multiple assignees. Is that taken into account here?

Szymongib: Fair point. As I understand `GetAssignee` returns only a single one. I have modified it so that all assignees are

Szymongib:
We agreed, that the functionality with labels changes will be implemented as part of a different issue https://github.com/mattermost/mattermost-plugin-github/issues/340

server/plugin/template.go
```diff
@@ -149,10 +169,24 @@ func init() {
...
 
+	template.Must(masterTemplate.New("labels").Funcs(funcMap).Parse(`
+{{- if .Labels }}
+Labels: {{range $i, $el := .Labels -}}` + "{{- if $i}}, {{end}}[`{{ $el.Name }}`]({{ $.RepositoryURL }}/labels/{{ $el.Name | pathEscape }})" + `{{end -}}
+{{ end -}}
+`))
+
+	template.Must(masterTemplate.New("assignee").Funcs(funcMap).Parse(`
+{{- if .Assignees }}
+Assignees: {{range $i, $el := .Assignees -}} {{- if $i}}, {{end}}[{{ $el.GetLogin }}]({{ $el.GetHTMLURL }}){{end -}}
```
hanzei, line 180: If the user has a connected GitHub account, should this link to the Mattermost account instead?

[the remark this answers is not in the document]
Szymongib: Sure, if the index is 0 - meaning single assignee, this evaluates to false and the coma is not added in front of the assignee.

Szymongib: Right, I see that you are using such a convention in other places where the user is displayed. I think it makes sense to change it so it is consistent unless anyone has some objections?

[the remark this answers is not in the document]
Szymongib: So this prevents adding a coma when there is for example only one assignee.

hanzei: Makes sense :+1:

PR #336, opened 2020-07-27
[GH-283] Fix parsing multi-word labels in subscription command

### LINKED ISSUE #283: Labels with spaces don't work when saving a subscription
When trying to set a subscription dependent on a label that contains a space, creating the subscription will fail. The user will receive a response that only 1 feature can be included.  
The same subscription command works if the GitHub label does not contain white space. 
This is not a regression from the previous release. 

Creating a subscription as follows works as expected 
`/github subscribe DHaussermann/hello-world issues,label:"Help"`
 
Creating a subscription as follows will return an error about the number of features.  
`/github subscribe DHaussermann/hello-world issues,label:"Help Wanted"`

utkuufuk:
#### Summary
This PR will update the command parser such that it will skip whitespaces in two double-quotes. This will fix the related bug by protecting the integrity of multi-word labels such as "Help Wanted".

#### Ticket Link
Fixes #283

### DIFF
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -258,7 +258,7 @@ func (p *Plugin) handleSettings(_ *plugin.Context, _ *model.CommandArgs, paramet
 type CommandHandleFunc func(c *plugin.Context, args *model.CommandArgs, parameters []string, userInfo *GitHubUserInfo) string
 
 func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*model.CommandResponse, *model.AppError) {
-	command, action, parameters := p.parseCommand(args.Command)
+	command, action, parameters := parseCommand(args.Command)
 
 	if command != "/github" {
 		return &model.CommandResponse{}, nil
@@ -316,7 +316,7 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 }
 
 // parseCommand parses the entire command input string and retrieves the command, action and parameters
-func (p *Plugin) parseCommand(input string) (command, action string, parameters []string) {
+func parseCommand(input string) (command, action string, parameters []string) {
 	// do not split if the whitespace is in between 2 quotes, e.g. "Hello World"
 	foundQuote := false
 	split := strings.FieldsFunc(input, func(c rune) bool {
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"unicode"
 
 	"github.com/google/go-github/v31/github"
 	"github.com/mattermost/mattermost-server/v5/mlog"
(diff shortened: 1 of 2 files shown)


server/plugin/command.go
```diff
@@ -323,3 +314,28 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
 	p.postCommandResponse(args, fmt.Sprintf("Unknown action %v", action))
 	return &model.CommandResponse{}, nil
 }
+
+// parseCommand parses the entire command input string and retrieves the command, action and parameters
+func parseCommand(input string) (command, action string, parameters []string) {
+	// do not split if the whitespace is in between 2 quotes, e.g. "Hello World"
+	foundQuote := false
```
hanzei, line 321: Nit pick: If `foundQuote` is only used inside the anonymous function for `FieldsFunc` it should be scoped to only that function.

server/plugin/command_test.go
```diff
@@ -83,3 +83,72 @@ func TestValidateFeatures(t *testing.T) {
...
+	type output struct {
+		command    string
+		action     string
+		parameters []string
+	}
+
+	tt := []struct {
+		name  string
+		input string
+		want  output
```
hanzei, line 97: Nit pick: I'm not sure if an additional depth makes the code more readable. Should this be `expectedCommand`, `expectedAction`, `expectedParameters` instead?

server/plugin/command_test.go
```diff
@@ -83,3 +83,72 @@ func TestValidateFeatures(t *testing.T) {
...
+			input: "/github",
+			want: output{
+				"/github",
+				"",
+				[]string(nil),
+			},
+		},
+		{
+			name:  "simple one-word label",
+			input: "/github subscribe DHaussermann/hello-world issues,label:\"Help\"",
```
hanzei, line 119: Would using Raw string literals make the code more readable?

server/plugin/command.go
larkox: Taking a look at the code, it seems that foundQuote needs to be outside, since you want to keep the "state" between different calls to the check function.

server/plugin/command.go
```diff
@@ -323,3 +314,28 @@ func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*mo
...
+	split := strings.FieldsFunc(input, func(c rune) bool {
```
larkox, line 322: This relies on FieldsFunc processing the string in order. Taking a look to the documentation

>  FieldsFunc splits the string s at each run of Unicode code points c satisfying f(c) and returns an array of slices of s. If all code points in s satisfy f(c) or the string is empty, an empty slice is returned. **FieldsFunc makes no guarantees about the order in which it calls f(c)**. If f does not return consistent results for a given c, FieldsFunc may crash.

Sadly, we may have to "reimplement" FieldsFunc to make sure this will work. Thoughts, @hanzei ?

server/plugin/command_test.go
```diff
@@ -83,3 +83,72 @@ func TestValidateFeatures(t *testing.T) {
...
+		},
+		{
+			name:  "multi-word label",
+			input: "/github subscribe DHaussermann/hello-world issues,label:\"Good First Issue\"",
+			want: output{
+				"/github",
+				"subscribe",
+				[]string{"DHaussermann/hello-world", "issues,label:\"Good First Issue\""},
+			},
+		},
```
larkox, line 143: I know they are corner cases, but I would like to see in the tests:
- Weird spacing (words separated by more than one space, both in between quotes and outside the quotes)
- Trailing spaces (spaces before and after the command)
- Line breaks

server/plugin/command_test.go
utkuufuk: @hanzei This makes the comparison easier in the actual test logic. We can just compare the received and expected `output` structs as opposed to comparing each value one by one.

server/plugin/command_test.go
utkuufuk: @larkox I added test cases for multiple spaces inside & outside double quotes, but I'm not sure if it's really necessary to accommodate for line breaks. When the user hits "enter" while typing, the command will be submitted anyway, right?

server/plugin/command.go
utkuufuk: @larkox Good catch, I didn't realize that. It's weird that the tests never failed so far though. I updated the PR with a custom implementation of the parser algorithm.

server/plugin/command_test.go
larkox: I can type shift+enter to add a new line. It is an edge case that will not happen often but I think we should make sure we are handling it.

larkox:
Re-requesting @hanzei 's review due to the rewriting of FieldsFunc.

server/plugin/command_test.go
utkuufuk: @larkox Added the test case: https://github.com/mattermost/mattermost-plugin-github/pull/336/commits/c98c8fdc72cc55d92c6c2c154c5176a56eec1c44 
Apparently my new implementation already handles line breaks :smile:

server/plugin/command_test.go
hanzei: Good point :+1:

server/plugin/command.go
larkox: Probably it hasn't failed because the go implementation does indeed always run in the same order. But since the specification is that way, they can potentially change that in other go implementations / versions. Therefore, better safe than sorry.

Issue #323, opened 2020-07-08
Code Preview functionality may post private repos content


larkox:
Code Preview functionality add a preview of the files from github linked on the post. In order to get the information of the files, it uses the token of the poster. Therefore, any file the poster may link from a private repo, will be exposed by this functionality into the corresponding mattermost channel. This may be considered a breach on security, specially since there is no warning to the user that they are disclosing private information.

@jupenur Any thoughts on this?

jupenur:
Well spotted @larkox. Seems like a minor security concern.

hanzei:
@jupenur @asaadmahmood Do you have thoughts on how do solve this issue? Would a warning message when subscribing to a private repo be a fine solution both security and UX wise?

jupenur:
@hanzei I'd be ok with that. The ideal solutions would be either refactoring the code to use the API token of the person *viewing* the post, or to completely disable previews for private repos, but I doubt either of those would be feasible.

An warning message should be enough as long as it's easy to notice and understand.

hanzei:
After this this a bit more thought I'm not longer sure if a warning is the right solution here. We could warn the user on connecting the account, but that will only happen once and users might forget about it. Warning once the post is created is too late, as at this point the information as already been exposed.


Using the API token of the viewing person is not feasible as these are "plain" posts and using a custom post would drop support for mobile.

Hence I'm leaning towards disabling code preview for private completely. 1/5 Maybe with a setting? 0/5

@larkox @jupenur Let me know what you think.

larkox:
A solution would be to reject the message if the repo is private, and send a confirmation message. Something like:
"You are sending a message with a link to a private repo, which will render the code preview. Are you sure you want to show the code preview? (Show Preview / Don't show preview / Don't send message)"

Nevertheless, I am not sure if will be a good enough UX. I guess that pasting private repos permalinks is not common enough for this being a pain point.

We can also add some settings for different use cases: disable for private repos, enable for private repos, or always ask.
In internal environments we will not be concerned of private code being showed, since it is internal.

jupenur:
> A solution would be to reject the message if the repo is private, and send a confirmation message.

I like this idea, for one, but can't comment on the UX.

hanzei:
While I quite like the proposed solution, it's not an easy one to implement. AFAIK it might even be impossible to also support such dialog on mobile.

[...]

PR #341, opened 2020-07-31
Add warning when subscribing to a private repository

hanzei:
#### Summary
I'm 1/5 on the wording. 

Screenshot from 2020-07-31 10-33-29

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/342


larkox:
Not sure if that would be enough to fix #323 . What about that message warn the user not to copy/paste a permalink of a line of code to any other channel?

jupenur:
The message should show up when connecting your account, right, not when subscribing to a repo?

hanzei:
I completely misunderstood the problem described in https://github.com/mattermost/mattermost-plugin-github/issues/323. My PR actually doesn't addresses the problem at all. I've opened  https://github.com/mattermost/mattermost-plugin-github/issues/342 which describes the issue this PR fixes.

See https://github.com/mattermost/mattermost-plugin-github/issues/323#issuecomment-667086308 for my take on the original issue.

Issue #348, opened 2020-08-10
Use a recovery middleware


hanzei:
Add a recovery middlewar similar to https://github.com/mattermost/mattermost-mattermod/pull/179

PR #349, opened 2020-08-10
Add recovery middleware

hanzei:
#### Summary
Add recovery middleware to avoid crashing the whole plugin if one API call panics.

Credit for the code goes to @agnivade 

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/348


hanzei:
@DHaussermann I don't see a good way to test this PR, as you would have to modify the code to trigger a `panic`. Maybe we skip QA testing?

PR #326, opened 2020-07-12
Removed commented text in Tooltip link body

abdulsmapara:
#### Summary

- Removed commented text present in body of PR/Issue that was displayed in the tooltip (on hovering over PR/issue link) by using `mdCommentRegex`, prepared from the regex for 'comments in markdown' available in `template.go`[1].

[1] [https://github.com/mattermost/mattermost-plugin-github/blob/master/server/plugin/template.go#L13-L28](https://github.com/mattermost/mattermost-plugin-github/blob/master/server/plugin/template.go#L13-L28)

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-plugin-github/issues/312

PR #351, opened 2020-08-12
Avoided panic on empty body of PR and Issue

abdulsmapara:
#### Summary

- Checks whether pointer to the body of PR/Issue is `nil` before accessing it to avoid panic.
- I have assumed that `result` obtained by calling `Get()` on PR/Issue is never `nil` as other functions do not check this condition (Reference: https://github.com/mattermost/mattermost-plugin-github/blob/master/server/plugin/api.go#L489).

#### Ticket Link

Extension to the PR https://github.com/mattermost/mattermost-plugin-github/pull/326 that fixes https://github.com/mattermost/mattermost-plugin-github/issues/312

### DIFF
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -946,7 +946,9 @@ func (p *Plugin) getIssueByNumber(w http.ResponseWriter, r *http.Request, userID
 		p.writeAPIError(w, &APIErrorResponse{Message: "Could not get issue", StatusCode: http.StatusInternalServerError})
 		return
 	}
-	*result.Body = mdCommentRegex.ReplaceAllString(result.GetBody(), "")
+	if result.Body != nil {
+		*result.Body = mdCommentRegex.ReplaceAllString(result.GetBody(), "")
+	}
 	p.writeJSON(w, result)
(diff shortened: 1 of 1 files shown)


hanzei:
@DHaussermann I'm happy to review this one myself, given that I also found the bug.

DHaussermann:
@hanzei or @larkox If you have a bit of time, could one of you peer-test this PR using the  peer testing process documented here https://mattermost.atlassian.net/wiki/spaces/INT/pages/619937892/Peer+Testing+Process

hanzei:
@DHaussermann :poi

DHaussermann:
Thanks for testing @hanzei. Follow-up on this So I can make sure there is appropriate release testing in place. How is it possible to create a PR where `Body` is empty i.e. `nil`? 
I have tested with and an empty description field but this is different?

hanzei:
@DHaussermann E.g. https://github.com/mattermost/mattermost-plugin-demo/pull/112 is a PR without a description, which would cause a panic. I haven't found an issue to reproduce the issue for far.

PR #353, opened 2020-08-19
Migrate logging to use RPC Logging methods

hanzei:
#### Summary
Migrate existing `fmt.Println` and `mlog.X` calls into RPC logging calls. I added come context to the log call when it was easy to do and appropriated. 

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/203

### DIFF
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -13,7 +13,6 @@ import (
 
 	"github.com/google/go-github/v31/github"
 	"github.com/gorilla/mux"
-	"github.com/mattermost/mattermost-server/v5/mlog"
 	"github.com/mattermost/mattermost-server/v5/model"
 	"github.com/mattermost/mattermost-server/v5/plugin"
 	"github.com/pkg/errors"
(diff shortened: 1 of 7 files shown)


larkox (approved):
LGTM. The only consideration I would make is that everything has been set as a LogWarn, and I wonder whether some of those should fall more into Error or more into Info. I haven't look in depth enough to make a distinction, but wanted to raiser the concern.

DHaussermann:
@hanzei or @larkox can you please provide some guidance on what functional testing is needed for this PR. I see there is a QA label.

larkox:
As far as I know, this should only change how errors are logged into the logs, and it is mostly refactoring.
The only functional testing I would consider is forcing errors and make sure they are logged and nothing breaks, but since there are too many error paths and most of them would be really complicated to reach without heavy changes in the deployment or the code (e.g. force a failure when writing to the KVStore), I think this PR can be merged without testing, or just with some regression testing.

larkox:
@DHaussermann :point_up:

PR #360, opened 2020-09-11
Update dependencies

hanzei:
#### Summary
Update dependencies for `v2.0.0` release.

#### Ticket Link
n/a


webapp/package.json
```diff
@@ -13,6 +13,7 @@
     "core-js": "3.1.4",
     "debounce-promise": "3.1.2",
     "mattermost-redux": "github:mattermost/mattermost-redux#b56e7e3a6f7f5de281fd3541eefec9345a8bbc89",
+    "node-fetch": ">=2.6.1",
```
larkox, line 16: What is the difference between using the `^` or `>=`? Also, I thought we preferred using fixed version numbers.

larkox (approved):
LGTM, but I have one question.

webapp/package.json
hanzei: That is a great point :+1:. Yes, we should be using fixed number. I updated `package.json` to only use fixed version and also updated all dependencies (besides `@primer/octicons-react`) while touching the file.

larkox (approved):

jfrerich (approved):
LGTM

hanzei:
Let's test this in the release process

PR #365, opened 2020-09-17
Post command response in RHS if triggered from there

hanzei:
#### Summary
Screenshot from 2020-09-17 10-09-04

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/364


larkox (approved):

jfrerich (approved):

DHaussermann (approved):
Tested and passed
- The command is now aware that is is posted in RHS and becomes part of that thread
- Tested with various commands
- There is no visual feedback in the main thread that of a specific discussion but, I can't find any negative impact to this as they are ephemeral posts

LGTM!

PR #368, opened 2020-09-18
Only render tooltip on hover

hanzei:
#### Summary
Make use of the `show` prop added in https://github.com/mattermost/mattermost-webapp/pull/6492 by only rendering the tooltip on hover. This means data is not longer pre-fetched. Therefore a small delay is visible when hovering. Please note, on re-hovering the link the tooltip is shown imminently. Switching channel caused the data to be lost and it's required to fetch is again. I opted for this approach because storing the data in redux would mean that we would have to clean it up from time to time and it's hard to figure out what data should be cleaned up.

The change is backward compatible to 

I did also simplify the existing code a bit.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-28591

### DIFF
--- a/webapp/src/components/link_tooltip/link_tooltip.jsx
+++ b/webapp/src/components/link_tooltip/link_tooltip.jsx
@@ -7,10 +7,10 @@ import ReactMarkdown from 'react-markdown';
 import Client from 'client';
 import {getLabelFontColor, hexToRGB} from '../../utils/styles';
 
-export const LinkTooltip = ({href, connected, theme}) => {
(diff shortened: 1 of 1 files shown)


mickmister (approved):
LGTM! Thanks for the backwards compatible check

larkox (approved):

DHaussermann (approved):
Tested and passed.
- Works as expected when corresponding webapp PR is in place
- No regressions found on older server versions 
See https://github.com/mattermost/mattermost-webapp/pull/6492 fore details. 
LGTM!

Issue #375, opened 2020-09-25
Assignments tab shows PR icon for issues


hanzei:
![Screenshot from 2020-09-25 20-29-24](https://user-images.githubusercontent.com/16541325/94303276-d21f3480-ff6d-11ea-889c-87afb2fed808.png)

The first item is an issue, but has PR icon

PR #376, opened 2020-09-25
Use issue icon in RHS

hanzei:
#### Summary

| Before  | After           |
|---------|-----------------|
| Screenshot from 2020-09-25 20-29-24 | Screenshot from 2020-09-25 20-26-52 |
| Screenshot from 2020-09-25 20-29-49| Screenshot from 2020-09-25 20-27-08 |

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/375

### DIFF
--- a/webapp/src/components/sidebar_right/github_items.jsx
+++ b/webapp/src/components/sidebar_right/github_items.jsx
@@ -45,12 +45,14 @@ function GithubItems(props) {
             }
             number = (
                 <strong>
-                    <Octicon
-                        icon={iconType}
-                        size='small'
-                        verticalAlign='text-bottom'
-                    />
-                    {' #' + item.number}
+                    <div style={{...style.icon}}>
+                        <Octicon
+                            icon={iconType}
+                            size='small'
+                            verticalAlign='text-bottom'
+                        />
+                    </div>
+                    {'#' + item.number}
                 </strong>);
         }
 
@@ -86,11 +88,11 @@ function GithubItems(props) {
                         style={
                             {
                                 ...style.milestoneIcon,
+                                ...style.icon,
                                 ...((item.created_at || userName) && {paddingLeft: 10}),
                             }
                         }
                     ><SignIcon/></div>
-                    {' '}
                     {item.milestone.title}
                 </span>);
         }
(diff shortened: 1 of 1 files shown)


asaadmahmood (approved):

jfrerich (approved):

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -34,9 +35,24 @@ function GithubItems(props) {
...
 
         if (item.number) {
+            let iconType;
+
+            // item is a pull request
+            if (item.pull_request) {
+                iconType = GitPullRequest;
+            } else {
+                iconType = IssueOpened;
+            }
```
larkox, line 45: Nit:
```suggestion
            let iconType = IssueOpened;

            // item is a pull request
            if (item.pull_request) {
                iconType = GitPullRequest;
            }
```

webapp/src/components/sidebar_right/github_items.jsx
```diff
@@ -34,9 +35,24 @@ function GithubItems(props) {
...
             number = (
                 <strong>
-                    <i className='fa fa-code-fork'/>{' #' + item.number}
+                    <div style={{...style.icon}}>
```
larkox, line 48: Nit: wouldn't it make more sense here a `<span>` instead of a `<div>`?

larkox (approved):
LGTM, just two nits.

webapp/src/components/sidebar_right/github_items.jsx
hanzei: Good suggestion :+1:

webapp/src/components/sidebar_right/github_items.jsx
hanzei: What I like about the current style is that it makes the code more readable IMO. It's clear that  it's an either/or and the reader doesn't have to check if `iconType` is defined.

But that is very much a subjective opinion.

DHaussermann (approved):
Tested and passed
- Confirmed the icons have been replaced as expected
- Link to Issue and PR are functional to navigate to their respective issues
- Tested with multiple themes to ensure no updates needed here
LGTM!

Thanks @hanzei  this is a nice little piece of polish!

PR #402, opened 2020-10-28
[FIX] Handle Issue Creation failures in the correct place

GodlikePenguin:
#### Summary
Originally the handler for 410 Gone (Issues not enabled on a repository) was placed after the generic error handling code, meaning it would never be called.


hanzei (approved):
Good catch :+1:

hanzei:
@DHaussermann I don't think there is a good way to test this. The interaction relies to the behaviour on GitHub, which we can't mock. Also given that this is a small change, I'm going to merge it without QA review.

PR #417, opened 2021-01-12
Update plugin.json

justinegeffen:
Updated punctuation for UI consistency.

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -85,7 +85,7 @@
                         "value": "privateAndPublic"
                     },
                     {
-                        "display_name": "Disable:",
+                        "display_name": "Disable",
                         "value": "disable"
                     }
                 ]
--- a/plugin.json
+++ b/plugin.json
@@ -81,7 +81,7 @@
                         "value": "public"
                     },
                     {
-                        "display_name": "Enable for public and private repositories. This might leak confidential code into public channels:",
+                        "display_name": "Enable for public and private repositories. This might leak confidential code into public channels",
                         "value": "privateAndPublic"
                     },
                     {
--- a/plugin.json
+++ b/plugin.json
@@ -19,73 +19,73 @@
         "bundle_path": "webapp/dist/main.js"
     },
     "settings_schema": {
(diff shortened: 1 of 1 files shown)


plugin.json
```diff
@@ -19,73 +19,73 @@
...
                 "key": "EnableCodePreview",
-                "display_name": "Enable Code Previews",
+                "display_name": "Enable Code Previews:",
                 "type": "dropdown",
                 "help_text": "Allow the plugin to expand permalinks to GitHub files with an actual preview of the linked file.",
                 "default": "public",
                 "options": [
                     {
-                        "display_name": "Enable for public repositories",
+                        "display_name": "Enable for public repositories:",
```
hanzei, line 80: That seems odd. Usually drop down options don't end with a punctuation. Can you please double check if that is intended?

hanzei (changes requested):

plugin.json
justinegeffen: No, you're correct, that was my bad.

justinegeffen, line 80:
```suggestion
                        "display_name": "Enable for public repositories",
```

plugin.json
```diff
@@ -19,73 +19,73 @@
...
                         "value": "public"
                     },
                     {
-                        "display_name": "Enable for public and private repositories. This might leak confidential code into public channels.",
+                        "display_name": "Enable for public and private repositories. This might leak confidential code into public channels:",
```
justinegeffen, line 84:
```suggestion
                        "display_name": "Enable for public and private repositories. This might leak confidential code into public channels",
```

plugin.json
```diff
@@ -19,73 +19,73 @@
...
                         "value": "privateAndPublic"
                     },
                     {
-                        "display_name": "Disable",
+                        "display_name": "Disable:",
```
justinegeffen, line 88:
```suggestion
                        "display_name": "Disable",
```

hanzei (approved):

PR #442, opened 2021-05-07
MM-35518 - Add CodeQL check

srkgupta:
#### Summary
Added step to run automated CodeQL checks. Once this file is merged, the codeql checks (code scanning checks) will automatically be performed on following conditions:
- new pull requests to master branch
- scheduled to run every week once (it automatically pulls the latest codeql checks added, if any)

#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-35518

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,42 @@
+name: "CodeQL"
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    # The branches below must be a subset of the branches above
+    branches: [ master ]
+  schedule:
+    - cron: '45 0 * * 0'
+
+jobs:
+  analyze:
+    name: Analyze
+    runs-on: ubuntu-latest
+    permissions:
+      actions: read
+      contents: read
+      security-events: write
+
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


hanzei (commented):
Do we really need to add such file into every repo? At least for the plugins that 30+ repos

srkgupta:
> Do we really need to add such file into every repo? At least for the plugins that 30+ repos

Yes, this is a one-time task. The code scanning results will be isolated and tracked per plugin repo.

hanzei:
> > Do we really need to add such file into every repo? At least for the plugins that 30+ repos
> 
> Yes, this is a one-time task. The code scanning results will be isolated and tracked per plugin repo.

What makes your confident that the pipeline will never change?

srkgupta:
> What makes your confident that the pipeline will never change?

It uses the pipeline provided by Github and I don't expect any breaking changes here. There are several repositories which is already utilising this pipeline. Even if there are some changes in the future, Github will most probably ensure that its automatically handled. Worst case, if at all there are breaking changes which requires us to update the codeql-analysis file individually, we might think of automating and push it across all the repos where codeql is enabled.

hanzei (approved):
LGTM :+1:

Issue #378, opened 2020-10-02
GitHub - Check for an existing webhook when subscribing to a repo


aaronrothschild:
When a user create a subscription to a repo the plugin should check the repo (or parent Organization for Enterprise) to see if a webhook is in place to deliver the events 

If no webhook is found on the GitHub side, an ephemeral post should occur to remind the user that a webhook must be set up. It can then post the url that must be used to deliver the events. 

Note that this functionality is already implemented for GitLab and some of this implementation may be salvageable.     cc @DHaussermann

hanzei:
One thing to consider while implementing this is which token to use for fetching the information about what webhook to use.

hanzei:
Requires https://github.com/mattermost/mattermost-plugin-github/pull/456 to be merged first.

sibasankar-demansol:
@hanzei  just wanted to check 

will it be okey if we return the message when the subscription is successful and there is no webhook for repo/owner

Successfully subscribed to mattermost/mattermost-plugin-github , can configure the webhook to receive notification here(link to plugin colnsole)

sibasankar-demansol:
@hanzei  have implemented this feature in this way , let me know if you have any suggestion from your side

![Screenshot 2021-08-02 at 11 39 26 AM](https://user-images.githubusercontent.com/86650698/127812184-9e4f384f-e3d6-4a46-8f76-f7bf9954bb68.png)

hanzei:
https://github.com/mattermost/mattermost-plugin-github/pull/466 can be used as a starting point. This ticket is only about warning the admin that the webhook is missing. `/github setup webhook` is the command that the admin should be pointed towards.

Issue #379, opened 2020-10-02
Create a webhook via a slash command in mattermost 


aaronrothschild:
Add a feature that will allow users to add and list webhooks in GitHub via a Mattermost slash command. The user would pass in a GitHub repository and a slash command would create a webhook on behalf of the connected Mattermost user. 

A list option would also show the existing webhooks for the repo name passed in by the user.   

Note that this functionality is already implemented for GitLab and some of this implementation may be salvageable.

sasha27:
@aaronrothschild is someone already working on this issue? If not then I would like to work on it. Can you please point me out to respective packages I can look into and also documentations on how I can set up a testing environment, please?

larkox:
Sure, you can take it, @sasha27 ! The implementation in Gitlab you can find it here:
https://github.com/mattermost/mattermost-plugin-gitlab/blob/dc2be260042454cd3db7cb9cc7cf35594e34eb80/server/command.go#L244
Regarding setting up the environment, please check here:
https://developers.mattermost.com/contribute/getting-started/

hanzei:
This is implemented via the `/github setup webhook` command

PR #466, opened 2021-08-02
Mm 378 - GitHub - Check for an existing webhook when subscribing to a repo

sibasankar-demansol:
when the [owner / repo] is subscribed , check if have setup any webhook if not let user know they need to do it

ticket here
https://github.com/mattermost/mattermost-plugin-github/issues/378

### DIFF
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -506,9 +506,7 @@ func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, param
 		}
 		hook.Config = config
 
-		githubHook, githubHookResponse, err := p.CreateHook(ctx, githubClient, owner, repo, hook)
-		p.API.LogWarn("githubHook logs ", "githubHook", githubHook, "githubHookResponse ", githubHookResponse)
-		p.API.LogWarn("githubHook logs ", "err", err)
+		githubHook, _, err := p.CreateHook(ctx, githubClient, owner, repo, hook)
 		if err != nil {
 			return err.Error()
 		}
@@ -545,13 +543,9 @@ func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, param
 			for ShouldCallNext {
 				Page++
 				githubHooks, githubResponse, err = githubClient.Organizations.ListHooks(ctx, owner, &github.ListOptions{PerPage: Page})
-				p.API.LogWarn("creating hook organ", "githubHooks", githubHooks, "githubResponse", githubResponse)
 				if err != nil {
 					return err.Error()
 				}
-				// if len(githubHooks) > 0 {
-				// 	hookList["-"] = githubHooks
-				// }
 				if githubResponse.NextPage == 0 {
 					ShouldCallNext = false
 				}
@@ -560,56 +554,16 @@ func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, param
 			for ShouldCallNext {
 				Page++
 				githubHooks, githubResponse, err = githubClient.Repositories.ListHooks(ctx, owner, repo, &github.ListOptions{PerPage: Page})
-				p.API.LogWarn("creating hook repo", "githubHooks", githubHooks, "githubResponse", githubResponse)
 
 				if err != nil {
 					return err.Error()
 				}
-				// if len(githubHooks) > 0 {
-				// 	hookList[repo] = githubHooks
-				// }
 				if githubResponse.NextPage == 0 {
 					ShouldCallNext = false
 				}
 			}
 		}
-		// var repos []string
-		// repos = append(repos, repo)
-		// if repo == "" {
-		// 	for ShouldCallNext {
-		// 		Page++
-		// 		projects, projectResponse, err := githubClient.Repositories.List(ctx, owner, &github.RepositoryListOptions{ListOptions: github.ListOptions{Page: Page}})
-		// 		if err != nil {
-		// 			return err.Error()
-		// 		}
-		// 		if projectResponse.NextPage == 0 {
-		// 			ShouldCallNext = false
-		// 		}
-		// 		for _, project := range projects {
-		// 			repos = append(repos, *project.Name)
-		// 		}
-		// 	}
-		// }
-		// var hookList = make(map[string][]*github.Hook)
-		// for i := 0; i < len(repos); i++ {
-		// 	ShouldCallNext = true
-		// 	Page = 0
-		// 	if repos[i] != "" {
-		// 		for ShouldCallNext {
-		// 			Page++
-		// 			githubHooks, githubResponse, err := githubClient.Repositories.ListHooks(ctx, owner, repos[i], &github.ListOptions{PerPage: Page})
-		// 			if err != nil {
-		// 				return err.Error()
-		// 			}
-		// 			if len(githubHooks) > 0 {
-		// 				hookList[repos[i]] = githubHooks
-		// 			}
-		// 			if githubResponse.NextPage == 0 {
-		// 				ShouldCallNext = false
-		// 			}
-		// 		}
-		// 	}
-		// }
+
 		var txt string
 		if len(githubHooks) == 0 {
 			txt = "Currently there are no webhook in this [Owner/Repo]"
@@ -617,12 +571,6 @@ func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, param
 			txt = "### Webhook in this Repositories\n"
 		}
 
-		// for repoName, repoHooks := range hookList {
-		// 	for _, val := range repoHooks {
-		// 		hookDetails, _, err := githubClient.Repositories.GetHook(ctx, owner, repoName, *val.ID)
-		// 		if err != nil {
-		// 			return err.Error()
-		// 		}
 		for _, hook := range githubHooks {
 			hookID := strconv.Itoa(int(*hook.ID))
 			hookURL := baseURL + owner
@@ -633,9 +581,6 @@ func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, param
 			txt += fmt.Sprintf(" *  [%s](%s) :  -  %s", hookID, hookURL+hookID, strings.Join(hook.Events, " , "))
 			txt += newLine
 		}
-
-		// 	}
-		// }
 
 		return txt
 	}
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -313,14 +313,21 @@ func (p *Plugin) handleSubscribesAdd(_ *plugin.Context, args *model.CommandArgs,
 
 	ctx := context.Background()
 	githubClient := p.getGithubClient(userInfo)
-
+	var page int
 	owner, repo := parseOwnerAndRepo(parameters[0], p.getBaseURL())
+
+	webhookURL := *p.API.GetConfig().ServiceSettings.SiteURL + "/admin_console/plugins/plugin_github"
+
 	if repo == "" {
 		if err := p.SubscribeOrg(ctx, githubClient, args.UserId, owner, args.ChannelId, features, flags); err != nil {
 			return err.Error()
 		}
-
-		return fmt.Sprintf("Successfully subscribed to organization %s.", owner)
+		msg := fmt.Sprintf("Successfully subscribed to organization %s.", owner)
+		githubHooks, _, _ := githubClient.Repositories.ListHooks(ctx, owner, repo, &github.ListOptions{PerPage: page})
+		if len(githubHooks) == 0 {
+			msg += fmt.Sprintf("can configure the webhook to receive notification [here](%s).", webhookURL)
+		}
+		return msg
 	}
 
 	if err := p.Subscribe(ctx, githubClient, args.UserId, owner, repo, args.ChannelId, features, flags); err != nil {
@@ -335,7 +342,11 @@ func (p *Plugin) handleSubscribesAdd(_ *plugin.Context, args *model.CommandArgs,
 	} else if ghRepo != nil && ghRepo.GetPrivate() {
 		msg += "\n\n**Warning:** You subscribed to a private repository. Anyone with access to this channel will be able to read the events getting posted here."
 	}
+	githubHooks, _, _ := githubClient.Repositories.ListHooks(ctx, owner, repo, &github.ListOptions{PerPage: page})
+	if len(githubHooks) == 0 {
+		msg += fmt.Sprintf("can configure the webhook to receive notification [here](%s).", webhookURL)
 
+	}
 	return msg
 }
 
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -38,7 +38,9 @@ var validFeatures = map[string]bool{
 
 const (
 	list      = "list"
+	add       = "add"
 	deleteAll = "delete-all"
+	newLine   = "\n"
 )
 const (
 	inboundWebhookURL = "/plugins/github/webhook"
@@ -188,7 +190,7 @@ func (p *Plugin) handleMuteCommand(_ *plugin.Context, args *model.CommandArgs, p
 	switch {
 	case command == list:
 		return p.handleMuteList(args, userInfo)
-	case command == "add":
+	case command == add:
 		if len(parameters) != 2 {
 			return "Invalid number of parameters supplied to " + command
 		}
@@ -224,7 +226,7 @@ func (p *Plugin) handleSubscribe(c *plugin.Context, args *model.CommandArgs, par
 	switch {
 	case len(parameters) == 0:
 		return "Please specify a repository or 'list' command."
-	case len(parameters) == 1 && parameters[0] == "list":
+	case len(parameters) == 1 && parameters[0] == list:
 		return p.handleSubscriptionsList(c, args, parameters[1:], userInfo)
 	default:
 		return p.handleSubscribesAdd(c, args, parameters, userInfo)
@@ -240,9 +242,9 @@ func (p *Plugin) handleSubscriptions(c *plugin.Context, args *model.CommandArgs,
 	parameters = parameters[1:]
 
 	switch {
-	case command == "list":
+	case command == list:
 		return p.handleSubscriptionsList(c, args, parameters, userInfo)
-	case command == "add":
+	case command == add:
 		return p.handleSubscribesAdd(c, args, parameters, userInfo)
 	case command == "delete":
 		return p.handleUnsubscribe(c, args, parameters, userInfo)
@@ -269,7 +271,7 @@ func (p *Plugin) handleSubscriptionsList(_ *plugin.Context, args *model.CommandA
 		if subFlags != "" {
 			txt += fmt.Sprintf(" %s", subFlags)
 		}
-		txt += "\n"
+		txt += newLine
 	}
 
 	return txt
@@ -465,9 +467,9 @@ func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, param
 	parameters = parameters[1:]
 	ctx := context.Background()
 	githubClient := p.getGithubClient(userInfo)
-	baseUrl := p.getBaseURL()
+	baseURL := p.getBaseURL()
 	switch command {
-	case "add":
+	case add:
 		if len(parameters) < 1 {
 			return "Unknown action, please use `/github help` to see all actions available."
 		}
@@ -484,15 +486,16 @@ func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, param
 		if len(parameters) > 2 {
 			for _, val := range strings.Split(parameters[2], ",") {
 				KVPair := strings.Split(val, "=")
-				if KVPair[0] == "insecure_ssl" {
+				switch {
+				case KVPair[0] == "insecure_ssl":
 					insecureSSL, err := strconv.ParseBool(KVPair[1])
 					if err != nil {
 						return "config filed insecure_ssl can only have on of the following true or false "
 					}
 					config[KVPair[0]] = insecureSSL
-				} else if KVPair[0] == "options" {
+				case KVPair[0] == "options":
 					hook.Events = strings.Split(KVPair[1], "&")
-				} else {
+				default:
 					config[KVPair[0]] = KVPair[1]
 				}
 			}
@@ -501,7 +504,6 @@ func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, param
 			config["insecure_ssl"] = false
 			config["content_type"] = "application/json"
 			hook.Events = []string{"*"}
-
 		}
 		hook.Config = config
 		githubHook, _, err := githubClient.Repositories.CreateHook(ctx, owner, repo, &hook)
@@ -512,14 +514,14 @@ func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, param
 		if err != nil {
 			return err.Error()
 		}
-		baseUrl1 := baseUrl + owner + "/" + repo + "/settings/hooks/"
+		hookURL := baseURL + owner + "/" + repo + "/settings/hooks/"
 		txt := "Webhook Created Successfully \n"
 		hookID := strconv.Itoa(int(*githubHook.ID))
-		txt += fmt.Sprintf(" *  [%s](%s) :  -  %s", hookID, baseUrl1+hookID, strings.Join(hookDetails.Events, " , "))
-		txt += "\n"
+		txt += fmt.Sprintf(" *  [%s](%s) :  -  %s", hookID, hookURL+hookID, strings.Join(hookDetails.Events, " , "))
+		txt += newLine
 		return txt
 
-	case "list":
+	case list:
 
 		if len(parameters) != 1 {
 			return "Unknown action, please use `/github help` to see all actions available."
@@ -544,7 +546,6 @@ func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, param
 					repos = append(repos, *project.Name)
 				}
 			}
-
 		}
 		var hookList = make(map[string][]*github.Hook)
 		for i := 0; i < len(repos); i++ {
@@ -566,7 +567,7 @@ func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, param
 				}
 			}
 		}
-		var txt = ""
+		var txt string
 		if len(hookList) == 0 {
 			txt = "Currently there are no webhook in this repository"
 		} else {
@@ -580,9 +581,9 @@ func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, param
 					return err.Error()
 				}
 				hookID := strconv.Itoa(int(*val.ID))
-				baseUrl1 := baseUrl + owner + "/" + repoName + "/settings/hooks/"
-				txt += fmt.Sprintf(" *  [%s](%s) :  -  %s", hookID, baseUrl1+hookID, strings.Join(hookDetails.Events, " , "))
-				txt += "\n"
+				hookURL := baseURL + owner + "/" + repoName + "/settings/hooks/"
+				txt += fmt.Sprintf(" *  [%s](%s) :  -  %s", hookID, hookURL+hookID, strings.Join(hookDetails.Events, " , "))
+				txt += newLine
 			}
 		}
 
@@ -683,10 +684,10 @@ func getAutocompleteData(config *Configuration) *model.AutocompleteData {
 
 	subscriptions := model.NewAutocompleteData("subscriptions", "[command]", "Available commands: list, add, delete")
 
-	subscribeList := model.NewAutocompleteData("list", "", "List the current channel subscriptions")
+	subscribeList := model.NewAutocompleteData(list, "", "List the current channel subscriptions")
 	subscriptions.AddCommand(subscribeList)
 
-	subscriptionsAdd := model.NewAutocompleteData("add", "[owner/repo] [features] [flags]", "Subscribe the current channel to receive notifications about opened pull requests and issues for an organization or repository. [features] and [flags] are optional arguments")
+	subscriptionsAdd := model.NewAutocompleteData(add, "[owner/repo] [features] [flags]", "Subscribe the current channel to receive notifications about opened pull requests and issues for an organization or repository. [features] and [flags] are optional arguments")
 	subscriptionsAdd.AddTextArgument("Owner/repo to subscribe to", "[owner/repo]", "")
 	subscriptionsAdd.AddTextArgument("Comma-delimited list of one or more of: issues, pulls, pushes, creates, deletes, issue_creations, issue_comments, pull_reviews, label:\"<labelname>\". Defaults to pulls,issues,creates,deletes", "[features] (optional)", `/[^,-\s]+(,[^,-\s]+)*/`)
 	if config.GitHubOrg != "" {
@@ -710,7 +711,7 @@ func getAutocompleteData(config *Configuration) *model.AutocompleteData {
 
 	mute := model.NewAutocompleteData("mute", "[command]", "Available commands: list, add, delete, delete-all")
 
-	muteAdd := model.NewAutocompleteData("add", "[github username]", "Mute notifications from the provided GitHub user")
+	muteAdd := model.NewAutocompleteData(add, "[github username]", "Mute notifications from the provided GitHub user")
 	muteAdd.AddTextArgument("GitHub user to mute", "[username]", "")
 	mute.AddCommand(muteAdd)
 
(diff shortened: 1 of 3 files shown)


mickmister:
@hanzei There's an issue here @sibasankar-demansol ran into where the OAuth token does not have permissions to create webhooks. We currently request the scopes:

- repo
- notifications
- read:org

Looking at the [available scopes](https://github.com/google/go-github/blob/0d84fe1b2f64d2813bb97e0fd5f1df4963a2c3a7/github/authorizations.go#L20-L44), we can add these for admin users:

- read:repo_hook
- write:repo_hook
- admin:repo_hook
- admin:org_hook

We can then have a `/github connect admin` command to apply these scopes. I also think we should be storing the scopes in the `GitHubUserInfo` struct, so we always know what permissions a user has inside of GitHub.

@hanzei Let me know your thoughts on this, and I'll make a separate ticket for the task.

server/plugin/command.go
```diff
@@ -453,6 +469,151 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+		var config = make(map[string]interface{})
+
+		if len(parameters) > 2 {
+			for _, val := range strings.Split(parameters[2], ",") {
+				KVPair := strings.Split(val, "=")
+				switch {
+				case KVPair[0] == "insecure_ssl":
+					insecureSSL, err := strconv.ParseBool(KVPair[1])
+					if err != nil {
+						return "config filed insecure_ssl can only have on of the following true or false "
```
sanjaydemansol, line 502:
```suggestion
						return "Config failed, `insecure_ssl` must have a boolean(true|false) value."
```

server/plugin/command.go
```diff
@@ -453,6 +469,151 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+				hookURL += "/" + repo
+			}
+			hookURL += "/settings/hooks/"
+			txt += fmt.Sprintf(" *  [%s](%s) :  -  %s", hookID, hookURL+hookID, strings.Join(hook.Events, " , "))
+			txt += newLine
+		}
+
+		return txt
+	}
+	return "invalid action only add and list command are supported"
```
sanjaydemansol, line 597:
```suggestion
	return "Invalid action, only `add` and `list` commands are supported."
```

server/plugin/template.go
```diff
@@ -350,7 +350,57 @@ Assignees: {{range $i, $el := .Assignees -}} {{- if $i}}, {{end}}{{template "use
 		"  * `/github mute list` - list your muted GitHub users\n" +
 		"  * `/github mute add [username]` - add a GitHub user to your muted list\n" +
 		"  * `/github mute delete [username]` - remove a GitHub user from your muted list\n" +
-		"  * `/github mute delete-all` - unmute all GitHub users\n"))
+		"  * `/github mute delete-all` - unmute all GitHub users\n" +
+		"* `/github Webhook` - Managed to List Webhook of [owner/repo] and add webhook to owoner[/repo].\n" +
```
sanjaydemansol, line 354:
```suggestion
		"  * `/github Webhook` - Manage all webhooks.\n" +
```

server/plugin/command.go
```diff
@@ -453,6 +469,151 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+					ShouldCallNext = false
+				}
+			}
+		}
+
+		var txt string
+		if len(githubHooks) == 0 {
+			txt = "Currently there are no webhook in this [Owner/Repo]"
+		} else {
+			txt = "### Webhook in this Repositories\n"
```
sanjaydemansol, line 581: could you add [Owner/Repo] details to the heading?

server/plugin/command.go
```diff
@@ -307,14 +313,21 @@ func (p *Plugin) handleSubscribesAdd(_ *plugin.Context, args *model.CommandArgs,
...
 	if repo == "" {
 		if err := p.SubscribeOrg(ctx, githubClient, args.UserId, owner, args.ChannelId, features, flags); err != nil {
 			return err.Error()
 		}
-
-		return fmt.Sprintf("Successfully subscribed to organization %s.", owner)
+		msg := fmt.Sprintf("Successfully subscribed to organization %s.", owner)
+		githubHooks, _, _ := githubClient.Repositories.ListHooks(ctx, owner, repo, &github.ListOptions{PerPage: page})
+		if len(githubHooks) == 0 {
+			msg += fmt.Sprintf("can configure the webhook to receive notification [here](%s).", webhookURL)
```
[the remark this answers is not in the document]
sanjaydemansol: Also, we can show something like **_contact admin_** if user is not Matermost Admin.

sanjaydemansol (commented):
Suggestions

server/plugin/command.go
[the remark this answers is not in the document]
maisnamraju: I would rephrase it to "Configure webhook to receive notification [here] (%s)"

server/plugin/command.go
```diff
@@ -453,6 +469,151 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+	}
+	command := parameters[0]
+	parameters = parameters[1:]
+	ctx := context.Background()
+	githubClient := p.getGithubClient(userInfo)
+	baseURL := p.getBaseURL()
+	switch command {
+	case add:
+		if len(parameters) < 1 {
+			return "Unknown action, Currently supported to add and List webhook "
```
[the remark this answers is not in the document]
maisnamraju: lgtm

server/plugin/command.go
```diff
@@ -453,6 +469,151 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+		}
+		owner, repo := parseOwnerAndRepo(parameters[0], p.getBaseURL())
+		if owner == "" {
+			return "Currently supported to add webhook to a specific [owner/repo] ."
```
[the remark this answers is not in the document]
maisnamraju: LGTM

server/plugin/command.go
maisnamraju: LGTM

server/plugin/template.go
maisnamraju: Just need to get rig of the white space before the `*` at 354

server/plugin/command.go
```diff
@@ -615,6 +776,30 @@ func getAutocompleteData(config *Configuration) *model.AutocompleteData {
...
+
+	webhook.AddCommand(webhookList)
+	webhookAdd := model.NewAutocompleteData(add, "[owner/repo] callback_url token content_type events insecure_ssl", "Add a webhook to desired owner[/repo] with optional --config")
+	webhookAdd.AddTextArgument("Owner/repo to create a webhook", "owner[/repo]", "")
+	webhookAdd.AddTextArgument("Call Back URL", "", "")
+	webhookAdd.AddTextArgument("token", "", "")
+	webhookAdd.AddTextArgument("content_type", "", "")
+	webhookAdd.AddTextArgument("Comma-delimited list of one or more of: create, delete, check_run, check_suite, code_scanning_alert, member, commit_comment, deploy_key, deployment_status, deployment, discussion_comment, discussion, fork, issue_comment, issues, label, meta, milestone, package, page_build, project_card, project_column, project, pull_request_review_comment, pull_request_review_thread, pull_request_review, pull_request, push, registry_package, release, repository, repository_import, repository_vulnerability_alert, secret_scanning_alert, star, status, team_add, public, watch, gollum . Defaults to *", "[features] (optional)", `/[^,-\s]+(,[^,-\s]+)*/`)
```
mickmister, line 789: I think the point of this feature is to make it so the admin doesn't need to think about the webhook. We don't want to simply "move" the burden to our platform. Here's how I think this feature should work:

The admin runs `/github subscriptions add ...` to create a subscription. We check to see if we have a webhook set up for this repo or the surrounding org. If one exists, we're good to go and nothing left to do except create our own internal subscription as usual.

Otherwise, we reply back to the user: "No webhook was found for this repository or organization. Would you like the webhook to be created?", and show a form in the ephemeral response using an [Interactive Message](https://docs.mattermost.com/developer/interactive-messages.html). If the user has privileges to create an org-level webhook, provide that as an option on the form.

When the form is submitted, we create the appropriate webhook and automatically choose to the webhook events that the plugins needs to use. We can include the URL to the webhook that we created in our final output.

We should then process the original command submission of `/github subscriptions add ...`, so that information needs to be stored in the form. You can use the `Context` field of [PostActionIntegration](https://pkg.go.dev/github.com/mattermost/mattermost-server/model#PostActionIntegration
) to store the information about the originally submitted command.

Once that is done, we respond to the user with a link to the GitHub webhook page, and a message saying the subscription was created.

@sibasankar-demansol @hanzei @aaronrothschild Thoughts on this?

mickmister (changes requested):

server/plugin/command.go
hanzei: While an interactive message is a very handy way to create the webhook, I would also be fine simply making use of https://github.com/mattermost/mattermost-plugin-github/pull/456 and returning an ephemeral post that contains the command that needs to be run.

hanzei (changes requested):
Blocking on [this](https://github.com/mattermost/mattermost-plugin-github/pull/466#discussion_r685694061) discussion

server/plugin/command.go
sibasankar-demansol: @hanzei  currently the scope of oAuth is for Repo level , could you let me know should we add that to scope 

rest suggestion i will work on

mickmister: @hanzei I like the slash command idea the ephemeral message idea. 2/5 I think we should show 2 commands, one for making a repo-level webhook, and one for making an org-level webhook.

sibasankar-demansol: @hanzei  , @mickmister  
i have few things from my side
1. currently #456  is yet to be done and is incomplete until we give the oAuth access for Org level or move on with creating webhook on Repo level .
2. Regarding this issue do you want me to show a ephemeral message with 2 commands for repo and org level differently , as the slash command is not yet completed to create webhook only way left is to give a link redirecting to github plugin console

mickmister: @sibasankarnayak 1/5 How about we assume the other PR is complete, and we put the two slash commands here. Then  we can put the "Do Not Merge/Awaiting PR" label on this PR to signal that it's waiting on changes.

sibasankarnayak: @mickmister  Thats sounds good,

just have a thought of show 
when user subscribe to a individual repo can show 2 commands
and when user subscribe to org then we can return one command to create org hook

what do you think does it makes any sense

sibasankarnayak: @mickmister  @hanzei 
can you suggest the changes over visual representation 
 Screenshot 2021-08-19 at 5 57 12 PM

hanzei: The response should make it clear that the command needs to be run. Also there is a  `/` missing at the beginning.

Are both webhooks needed in any case? Is the org specific enough?

sibasankarnayak: have updated the message , can you check now

Screenshot 2021-08-25 at 9 59 58 PM

mickmister: @sibasankarnayak Make sure you either mention or re-request review from the person you're communicating with on in issues and PRs. The other person doesn't get a push notification unless you do one or the other. The one time you don't need to mention someone is when you are commenting on their issue or PR. They will automatically get notifications for the comment.

When you have made changes that you need to be reviewed (which is the case here), your comment here is fine but you should also explicitly re-request review so the reviewer knows they need to review again.

mickmister: Actually since you wouldn't be requesting a full re-review here, you can just mention the reviewer in your comment cc @hanzei

sibasankarnayak: @mickmister  Sure will keep this in mind

server/plugin/command.go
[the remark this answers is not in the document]
sibasankarnayak: > Blocking on [this](https://github.com/mattermost/mattermost-plugin-github/pull/466#discussion_r685694061) discussion

Made the changes required

hanzei (changes requested):
While testing this change I'm running into a similar issue as in https://github.com/mattermost/mattermost-plugin-github/pull/456. 
Screenshot from 2021-09-07 10-00-55

Is there anything specific requirement to test this PR?

plugin.json
```diff
@@ -105,5 +105,50 @@
             }
         ],
         "footer": "* To report an issue, make a suggestion or a contribution, [check the repository](https://github.com/mattermost/mattermost-plugin-github)."
+    },
+    "props":{
```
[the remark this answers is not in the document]
sibasankarnayak: this is a branch created from issue  #379  , which is created for  adding variables to mainfest.go

sibasankarnayak:
> While testing this change I'm running into a similar issue as in #456.
> Screenshot from 2021-09-07 10-00-55
> 
> Is there anything specific requirement to test this PR?

Can you let me know the Command you are using for ?

hanzei:
@sibasankarnayak `/github hooks add matterpoll/matterpoll some_secret`

server/plugin/command.go
```diff
@@ -506,6 +522,171 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+	command := parameters[0]
+	parameters = parameters[1:]
+	ctx := context.Background()
+	githubClient := p.getGithubClient(userInfo)
+	baseURL := p.getBaseURL()
+	switch command {
+	case add:
+		if len(parameters) < 1 {
+			return "Invalid parameters for add command, Provide [owner/repo] details."
+		} else if len(parameters) < 2 {
```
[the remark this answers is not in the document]
mickmister:
```suggestion
		}
		
		if len(parameters) < 2 {
```

plugin.json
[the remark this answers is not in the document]
mickmister: If the change isn't related to this PR, the change should be removed from this PR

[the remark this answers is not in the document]
mickmister: I haven't ever seen this pattern either. This is to share values between the server and webapp portions of the plugin?

server/plugin/command.go
```diff
@@ -345,6 +349,12 @@ func (p *Plugin) handleSubscribesAdd(_ *plugin.Context, args *model.CommandArgs,
 			}
 			subOrgMsg += "\n\n" + fmt.Sprintf("Notifications are disabled for %s", excludeMsg)
 		}
+		githubHooks, _, _ := githubClient.Repositories.ListHooks(ctx, owner, repo, &github.ListOptions{PerPage: page})
+		if len(githubHooks) == 0 {
+			subOrgMsg += "\n No webhook was found for this repository or organization. Would you like the webhook to be created?"
```
mickmister, line 354: I don't think we want the spaces after each `\n`. Can you go through the PR and remove the extra space at the beginning of each new line?

```suggestion
			subOrgMsg += "\nNo webhook was found for this repository or organization. Would you like the webhook to be created?"
```

plugin.json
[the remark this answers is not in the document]
mickmister: To minimize the diff between this PR and the other PR that is in progress, you may be able to have this PR's base branch set to the other PR's branch. Though it may not let you do this because it's on a fork

mickmister (changes requested):

plugin.json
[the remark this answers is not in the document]
sibasankarnayak: yes have tried creating base branch as the other one but it only let us to create PR with MM branch

sibasankarnayak:
> matterpoll



> @sibasankarnayak `/github hooks add matterpoll/matterpoll some_secret`

@hanzei  seems the org/repo doesn't exist , this 404 error occurs only in 2 case 
1.  trying to create out of scope 
2.  passed invalid org/repo 

you need to pass a valid org/repo  i.e. `/github webhook add sibasankarnayak/mattermost-plugin-github some_secret htpps://callback_url`

server/plugin/command.go
sibasankarnayak: Done

server/plugin/command.go
```diff
@@ -345,6 +349,12 @@ func (p *Plugin) handleSubscribesAdd(_ *plugin.Context, args *model.CommandArgs,
 			}
 			subOrgMsg += "\n\n" + fmt.Sprintf("Notifications are disabled for %s", excludeMsg)
 		}
+		githubHooks, _, _ := githubClient.Repositories.ListHooks(ctx, owner, repo, &github.ListOptions{PerPage: page})
```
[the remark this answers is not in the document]
sibasankarnayak: @hanzei i thing there is no need of pagination here , as the main objective is to find do we have any existing web hook , 

even fetching hooks for page 1 can confirm if any hook is already there or not
feel if the page 1 have 0 hooks there won't be any page 2

plugin.json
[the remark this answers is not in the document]
sibasankarnayak: > I haven't ever seen this pattern either. This is to share values between the server and webapp portions of the plugin?

yes in the #379  this was a suggestion to use by @hanzei  [here](https://github.com/mattermost/mattermost-plugin-github/pull/456#discussion_r670259213) 
after figuring out is there any way , i felt props is the only where we can add new values as rest is with static structure

server/plugin/command.go
```diff
@@ -506,6 +522,171 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+	return nil
+}
+func (p *Plugin) GetHook(ctx context.Context, githubClient *github.Client, owner, repo string, hookID int64) (*github.Hook, *github.Response, error) {
+	if repo != "" {
+		return githubClient.Repositories.GetHook(ctx, owner, repo, hookID)
+	}
+	return githubClient.Organizations.GetHook(ctx, owner, hookID)
+}
+func (p *Plugin) CreateHook(ctx context.Context, githubClient *github.Client, owner, repo string, hook github.Hook) (*github.Hook, *github.Response, error) {
+	p.API.LogWarn("log ", "githubClient", githubClient, "owner", owner)
```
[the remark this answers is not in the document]
sibasankarnayak: removed

server/plugin/command.go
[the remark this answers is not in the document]
mickmister: Does this call also list org-level webhooks? Also, how do we know the webhook is pointing to the Mattermost server, and that it has the correct settings?

server/plugin/command.go
```diff
@@ -358,7 +369,13 @@ func (p *Plugin) handleSubscribesAdd(_ *plugin.Context, args *model.CommandArgs,
 	} else if ghRepo != nil && ghRepo.GetPrivate() {
 		msg += "\n\n**Warning:** You subscribed to a private repository. Anyone with access to this channel will be able to read the events getting posted here."
 	}
-
+	githubHooks, _, _ := githubClient.Repositories.ListHooks(ctx, owner, repo, &github.ListOptions{PerPage: page})
+	if len(githubHooks) == 0 {
+		msg += "\nNo webhook was found for this repository or organization. Would you like the webhook to be created?"
+		msg += "\nSuggested Command, Can run any one command to create a wehook"
+		msg += fmt.Sprintf("\n`/github webhook add %s [secret:replace with webhook secret from Plugin console] [content-type:application-json/application/x-www-form-urlencoded] [features:create/delete..]  [ssecure-ssl:true/false]`", owner)
+		msg += fmt.Sprintf("\n`/github webhook add %s/%s [secret:replace with webhook secret from Plugin console] [content-type:application-json/application/x-www-form-urlencoded] [features:create/delete..]  [ssecure-ssl:true/false]`", owner, repo)
```
mickmister, line 377: @hanzei Should we put the secret here in the output, or have the admin go to the system console as noted here?

plugin.json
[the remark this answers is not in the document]
mickmister: > yes have tried creating base branch as the other one but it only let us to create PR with MM branch

@sibasankarnayak Are you able to create branches in Mattermost's repo with your current permissions? cc @hanzei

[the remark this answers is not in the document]
sibasankarnayak: @mickmister No tried creating a branch and push but unable to do so
cc @hanzei 
Screenshot 2021-09-10 at 7 48 02 AM

server/plugin/command.go
sibasankarnayak: @mickmister @hanzei 
made a change where the secret is picked directly from github plugin console , no need to pass it explicitly

```
   /github webhook add sibasankarnayak/mattermost-server  callback_url
```

Note : callback url is only need to pass it runs on localhost as github doesn't permit to add localhost url

mickmister:
@sanjaydemansol @maisnamrajusingh For any requests you've made on this PR that have been fulfilled, please click "Resolve Conversation" at the bottom of the request's discussion

server/plugin/command.go
```diff
@@ -506,6 +522,171 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
 	}
 }
 
+func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, parameters []string, userInfo *GitHubUserInfo) string {
+	if len(parameters) == 0 {
+		return "Invalid webhook command. Available command are `add` and `list` ."
```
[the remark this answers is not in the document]
mickmister:
```suggestion
Invalid webhook command. Available commands are `add` and `list`.
```

server/plugin/command.go
mickmister: If the site URL is configured to use `localhost`, can we just return an error saying "please set your site URL to something other than localhost". The secret and callback URL will then be optional parameters, then we can keep the command nice and short.

sibasankarnayak: @mickmister currently site url will give  error if it is local host and secret is not at all needed  and it is optional
picking up from plugin console directly

server/plugin/command.go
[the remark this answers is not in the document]
sibasankarnayak: Yes it can , but it can not verify automatically , 
Ul will be showing all the web-hook id which is hyperlink can visit and verify if they want

server/plugin/command.go
mickmister: @sibasankarnayak Can we make it so the `callback_url` is optional? I was suggesting we assume that the site URL is not localhost in the main case, so the admin would not need to supply a `callback_url`.

The only time they need to supply a `callback_url` is if the site URL _is_ localhost. So let's not make them supply the `callback_url` when the site URL is not localhost.

sibasankarnayak: @mickmister current the callback url is optional , you can try it if your ulr is not local host

maisnamrajusingh:
@mickmister doing now. will update you once i sync with @sibasankarnayak  on this.

mickmister:
@hanzei I'm seeing this feature as an opportunity to support multiple webhook secrets. There was recently a request for this functionality. https://community-daily.mattermost.com/core/pl/uf5fzfmu97bbfqfw1qgjjebh4h

Out of scope of this PR, but if we move secrets to the kv store instead of plugin settings, any webhooks we automate the creation of we can store a separate secret for.

@sibasankarnayak Something else that comes to mind: Are we able to put some sort of "id" on the webhook request, so we can know if this plugin was the creator of the webhook? Then when we fetch the webhooks, we can definitively know which ones are for our Mattermost instance.

server/plugin/command.go
```diff
@@ -501,6 +518,163 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+				if githubResponse.NextPage == 0 {
+					break
+				}
+				opt.Page = githubResponse.NextPage
+			}
+		}
+
+		var txt string
+		if len(githubHooks) == 0 {
+			txt = "Currently there are no webhook in this [Owner/Repo]"
```
mickmister, line 635: Can we be more specific here? Like:

```suggestion
			txt = "There are currently no GitHub webhooks created for the [%s](https://github.com/%s) org or the [%s/%s](https://github.com/%s/%s) repository."
```

Which should ultimately render as:

There are currently no GitHub webhooks created for the [mattermost](https://github.com/mattermost) org or the [mattermost/mattermost-plugin-github](https://github.com/mattermost/mattermost-plugin-github) repository.

server/plugin/command.go
mickmister:
```suggestion
			txt = "### Webhooks in this Repository\n"
```

server/plugin/command.go
```diff
@@ -501,6 +518,163 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+			if repo != "" {
+				hookURL += "/" + repo
+			}
+			hookURL += "/settings/hooks/"
+			txt += fmt.Sprintf(" *  [%s](%s) :  -  %s \n", hookID, hookURL+hookID, strings.Join(hook.Events, " , "))
+		}
+
+		return txt
+	}
+	return "Invalid action, only `add` and `list` commands are supported."
```
mickmister, line 652: Can we split `handlewebhook` into a few different functions, so this is a little more readable? Something like:

```go
switch command {
	case add:
            txt, err := p.handleWebhookAdd(args)
            // log err if not nil
            return txt
        case list:
            txt, err := p.handleWebhookList(args)
            // log err if not nil
            return txt
        default:
            return "Invalid action, only `add` and `list` commands are supported."
```

func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, parameters []string, userInfo *GitHubUserInfo) string {

server/plugin/command.go
```diff
@@ -501,6 +518,163 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+}
+func (p *Plugin) CheckOptionsValid(options []string) error {
+	for _, val := range options {
+		for _, item := range Manifest.Props["webhook_events"].([]interface{}) {
+			if item.(string) != val {
+				return errors.New(val + " is not a valid events to trigger webhook")
+			}
+		}
+	}
```
mickmister, line 661: What's the point of having these in the manifest versus in the code?

mickmister (changes requested):
Thanks for the updates on this PR :+1: I have some requests and questions

server/plugin/command.go
```diff
@@ -501,6 +518,163 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+					if strings.Contains(val, "*") && len(val) == 1 {
+						hook.Events = []string{"*"}
+						break
+					}
+					if err := p.CheckOptionsValid(strings.Split(val, ",")); err != nil {
+						return err.Error()
+					}
+					hook.Events = strings.Split(val, ",")
+				}
+			}
```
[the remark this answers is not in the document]
sibasankarnayak: @mickmister tried with slash command but it was too confusing for user i felt

currently we need 5 parameters from user
[org/repo]  - 
callBackUrl - optional ( default to MMURL +"plugin/github/plugin")
content-type - optional ( default x-www-form-urlencoded)
event - optional ( * -default , create ,delete .... ) 
insecureSSL -  optional ( true / false , default to false) - already implemented with flags

but while parsing instead of taking in sequence as there are lot of parameter so tried this way,
if needed can do otherway , please let me know your view

server/plugin/command.go
sibasankarnayak: Sorry , i misunderstood the comment from hanzei https://github.com/mattermost/mattermost-plugin-github/pull/456#discussion_r670259213 , fixed this in this commit.

server/plugin/command.go
sibasankarnayak: Sure , implemented this

server/plugin/command.go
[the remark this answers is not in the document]
mickmister: > tried with slash command but it was too confusing for user i felt

Confusing in what way?

> currently we need 5 parameters from user

Since 4 are optional, I would say we only need 1, but that is beside the point. 2/5 we should use flags instead of guessing what field they are filling out. The required one can be positional if that is possible, but all the others should be flag arguments.

server/plugin/command.go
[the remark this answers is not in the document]
mickmister: @hanzei Please see my question above:

> Are we intending to have the plugin be a tool for managing other webhooks not related to the plugin? It seems like we're straying away from the purpose of the feature.

What purpose does the webhook command serve other than just accepting the org/repo to subscribe to? I'm not sure I understand why we support so many arguments when the main use case is to configure the webhooks for this plugin, which only requires the admin to provide the org or repo.

server/plugin/command.go
```diff
@@ -501,6 +562,173 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+	}
+
+	hookURL := baseURL + owner
+	if repo != "" {
+		hookURL += "/" + repo
+	}
+	hookURL += "/settings/hooks/"
+	txt := "Webhook Created Successfully \n"
+	hookID := strconv.Itoa(int(*githubHook.ID))
+	txt += fmt.Sprintf(" *  [%s](%s) :  -  %s \n", hookID, hookURL+hookID, strings.Join(hookDetails.Events, " , "))
```
mickmister, line 629: 0/5 what do you think of this approach?

```suggestion
	txt := "Webhook Created Successfully \n"

	hookURL += "/settings/hooks"
	events := strings.Join(hookDetails.Events, " , ")
	txt += fmt.Sprintf(" *  [%s](%s/%v) :  -  %s \n", hookID, hookURL, *githubHook.ID, events)
```

server/plugin/command.go
```diff
@@ -501,6 +562,173 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+		}
+		hookURL += "/settings/hooks/"
+		txt += fmt.Sprintf(" *  [%s](%s) :  -  %s \n", hookID, hookURL+hookID, strings.Join(hook.Events, " , "))
+	}
+
+	return txt
+}
+func (p *Plugin) handlewebhook(_ *plugin.Context, args *model.CommandArgs, parameters []string, userInfo *GitHubUserInfo) string {
+	if len(parameters) == 0 {
+		return "Invalid webhook command. Available command are `add` and `list` ."
```
mickmister, line 693:
```suggestion
		return "Invalid webhook command. Available commands are `add` and `list` ."
```

server/plugin/command.go
```diff
@@ -501,6 +562,173 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+	case list:
+		return p.handleWebhookList(ctx, parameters, baseURL, githubClient)
+	}
+	return "Invalid action, only `add` and `list` commands are supported."
+}
+func (p *Plugin) CheckOptionsValid(options []string) error {
+	for _, val := range options {
+		for _, item := range webhookEvents {
+			if item != val {
+				return errors.New(val + " is not a valid events to trigger webhook")
```
mickmister, line 712:
```suggestion
				return errors.New(val + " is not a valid event to trigger webhook")
```

server/plugin/command.go
```diff
@@ -501,6 +562,173 @@ func (p *Plugin) handleIssue(_ *plugin.Context, args *model.CommandArgs, paramet
...
+	baseURL := p.getBaseURL()
+	switch command {
+	case add:
+		return p.handleWebhookAdd(ctx, parameters, baseURL, githubClient)
+	case list:
+		return p.handleWebhookList(ctx, parameters, baseURL, githubClient)
+	}
+	return "Invalid action, only `add` and `list` commands are supported."
+}
+func (p *Plugin) CheckOptionsValid(options []string) error {
```
mickmister, line 708: Functions should always have a space between them. Not sure how CI didn't catch this

```suggestion
}

func (p *Plugin) CheckOptionsValid(options []string) error {
```

server/plugin/command.go
```diff
@@ -340,6 +389,12 @@ func (p *Plugin) handleSubscribesAdd(_ *plugin.Context, args *model.CommandArgs,
 			}
 			subOrgMsg += "\n\n" + fmt.Sprintf("Notifications are disabled for %s", excludeMsg)
 		}
+		githubHooks, _, _ := githubClient.Repositories.ListHooks(ctx, owner, repo, &github.ListOptions{PerPage: page})
+		if len(githubHooks) == 0 {
+			subOrgMsg += "\nNo webhook was found for this repository or organization. Would you like the webhook to be created?"
+			subOrgMsg += "\nSuggested Command to create a wehook"
+			subOrgMsg += "\n`/github webhook add " + owner + "[content-type:application-json/application/x-www-form-urlencoded] [features:create/delete..]  [ssecure-ssl:true/false]`"
```
mickmister, line 396: We should only be showing the standard command. Also, please put this text in a shared function to be used in the other case where we output this information.

server/plugin/command.go
mickmister: > The response should make it clear that the command needs to be run. Also there is a `/` missing at the beginning.
> 
> Are both webhooks needed in any case? Is the org specific enough? Can the text make the answer to these questions clear?

@sibasankarnayak Please make sure you address @hanzei's points here

mickmister:
There are a lot of unresolved comments on this PR. @sanjaydemansol @maisnamrajusingh Can you please resolve the conversations for any requests you've made on this PR that have been fulfilled? Please click "Resolve Conversation" at the bottom of the request's discussion.

dipak-demansol:
@mickmister some days ago this PR assigned directly to me by siba but as per your recent comment can you pls tell me should i test this PR?

hanzei (commented):
Let's keep this PR on the side until https://github.com/mattermost/mattermost-plugin-github/pull/456 is merged as they will heavily conflict

server/plugin/command.go
mickmister: @sibasankarnayak We should make it clear that one command will create a webhook for the repo, and the other will create one for the whole org. And we should explain that if the webhook is made for the org, then you won't need to create another webhook for any repos in that org going forward.

PR #467, opened 2021-08-02
MM-36778 - Add support for Chimera Proxy

Szymongib:
#### Summary
Chimera Proxy is an easier way to connect plugins to external OAuth providers. It provides pre-registered OAuth applications allowing users to skip the app registration step.

Design doc: https://docs.google.com/document/d/1a1AzNwxrqA5bcVonZOVu94b4vV6tT4U7fDIKBwf_aUI/edit

This PR makes it possible to connect GitHub Plugin via Chimera. It introduces:
- New config option, whether to use pre-registered application (Chimera)
- Changes validation if connection through Chimera is enabled
- Creates different OAuth config when using Chimera

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-36778


go.mod
```diff
@@ -6,9 +6,9 @@ require (
 	github.com/Masterminds/sprig/v3 v3.1.0
 	github.com/google/go-github/v31 v31.0.0
 	github.com/gorilla/mux v1.8.0
-	github.com/mattermost/mattermost-plugin-api v0.0.12-0.20200917103517-788aef584648
-	github.com/mattermost/mattermost-server/v5 v5.28.0
+	github.com/mattermost/mattermost-plugin-api v0.0.16
+	github.com/mattermost/mattermost-server/v5 v5.3.2-0.20210721100405-6688195dce2a
```
Szymongib, line 10: Note to reviewers: I want to bump it to version v5.38.0 when it is out as earlier versions do not contain `ChimeraOAuthProxyURL` config.

plugin.json
```diff
@@ -21,6 +21,13 @@
     "settings_schema": {
         "header": "The GitHub plugin for Mattermost allows users to subscribe to notifications, stay up-to-date with reviews, see the status of pull requests at a glance, and other common GitHub actions - directly from Mattermost. \n \n Instructions for setup are [available here](https://www.mattermost.com/pl/default-github-plugin#configuration).",
         "settings": [
+            {
+                "key": "UsePreregisteredApplication",
+                "display_name": "Use Preregistered OAuth Application:",
+                "type": "bool",
+                "help_text": "When true, instructs plugin to use preregistered GitHub OAuth application - application registration steps can be skipped. Requires Chimera Proxy URL to be configured for the server.",
```
larkox, line 28: I wonder if we should add more context to this. Maybe a link to what is Chimera and what are the implications of using Chimera. @aaronrothschild thoughts?

larkox (approved):
LGTM. Just one small detail.

plugin.json
Szymongib: Yes, for sure. Chimera is not yet available for users so we do not have any documentation yet, but I think adding some link to documentation could be a way to go if we are comfortable hardcoding URL here.

larkox: I don't expect documentation to change much their links, so I would go that direction. Also, this configuration will be visible for all users (including on-prem), so I feel it would be nice for them to know what is chimera.

I would like it even better if this setting is only visible for cloud, but that would mean custom settings, and I am not a big fan of those.

hanzei (approved):
Great work :tada:

hanzei:
@Szymongib Is there a way for QA to test these changes right now?

go.mod
hanzei: Should the `min_server_version` also get updated?

Szymongib:
@hanzei I have a server running on a cloud Dev environment that is running this version of the plugin and is configured with Chimera. I can invite/provide credentials to whoever needs them. I hope soon regular Cloud Plugin installations will be configured accordingly so we can wait until then.

go.mod
Szymongib: Older server versions should work fine it is just the new option will not work for them. I am not sure what are exact implications of bumping `min_server_version` (I do not want people running self-managed instances to force the upgrade for the plugin to work) so I would follow your lead on that.

Szymongib: After double-checking, it seems that we do need to upgrade the `min_server_version` or otherwise plugin configuration fails to load. I will do it together with `go.mod` update after server release.

Szymongib:
Chimera is now configured for Cloud plugin and PR servers so it should be possible to test it this way. In case of any issues, feel free to ping me!

hanzei:
Awesome, thanks for the preparation @Szymongib :+1: @DHaussermann this is ready for testing.

go.mod
hanzei: Did you meant to update the `min_server_version`?

plugin.json
```diff
@@ -21,6 +21,13 @@
     "settings_schema": {
         "header": "The GitHub plugin for Mattermost allows users to subscribe to notifications, stay up-to-date with reviews, see the status of pull requests at a glance, and other common GitHub actions - directly from Mattermost. \n \n Instructions for setup are [available here](https://www.mattermost.com/pl/default-github-plugin#configuration).",
         "settings": [
+            {
+                "key": "UsePreregisteredApplication",
+                "display_name": "Use Preregistered OAuth Application:",
+                "type": "bool",
+                "help_text": "When true, instructs plugin to use preregistered GitHub OAuth application - application registration steps can be skipped. Requires Chimera Proxy URL to be configured for the server. Cannot be used with GitHub enterprise.",
+                "default": false
```
hanzei, line 29: Is there a desire to enable this by default for cloud?

go.mod
Szymongib: Yes, sorry for confusion, I did not realise I phrase it pretty badly 😄

plugin.json
Szymongib: We do not want to enable it by default for on-prem for sure, so I am not sure if/how could we do it for cloud only. It is not super important tho but could be useful.

Szymongib:
@hanzei I have bumped the version in `go.mod` and changed `min_server_version`. Additionally added unit test for configuration.

hanzei (approved):

DHaussermann:
@Szymongib I have done a round of testing on this and it's looking very good!  No functional issues per-say. A couple follow-up questions here:
1. For On-prem customers the UX could be poor as nothing prevents you from enabling the new option. Attempting to use the plugin then throws an error when a connect command is used. Is possible to grey-out the new option when the config value for the proxy URL is empty? Not sure if this can be done with a reasonable amount of effort or if there is an alternate UX to prevent misconfiguration.
2. Will the name of the oAuth app be configurable or be updated before this goes to production. 
image
This name as well as the name you are authorizing should be more formal and I'm just curious how this can be changed.

Szymongib:
Thanks, @DHaussermann!

1. I see your point about UX for on-prem customers. We have some plans to allow them to use Chimera in the future but for now, I am not familiar with any way to gray-out the option based on whether the URL is set or not. Maybe @larkox could provide some insight if it is possible and how much effort would it require?
2. Yes, for the production environment there will be a completely different application created as part of Mattermost GitHub organization, with a proper name, description, etc. For the test environment, I have created an application on my account so that we can test things and easily change and experiment.

larkox:
One option that I don't like too much (but still doable) is having a custom setting that will show or not depending on the SKU. That way it will be completely gone for on-prem, and visible for cloud. I don't like it much because makes a bit more complex the handling of settings, and implies webapp components, which if we ever migrate the system console to work on mobile (which we may never do), would be problematic.

Szymongib:
Thanks for the information. I would opt against it then and we can speed up the process of enabling it for on-prem customers.

Szymongib:
Does it address your concerns @DHaussermann?

DHaussermann:
@stylianosrigas since we have alot of on-prem deployments, my preference would be to block off this path for on-prem customers before we package the feature in a release. 
As it stands now - The description in the config UI would at least need to change to inform people that it's a Cloud only feature. The current text will cause admins to go search the system console for a proxy config which they won't find. 

@aaronrothschild  do you have any thoughts on the impact here? I know in the past we have shipped `cloud` versions of plugins when the feature set differed but I don't think that's a path we want to go down again.

DHaussermann:
@Szymongib Sorry I forgot the ping above ☝️ 

Also, @aaronrothschild had a suggestion that may help solve this quickly. We did implement functionality to show custom controls on the plugin config pages. @larkox do you know if these may provide a faster solution so we can solve this before we ship the feature?

larkox:
As I said, we can have custom settings (the custom controls on the plugin config pages you are talking about). Here is the documentation for reference: https://developers.mattermost.com/integrate/plugins/webapp/reference/#registerAdminConsoleCustomSetting

I am not a big fan of this approach, but probably is more a personal taste than a objective decision. It would solve the problem, at least on short term.

hanzei:
@Szymongib Is there a case where a cloud customer doesn't want to use Chimera? If not, can we drop the setting and enable the feature bases on the fact that the plugin is running in Cloud?

Szymongib:
@hanzei I think we want to leave customers with the option to use their own OAuth application, which means not using Chimera. One reason is that they may already be using it so we could potentially break existing integrations, the other is that we probably want them to have a choice anyway.

hanzei:
Thanks for sharing your thoughts @Szymongib. In this case I don't see a fast option to improve the UX of on-prem. A custom setting take a lot of effort.

Could we log with a better error message when the feature is enabled for on-prem?

Szymongib:
> Could we log with a better error message when the feature is enabled for on-prem?

Sure @hanzei, we can also extend the setting description with something like `Setting intended for Cloud installations` (technically nothing prevents you from running Chimera and configuring it with env). Maybe adding a note about it to the README guide of how to set up the plugin could be useful too.

plugin.json
```diff
@@ -21,6 +21,13 @@
     "settings_schema": {
         "header": "The GitHub plugin for Mattermost allows users to subscribe to notifications, stay up-to-date with reviews, see the status of pull requests at a glance, and other common GitHub actions - directly from Mattermost. \n \n Instructions for setup are [available here](https://www.mattermost.com/pl/default-github-plugin#configuration).",
         "settings": [
+            {
+                "key": "UsePreregisteredApplication",
+                "display_name": "Use Preregistered OAuth Application:",
```
aaronrothschild, line 26:
```suggestion
                "display_name": "Use the Mattermost Integration App listed in Github's Marketplace:",
```

plugin.json
```diff
@@ -21,6 +21,13 @@
...
+                "type": "bool",
+                "help_text": "When true, instructs plugin to use preregistered GitHub OAuth application - application registration steps can be skipped. Requires [Chimera Proxy](https://github.com/mattermost/chimera) URL to be configured for the server. Cannot be used with GitHub enterprise. **This setting is intended for Cloud instances.**",
```
aaronrothschild, line 28:
```suggestion
                "help_text": "When true, instructs the plugin to use the preregistered GitHub/Mattermost OAuth application from the GitHub Marketplace.  As a result, application registration steps in the documentation can be skipped. Requires [Chimera Proxy](https://github.com/mattermost/chimera) URL to be configured for the server. Cannot be used with GitHub enterprise. **This setting is intended to be used with Cloud instances only.**",
```

aaronrothschild (commented):
Some small changes to make it clearer for the admins.  

in future, we could add a link to the actual Marketplace listing for the Mattemrost app when it becomes available on the [Github marketplace](https://github.com/marketplace?type=&verification=&query=mattermost+), then this will make more sense to the average user.

plugin.json
Szymongib: @aaronrothschild  Do we actually plan to list the app in Github Marketplace? I do not think there is actually any point in doing that as with Chimera customers will just authorize our existing OAuth app and it does not have anything to do with Marketplace. 

I might not be familiar with the whole vision, so could elaborate what is the goal here?

Szymongib:
FYI I have added a link to the Chimera repository in the custom setting description. This link should not change and this PR https://github.com/mattermost/chimera/pull/5 will add some documentation to give users a better understanding. We can later add links to the official documentation in Chimera README after we will have it.

Szymongib:
So after bumping the Mattermost dependency version to v6.0 there is an issue with Chimera as due to the convention changes the setting is now called `ChimeraOAuthProxyURL` instead of `ChimeraOAuthProxyUrl` which would result in plugin requiring version 6+ (otherwise the value is empty).

We can work around it by trying to directly read `MM_PLUGINSETTINGS_CHIMERAOAUTHPROXYURL` env var if the setting was not provided (which is the way this will be set on Cloud) but I am not sure how you feel about that @hanzei.

I will wait with resolving conflicts until we decide on some plan.

Also if I am not mistaken, with changes to v6 the plugin might not work with servers older than v5.37 as there were some breaking changes. I encountered those when implementing initial Chimera support.

cc @gabrieljackson

Szymongib:
So I decided to move forward with the implementation of fallback to the env var if the Chimera URL setting is empty.  We probably can get rid of it when we drop support for Mattermost v5 in the plugin. 

Let me know if you have any objections or see any issues with that @hanzei.

jasonblais:
@aaronrothschild @hanzei @DHaussermann Thanks everyone for the thoughtful feedback so far.

Curious about next steps and how we can help move the reviews forward? We'd like to get ready to ship the changes ahead of the Oct 13 launch, and would be happy to help where we can.

DHaussermann:
@jasonblais Unless I'm mistaken, I think we were okay to ship this after the last change in the System Console config text explaining it's for Cloud instances only.

cc @aaronrothschild

DHaussermann:
@Szymongib Thanks for the last set of changes. 

Not sure I understand why we need a fall back to an environment var when the config is empty. Do we have a case where the config may not be set? Either way I can test this locally with some token value and see if the plugin starts.

Also Regarding the wording I wonder if you could ass 1 more small change in the last sentence.
image
Can we change "This setting is intended for Cloud instances." to "This setting is intended for **Mattermost** Cloud instances."
My concern is that the term cloud be ambiguous to some people. 
cc @aaronrothschild

hanzei (commented):
I'm wondering if the workaround for the naming change in `ChimeraOAuthProxyURL` is really needed. Cloud will use `v6.0` anyway and the Chimera Proxy feature isn't suitable for on-prem. Is there any server that runs v5.X and will use this feature?

hanzei (approved):

Szymongib:
Sure @hanzei I guess we can do that.

@DHaussermann the reasoning behind the env was explained here https://github.com/mattermost/mattermost-plugin-github/pull/467#issuecomment-918178108 but we agreed to remove it so this feature will now only work with v6 servers.

I have also adjusted the setting description.

plugin.json
aaronrothschild: Generally, we want customers to be aware that our solution can be integrated with other solutions.  It builds awareness of Mattermost as an option and if someone is using Mattermost when scanning the Marketplace, they are more likely to then integrate it.  I realize now that you are using the single user's account to generate the app programmatically in their environment, but if there is an App that CAN be listed in their Marketplace, I would like us to do that.

plugin.json
aaronrothschild: IMO, yes this should be enabled by default in Cloud so the Admin could simply add Github, then install the app after first logging in,right?

aaronrothschild (approved):

DHaussermann (approved):
Tested on passed
- Proxy and oAuth are functional based on previous testing
- Plugin config label has been updated for clarity
- Confirmed with local testing that the plugin will start-up with proxy enabled using an environment variable instead of a value form the config. 

LGTM!

jasonblais:
> IMO, yes this should be enabled by default in Cloud so the Admin could simply add Github, then install the app after first logging in,right?

Yes, the plan is to enable the Chimera proxy to all Cloud workspaces by default (may do a staged rollout, where we enable it to a subset of workspaces at first).

PR #459, opened 2021-07-14
 mm 332 - Use SiteURL for all calls from the webapp

sibasankar-demansol:
Added Site Url for call from webapp

ticket here
https://github.com/mattermost/mattermost-plugin-github/issues/332

PR #478, opened 2021-08-31
[mm 477] Master branch no longer shows UI elements when user connects to GitHub fixes #477

### LINKED ISSUE #477: Master branch no longer shows UI elements when user connects to GitHub
I have notice that there seems to be an issue in the master branch of GitHub. When the user connects to GitHub the LHS no longer shows the list of tasks. (the slash command still works to retrieve this data)

Also the post menu options to create or attach to GitHub are no longer shown.
I can't see anything in the console or logs to show what specifically is causing this

This does not occur when installing the latest production release on an affected server.   

Steps:
- Install and configure GitHub
- Enable the plugin
- Connect to GitHub
**Observed** : User is connected but the UI has not changed.

Captured in Jira here: https://mattermost.atlassian.net/browse/MM-38232

sibasankarnayak:
Fixed issue where we made HTTP requests prior to fetching URL Details which lead to all HTTP being made to wrong URLs

ticket here
https://github.com/mattermost/mattermost-plugin-github/issues/477

### DIFF
--- a/webapp/src/index.js
+++ b/webapp/src/index.js
@@ -49,7 +49,6 @@ class PluginClass {
         registry.registerWebSocketEventHandler(`custom_${pluginId}_createIssue`, handleOpenCreateIssueModal(store));
         registry.registerReconnectHandler(handleReconnect(store));
 
-
         activityFunc = () => {
             const now = new Date().getTime();
             if (now - lastActivityTime > activityTimeout) {
--- a/webapp/src/index.js
+++ b/webapp/src/index.js
@@ -24,6 +24,7 @@ const activityTimeout = 60 * 60 * 1000; // 1 hour
 class PluginClass {
     async initialize(registry, store) {
         registry.registerReducer(Reducer);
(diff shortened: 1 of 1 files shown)


sanjaydemansol:
LGTM

hanzei (approved):
Thanks for the fix :+1: 

Can you elaborate why the testing in https://github.com/mattermost/mattermost-plugin-github/pull/459 didn't found this bug? Is there something we can learn for this issue?

hanzei:
@sibasankarnayak What do you think about my comment above https://github.com/mattermost/mattermost-plugin-github/pull/478#pullrequestreview-742713796?

sibasankarnayak:
> @sibasankarnayak What do you think about my comment above [#478 (review)](https://github.com/mattermost/mattermost-plugin-github/pull/478#pullrequestreview-742713796)?

this was a edge case where we missed to store the state value , every time the page refreshed the state gets reset to default value

hanzei:
Was the really an edge case? Without the correct server route the UI elements would never show up, right?

PR #519, opened 2022-01-03
Update README.md

bobychaudhary:
# Minor typo mistake
```diff
-respository 
+repository
```


hanzei (approved):
Good catch and thanks for the fix :+1:

PR #534, opened 2022-02-16
[GH-470] allow users to specify notifications verbosity using a new render-style flag

ijansky:
#### Summary
This pull request adds a `--render-style` flag to the subscription command to define the notification verbosity of an event rather than the full body as it can be quite lengthy and easily spam a team mattermost channel. It is a shared effort of myself and @TeodorPt and we raised this PR to get early feedback on the direction that this PR takes

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/470


hanzei (approved):
Great work :+1:

This change will increase the load of the webhook handler a bit as there is additional parsing of the template, but that is fine.

ijansky:
Thanks for the review @hanzei!

hanzei:
Your welcome! There was nothing to complain about :wink:

rinkimekari:
I tested this fix out because it has to do with my PR https://github.com/mattermost/mattermost-plugin-github/pull/545 and I wanted to compare them. However, I can't seem to get it to work. Deploying the plugin works fine, and autocomplete for the command works perfectly, but whether or not the flag is enabled, it always shows the normal expanded message for notifications. There should definitely be tests in `template_test.go` for the cases when the flag is enabled.

mickmister:
@ijansky Do you mind posting a screenshot of what this feature looks like?

TeodorPt:
Hi @mickmister @rinkimekari, sorry for the delay, had a busy week. I have tested this locally now and it works - the way to subscribe would be `/github subscriptions add [owner/repo] [features] --title-only`. See screenshots below:

Long version:
image

Title-only:
image

Of course I will add unit tests for the flag enabled as well before merging. Also the "title-only" version could be displayed differently if you have other preferences. 
But we opened this PR initially as a request for comments, to get a sense of the direction and see if you would prefer this approach, since we identified 2 ways of doing it:
1. *This approach:* passing an event with config options to the template, and then letting the template decide how to present the event based on the config.
2. *Multiple templates approach* (similar to @rinkimekari 's solution): having one template for no-title/collapsed and one for full and deciding between the two.

We went for the former approach because it keeps a nice separation of rendering logic and business logic (whole rendering is the responsibility of the template itself), and it makes it easier to add additional configuration options in the future, without having to add if-else statements where the templates are used.

It seems there was duplicate work here, since there were 2 opened issues for the same issue. Let me know if you're happy for me to proceed and I'll turn this PR into something merge-worthy asap.

Thanks!

TeodorPt:
@mickmister can you help me with a review when you have the chance? Thank you very much!

server/plugin/command.go
```diff
@@ -600,12 +600,18 @@ func getAutocompleteData(config *Configuration) *model.AutocompleteData {
...
+				HelpText: "Events triggered by organization members will not be delivered (the organization config should be set, otherwise this flag has not effect)",
+				Hint:     "(optional)",
+				Item:     "--exclude-org-member",
+			},
+			{
+				HelpText: "Notifications will be delievered with titles only (for example, the body of a pull request will not be displayed)",
+				Hint:     "(optional)",
+				Item:     "--title-only",
+			}}
+		subscriptionsAdd.AddStaticListArgument("Currently supports --exclude-org-member and --title-only", false, flags)
```
mickmister, line 614: In the state of this PR, we are not able to use both the `exclude-org-member` _and_ `title-only` flags simultaneously. This is because they are being treated as a single positional argument, each showing up as a single selectable option in the same argument.

Similar to what was discussed in https://github.com/mattermost/mattermost-plugin-github/pull/505#discussion_r796253955, we should use "named" arguments with explicit `true/false` values, rather than using positional arguments that implicitly mean `true`.

Also, this `title-only` flag should be available even when `config.GitHubOrg` is not defined, but it's currently added in this if block which restricts to when `config.GitHubOrg` is defined.

server/plugin/webhook.go
```diff
@@ -61,6 +74,17 @@ func signBody(secret, body []byte) ([]byte, error) {
 	return computed.Sum(nil), nil
 }
 
+// GetEventWithRenderConfig wraps any github Event into an EventWithRenderConfig
+// which also contains per-subscription configuration options.
+func GetEventWithRenderConfig(event interface{}, sub *Subscription) *EventWithRenderConfig {
+	return &EventWithRenderConfig{
+		Event: event,
```
mickmister, line 81: Do we lose any static typing from resorting to `interface{}`? I think not because these are just being passed into templates, so the event is not referenced in statically-typed Go code anywhere.

mickmister (changes requested):
Thanks @ijansky and @TeodorPt for implementing this!

This mostly looks good. I just have one request to modify how the command flags work in the command autocomplete, so that we can use multiple flags in the same command.

server/plugin/template_test.go
```diff
@@ -301,11 +310,14 @@ git-get-head gets the non-sent upstream heads inside the stashed non-cleaned app
...
-			Sender:      &user,
-		})
+		actual, err := renderTemplate("newPR", GetEventWithRenderConfig(
+			&github.PullRequestEvent{
+				Repo:        &repo,
+				PullRequest: &pullRequestWithMultipleLabelsAndAssignees,
+				Sender:      &user,
+			},
+			nil,
+		))
```
mickmister, line 320: Can we add tests for all templates that support the `title-only` flag?

TeodorPt:
Hi @mickmister, I've made several changes to the PR, PTAL again:
- flags now accept a value as well, per your suggestion
- taking advantage of that, the flag is not not `title-only` anymore, but `render-style`, which could either be `collapsed`, `skip-body` or default
- added tests

Here's the rendered output for all 3 formats in my tests:
Screenshot from 2022-04-01 14-55-50

server/plugin/subscriptions.go
```diff
@@ -15,24 +16,39 @@ import (
 const (
 	SubscriptionsKey     = "subscriptions"
 	excludeOrgMemberFlag = "exclude-org-member"
+	renderStyleFlag      = "render-style"
```
TeodorPt, line 19: We could enrich the flags to also include supported values, instead of only the name. Then we could have a generic validation function lower down in `AddFlag`.

mickmister: This is likely out of scope, since we would want to apply this to all flags potentially. It would be a great addition if you're looking to implement this separately 🙂

mickmister: @TeodorPt I meant to mention you in the comment above

TeodorPt: @mickmister sure, I'll take that as a separate thing, but it might take a while as my schedule is quite full. In the meantime, this PR is ready for a re-review :)

README.md
```diff
@@ -149,6 +149,8 @@ When you’ve tested the plugin and confirmed it’s working, notify your team s
    ```
   - The following flags are supported:
      - `--exclude-org-member`: events triggered by organization members will not be delivered. It will be locked to the organization provided in the plugin configuration and it will only work for users whose membership is public. Note that organization members and collaborators are not the same.
+     - `--render-style`: notifications will be delievered in the specified style (for example, the body of a pull request will not be displayed). Supported
```
mickmister, line 152:
```suggestion
     - `--render-style`: notifications will be delivered in the specified style (for example, the body of a pull request will not be displayed). Supported 
```

server/plugin/template.go
```diff
@@ -352,6 +364,7 @@ Assignees: {{range $i, $el := .Assignees -}} {{- if $i}}, {{end}}{{template "use
 		"    * Defaults to `pulls,issues,creates,deletes`\n" +
 		"  * `flags` currently supported:\n" +
 		"    * `--exclude-org-member` - events triggered by organization members will not be delivered (the GitHub organization config should be set, otherwise this flag has not effect)\n" +
+		"    * `--render-style` - notifications will be delievered in the specified style (for example, the body of a pull request will not be displayed). Supported values are `collapsed`, `skip-body` or `default` (same as omitting the flag).\n" +
```
mickmister, line 367:
```suggestion
		"    * `--render-style` - notifications will be delivered in the specified style (for example, the body of a pull request will not be displayed). Supported values are `collapsed`, `skip-body` or `default` (same as omitting the flag).\n" +
```

server/plugin/subscriptions.go
```diff
@@ -15,24 +16,39 @@ import (
...
 )
 
 type SubscriptionFlags struct {
 	ExcludeOrgMembers bool
+	RenderStyle       string
 }
 
-func (s *SubscriptionFlags) AddFlag(flag string) {
+func (s *SubscriptionFlags) AddFlag(flag string, value string) error {
 	switch flag { // nolint:gocritic // It's expected that more flags get added.
```
mickmister, line 28: I think we can get rid of this linter exclusion here now

```suggestion
	switch flag {
```

server/plugin/command.go
```diff
@@ -656,15 +664,33 @@ func getAutocompleteData(config *Configuration) *model.AutocompleteData {
...
+		})
+	}
+	subscriptionsAdd.AddNamedStaticListArgument("render-style", "Determine the rendering style of various notifications.", false, []model.AutocompleteListItem{
+		{
+			Item:     "default",
+			HelpText: "The default rendering style for all notifications (includes all information). Can be omitted.",
+		},
+		{
+			Item:     "skip-body",
+			HelpText: "Skips the body part of various long notifications that have a body (e.g. new PRs and new Issues).",
```
mickmister, line 685:
```suggestion
			HelpText: "Skips the body part of various long notifications that have a body (e.g. new PRs and new issues).",
```

server/plugin/command.go
```diff
@@ -656,15 +664,33 @@ func getAutocompleteData(config *Configuration) *model.AutocompleteData {
...
+			{
+				Item:     "false",
+				HelpText: "Include posts from members of the configured organization",
+			},
+		})
+	}
+	subscriptionsAdd.AddNamedStaticListArgument("render-style", "Determine the rendering style of various notifications.", false, []model.AutocompleteListItem{
+		{
+			Item:     "default",
+			HelpText: "The default rendering style for all notifications (includes all information). Can be omitted.",
```
mickmister, line 681: I think "Can be omitted" is a little confusing because it sounds like the `value` of this flag can be ommitted, even though the flag parsing code will not allow for that.

```suggestion
			HelpText: "The default rendering style for all notifications (includes all information).",
```

server/plugin/command.go
```diff
@@ -656,15 +664,33 @@ func getAutocompleteData(config *Configuration) *model.AutocompleteData {
...
+		},
+		{
+			Item:     "skip-body",
+			HelpText: "Skips the body part of various long notifications that have a body (e.g. new PRs and new Issues).",
+		},
+		{
+			Item:     "collapsed",
+			HelpText: "Notifications come in a one line format, without enlarged fonts or advanced layouts.",
```
mickmister, line 689:
```suggestion
			HelpText: "Notifications come in a one-line format, without enlarged fonts or advanced layouts.",
```

server/plugin/command.go
```diff
@@ -656,15 +664,33 @@ func getAutocompleteData(config *Configuration) *model.AutocompleteData {
...
+				Item:     "true",
+				HelpText: "Exclude posts from members of the configured organization",
+			},
+			{
+				Item:     "false",
+				HelpText: "Include posts from members of the configured organization",
+			},
+		})
+	}
+	subscriptionsAdd.AddNamedStaticListArgument("render-style", "Determine the rendering style of various notifications.", false, []model.AutocompleteListItem{
```
mickmister, line 678: Suggestion to add a newline here

```suggestion
	}

	subscriptionsAdd.AddNamedStaticListArgument("render-style", "Determine the rendering style of various notifications.", false, []model.AutocompleteListItem{
```

mickmister (approved):
LGTM :tada: I just have some wording suggestions

Thanks @ijansky and @TeodorPt!

mickmister:
@TeodorPt Please avoid force-pushing, as GitHub makes it difficult to keep track of changes over time. Personally, I like deterministically seeing the diff between reviews. Unfortunately when this occurs I have to re-review the whole PR since there is no way of me knowing that the previous commits do not have other changes in them.

TeodorPt:
> @TeodorPt Please avoid force-pushing, as GitHub makes it difficult to keep track of changes over time when this happens. Personally, I like deterministically seeing the diff between reviews. Unfortunately when this occurs I have to re-review the whole PR since there is no way of me knowing that the previous commits do not have other changes in them.

I agree, I didn't change anything though. My email alias was wrong and that's why I forced pushed.

mickmister:
@TeodorPt Also, if you'd like to have me re-review after changes have been addressed, please explicitly re-request review via GitHub's request feature

mickmister:
> there is no way of me knowing that the previous commits do not have other changes in them.

> I agree, I didn't change anything though.

@TeodorPt This is mainly a concern for security. Being an open-source project, we have to be absolutely sure what changes are being introduced, which means we need to review every line. I'm sure you are trustworthy, but I need to verify the changes nonetheless.

TeodorPt:
> > there is no way of me knowing that the previous commits do not have other changes in them.
> 
> > I agree, I didn't change anything though.
> 
> @TeodorPt This is mainly a concern for security. Being an open-source project, we have to be absolutely sure what changes are being introduced, which means we need to review every line. I'm sure you are trustworthy, but I need to verify the changes nonetheless.

From that angle I have to agree. Thanks for the perspective :) please re-review then and I will avoid force-pushing any further.

mickmister:
@dipak-demansol This is ready for QA review

mickmister:
@dipak-demansol Please prioritize this review if you can

dipak-demansol (changes requested):
so when i use command like this `/github subscriptions add dipak-demansol pull --render-style default ` then its LGTM but if i don't write the event and i want to use the default events at that point i don't see the auto-complete option, so my request is when user write   `/github subscriptions add dipak-demansol` and add space then the **--render-style** auto-complete + other 3 style autocomplete like default and other Two should visible to the user.

TeodorPt:
So you're saying flags don't appear in the auto-complete if no event is specified? I think this was pre-existing with the other flag as well, but I'll take a look to see how we can make it appear.

dipak-demansol:
> So you're saying flags don't appear in the auto-complete if no event is specified? I think this was pre-existing with the other flag as well, but I'll take a look to see how we can make it appear.

Yes, flags is not appeared in the auto-complete if no event is specified.

mickmister:
Let's move the `features` argument to a flag argument instead of a positional, to properly treat it as an optional argument among the other optional arguments. `features` should be the first listed flag in the list for discoverability.

cc @TeodorPt @dipak-demansol

TeodorPt:
@mickmister @dipak-demansol ptal again, now features is a flag and needs to be added by something like `--features pulls,issues`

dipak-demansol (approved):
LGTM

TeodorPt:
@mickmister ptal, this needs a maintainer approval it seems

server/plugin/subscriptions.go
```diff
@@ -17,6 +17,7 @@ const (
 	SubscriptionsKey     = "subscriptions"
 	excludeOrgMemberFlag = "exclude-org-member"
 	renderStyleFlag      = "render-style"
+	FeaturesFlag         = "features"
```
mickmister, line 20: @TeodorPt What do you think about namespacing the constants like so:

```suggestion
	flagExcludeOrgMember = "exclude-org-member"
	flagRenderStyle      = "render-style"
	flagFeatures         = "features"
```

mickmister (approved):
LGTM! Just one suggestion to namespace the flag name constants

Thanks @TeodorPt @ijansky @dipak-demansol

TeodorPt:
It's been a while since updating this, been caught up in a lot of other things. I've made the changes and will do a re-test asap and ping you when it's ready (some changes to the documentation were also needed).

TeodorPt:
@mickmister I finally got around to addressing your comments. I also identified some omissions in the `help` documentation - I've updated that as well and re-tested everything end-to-end to make sure things are still working.

I've forced-pushed due to CI tests timing out (you can check with Compare that no changes were made in the force push). However, the tests are still timing out, do you know why?

Thank you!

mickmister (approved):
LGTM!

mickmister:
@DHaussermann After looking at the changes from last QA review from Dipak, there are no significant changes, so I think this is good to merge

DHaussermann (approved):
Tested and passed
- All 3 renders styles work as expected
- Render styles behaves as default when not set
- ` --render-style` shows in subscriptions list when set
LGTM!

Huge thanks @ijansky for this enhancement 🎉  
Sorry for the delay on this.

ijansky:
@TeodorPt thank you for addressing code review comments. 👍

catalintomai:
@ijansky , feel free to merge it in. Thanks!

PR #542, opened 2022-02-25
Fix MM-42096

hanzei:
#### Ticket Link
Fixes https://mattermost.atlassian.net/browse/MM-42096

### DIFF
--- a/server/plugin/webhook.go
+++ b/server/plugin/webhook.go
@@ -6,7 +6,7 @@ import (
 	"crypto/sha1" //nolint:gosec // GitHub webhooks are signed using sha1 https://developer.github.com/webhooks/.
 	"encoding/hex"
 	"encoding/json"
-	"io/ioutil"
+	"io"
 	"net/http"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


levb (approved):

server/plugin/webhook.go
```diff
@@ -147,13 +147,25 @@ func (wb *WebhookBroker) Close() {
 func (p *Plugin) handleWebhook(w http.ResponseWriter, r *http.Request) {
 	config := p.getConfiguration()
 
-	signature := r.Header.Get("X-Hub-Signature")
-	body, err := ioutil.ReadAll(r.Body)
+	body, err := io.ReadAll(r.Body)
```
jupenur, line 150: I'm not sure what's happening here; why is this any better than the original? We're still reading the whole body regardless of size, right?

hanzei: Yes, this PR doesn't change how the body is read, But it changes how the body is parsed. We now check the HMAC before calling `github.ParseWebHook`.

jupenur: Got it; and I think that's fine and a good defense-in-depth improvement, knowing that there is a separate size limit in the server.

jupenur (approved):
LGTM

Issue #544, opened 2022-02-25
Mattermost system admin must reconnect their account after upgrading.     


DHaussermann:
When GitHub is upgraded for `v2.0.1` to `v2.1.0`  sysadmin will need to reconnect their accounts. 

As shown by this screen, users who are not admins seem unaffected. 
![image](https://user-images.githubusercontent.com/35588783/155808244-e627abbb-8f89-4084-b3d2-b9a13693b8eb.png)

Not necessarily an issue but, the owner of the GitHub oAuth app will get an email about change in scope which seems directly related to this issue. 
![image](https://user-images.githubusercontent.com/35588783/155808150-de69bab8-483c-461d-b5b4-b979d73d871b.png)

Steps:
- Install GitHub 2.0.1 from the marketplace
- Configure the plugin and connect a connect a system admin
- Upgrade to GitHub 2.1.0
- Run `/ghithub me`

**Observed**: GitHub profile is no longer connected.

hanzei:
@DHaussermann I can't reproduce this issue. If you `/github connect` again after the update?

DHaussermann:
```
{"timestamp":"2022-02-25 17:45:38.463 Z","level":"warn","msg":"ignored duplicate step name","caller":"app/plugin_api.go:934","plugin_id":"github","name":"oauth-info","flow":"oauth"}
{"timestamp":"2022-02-25 17:45:44.154 Z","level":"warn","msg":"Failed to search for PRs","caller":"app/plugin_api.go:934","plugin_id":"github","error":"GET https://api.github.com/search/issues?q=is%3Apr+is%3Aopen+author%3ADHaussermann+archived%3Afalse+: 401 Bad credentials []","query":"is:pr is:open author:DHaussermann archived:false ","userid":"3gngahnsffnq5kuro9zcuect4h","github username":"DHaussermann"}
{"timestamp":"2022-02-25 17:45:44.193 Z","level":"warn","msg":"Failed to search for assignments","caller":"app/plugin_api.go:934","plugin_id":"github","error":"GET https://api.github.com/search/issues?q=is%3Aopen+assignee%3ADHaussermann+archived%3Afalse+: 401 Bad credentials []","query":"is:open assignee:DHaussermann archived:false ","userid":"3gngahnsffnq5kuro9zcuect4h","github username":"DHaussermann"}
[...]

PR #551, opened 2022-03-17
GH-544: Removed racy in-line token refresh from githubConnect

levb:
#### Summary
Removed racy in-line token refresh from githubConnect. 
Also:
- made sure that a cluster.Mutex is acquired before running the job
- added a KV "done" key to avoid repeating executions

#### Ticket Link
#544

### DIFF
--- a/server/plugin/mm_34646_token_refresh.go
+++ b/server/plugin/mm_34646_token_refresh.go
@@ -7,17 +7,34 @@ import (
 
 	"github.com/google/go-github/v41/github"
 	"github.com/pkg/errors"
+
+	"github.com/mattermost/mattermost-plugin-api/cluster"
 )
 
 const pageSize = 100
 const delayBetweenUsers = 1 * time.Second
-const delayToStart = 5 * time.Minute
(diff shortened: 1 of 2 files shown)


server/plugin/mm_34646_token_refresh.go
```diff
@@ -63,14 +80,16 @@ func (p *Plugin) forceResetAllMM34646() error {
 			time.Sleep(delayBetweenUsers)
 		}
 
-		if len(keys) < pageSize {
+		if len(keys) == 0 {
 			break
 		}
 	}
+
+	_ = p.API.KVSet(mm34646DoneKey, []byte("done"))
```
catalintomai, line 88: @levb - could we have used instead of "done" `tryInfo.MM34646ResetTokenDone`?

levb: sure, but the key name is already a reference

levb:
The token backfill fires 1 min after the plugin startup, and should log. You may want to check that the tokens work before and after the backfill.

catalintomai (approved):

PR #568, opened 2022-06-17
Added extra debug logs to diagnose an issue on community

levb:
DEBUGGING, DO NOT MERGE


jwilander:
Will this get us the user ID for the panic generated here: https://github.com/mattermost/mattermost-plugin-api/blob/v0.0.26/post.go#L309 ?

jwilander:
Which is called from https://github.com/mattermost/mattermost-plugin-github/blob/v2.1.1/server/plugin/plugin.go#L291

levb:
@jwilander https://github.com/mattermost/mattermost-plugin-api/tree/lev-extra-logging and updated `go.mod`; good catch!

levb:
See https://github.com/mattermost/mattermost-plugin-api/compare/lev-extra-logging?expand=1 for pluginapi

PR #598, opened 2022-11-02
Add build info

hanzei:
#### Summary
Make use of https://github.com/mattermost/mattermost-plugin-api/pull/121.

#### Ticket Link
None


plugin.json
```diff
@@ -7,7 +7,7 @@
     "release_notes_url": "https://github.com/mattermost/mattermost-plugin-github/releases/tag/v2.1.2",
     "icon_path": "assets/icon.svg",
     "version": "2.1.2",
-    "min_server_version": "6.5.0",
+    "min_server_version": "7.5.0",
```
hanzei, line 10: On older versions, the plugin does fail to start due to https://github.com/mattermost/mattermost-server/pull/21560.

javaguirre (approved):
LGTM!

plugin.json
mickmister: @hanzei How does that make it fail to start? Can we instead set the `OwnerId` in this plugin? I don't like the situation that puts the plugin in, but it's backwards compatible and forwards compatible. The newer server will ignore any value passed in as expected, so it's not affected there.

hanzei: > Can we instead set the `OwnerId` in this plugin?

Yes, we can do that, but I'm hesitant to tinker around with "manual" changes like that. The API already set the `OwnderId`.

Do you mind the updated `min_server_version`?

mickmister: > Do you mind the updated `min_server_version`?

Yes, I would very much prefer keeping the server version lower. Having it this high will make it difficult to deliver bug fixes to (even slightly) older server versions. Any way around bumping the server version?

hanzei: 7.1 is new the required minimum version as https://github.com/mattermost/mattermost-server/blob/c3c81cb3d6a29ef8164a8e7ef4ba172d60117176/plugin/api.go#L1156-L1159 was added in 7.1.

mickmister: Is there a further workaround we can provide to avoid bumping to 7.1? Some conditional logic to avoid using an unavailable plugin API method. I know this would be hacky, but this is one of our most used plugins and I would like to support the most server versions possible.

hanzei: Well, surely there is something that _can_ be done: Don't use `pluginapi.EnsureBot` and use a custom implementation instead. But that is quite a workaround.

mickmister: Ideally we would have an ESR process that spans out dot releases into older server versions, but we don't have that at the moment. It just seems like a big jump of server version requirements here. How much value are we gaining from these updates that require the newer server versions?

Also as far as compiling against a server version, which packages in the mattermost-server repo are needed to be updated, outside of what is assumed to be available via plugin API? I ask because we may be able to have separate Go modules in the server repo for plugins to add to their dependencies. `model` package or `plugin` package could be separate imported modules. This would also help with potential dependency bloat and dependency version management issues on the plugin project side.

hanzei: The packages that are commonly imported by plugins are `model` and `plugin`. Surely, we could release a separate modules file, but why? 

> This would also help with potential dependency bloat and dependency version management issues on the plugin project side.

I don't get that point. What bloat do you refer to? Currently, plugins import a single dependency: `github.com/mattermost/mattermost-server/v6`. What other issues are you referring to?

hanzei: > How much value are we gaining from these updates that require the newer server versions?

It's a necessity to continue using the `pluginapi` package. By it's nature, it stays up to date relatively well with the MM server. And plugins just have to follow the imported server version.

hanzei: > Ideally we would have an ESR process that spans out dot releases into older server versions, but we don't have that at the moment.

What is missing here in terms of process? If we found an bug as significant, that we would backport it to ESR, we can release of the latest release that is compatible with ESR.

mickmister: I meant having a separate ESR for plugin projects, but that would likely be too much overhead for the benefit it would bring. We do our best to deliver dot releases, but we don't necessarily take into account delivering those fixes to an older version of a plugin that has less minimum server version requirements.

> I don't get that point. What bloat do you refer to? Currently, plugins import a single dependency:` github.com/mattermost/mattermost-server/v6`. What other issues are you referring to?

I meant the dependencies added to `go.sum`. Do these contribute to the bundle size of the plugin?

server/plugin/command.go
```diff
@@ -531,9 +531,18 @@ func (p *Plugin) isAuthorizedSysAdmin(userID string) (bool, error) {
...
 func (p *Plugin) ExecuteCommand(c *plugin.Context, args *model.CommandArgs) (*model.CommandResponse, *model.AppError) {
-	command, action, parameters := parseCommand(args.Command)
+	cmd, action, parameters := parseCommand(args.Command)
 
-	if command != "/github" {
+	if cmd != "/github" {
+		return &model.CommandResponse{}, nil
+	}
+
+	if action == "about" {
```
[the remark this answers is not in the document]
mickmister: Since `info` is already used in other plugin projects, and has been used as a tool with the support team to gather information plugin installations, 2/5 I'm thinking we should use `info` here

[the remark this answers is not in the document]
hanzei: >  has been used as a tool with the support team to gather information plugin installations,

The only plugin used in production that uses an `info` command is Jira, right? Even that command works slightly differently. I would like for the `about` command to not include any confidential information, so it can be used for debugging without manual cleanup.

plugin.json
hanzei: >  I meant the dependencies added to `go.sum`. Do these contribute to the bundle size of the plugin?

No, they don't. Only direct and indirect - i.e. transitive - dependencies increase the bundle size. E.g.  `github.com/mattermost/mattermost-server/v6/model` imports `github.com/mattermost/mattermost-server/v6/shared/` so that package also gets compiled. `github.com/francoispqt/gojay` is another example of a package that `model` imports. Packages that are not direct or indirect imports do not contribute to the bundle site. `github.com/mattermost/morph` gets imported by `github.com/mattermost/mattermost-server/v6/store/sqlstore` and is part of `go.sum` because the package `github.com/mattermost/mattermost-server/v6/` includes it in it's `go.mod`. But it does not contribute to the bundle size as it's not an import.

If one wants to dig deeper: `nm` can be used to get the list of symbols from an object file, e.g. `nm server/dist/plugin-linux-arm64`. That list contains all the symbols that contribute to the bundle size. If we want to improve the bundle size, we need to find imports that could be avoided if packaged in a different way. `github.com/mattermost/mattermost-server/v6/model` imports `github.com/mattermost/mattermost-server/v6/shared/markdown` only to use it in `RewriteImageURLs`. Maybe `RewriteImageURLs` could be moved in `github.com/mattermost/mattermost-server/v6/shared/markdown` and the import could be avoided.

But these are small optimisation. In general, the compiler is smarter than the programmer :wink:

server/plugin/command.go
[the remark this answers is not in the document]
hanzei: @mickmister What do you think about my comment?

mickmister (approved):

go.mod
```diff
@@ -1,15 +1,93 @@
...
 require (
 	github.com/Masterminds/sprig/v3 v3.2.2
 	github.com/google/go-github/v41 v41.0.0
 	github.com/gorilla/mux v1.8.0
-	github.com/mattermost/mattermost-plugin-api v0.0.27
-	github.com/mattermost/mattermost-server/v6 v6.5.2
-	github.com/microcosm-cc/bluemonday v1.0.18
+	github.com/mattermost/mattermost-plugin-api v0.1.1
+	// mmgoget: github.com/mattermost/mattermost-server/v6@v7.5.0 is replaced by -> github.com/mattermost/mattermost-server/v6@21aec2741b
+	github.com/mattermost/mattermost-server/v6 v6.0.0-20221109191448-21aec2741bfe
```
srkgupta, line 11: Is this intentional, i.e. downgrading the from `v6.5.2` to `v6.0.0-x`?

PR #625, opened 2022-12-20
Use for CI reusable workflows for DRY (#350)

spirosoik:
#### Summary
We can use across all the repositories of plugins and apps the same reusable workflow which give us the ability to change only one file instead of multiple and apart from this the ability to use different versions as we prefer.

#### Ticket Link
Ticket: https://mattermost.atlassian.net/browse/CLD-4681


toninis:
@spirosoik Are we going to inform users for rebase or Circle CI will just phase out when all PRs are merged ? It would be good to globally test this across all PRs .

spirosoik:
@toninis we are going to update the PRs as we use / commands in github comments .eg /update-branch

toninis (approved):

.github/workflows/cd.yml
```diff
@@ -0,0 +1,17 @@
+name: cd
+on:
+  workflow_run:
+    workflows: ["ci"]
+    types:
+      - completed
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
```
hanzei, line 9: What about RC releases like https://github.com/mattermost/mattermost-plugin-github/releases/tag/v2.1.3-rc2? Shouldn't CI/CD also trigger for them?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,16 @@
+name: ci
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+  pull_request:
```
hanzei, line 8: Why is CI limited to only PRs? Given that it takes some time to run, it's great if CI is already "don" when a PR gets submitted so any fixes can be addressed quickly.

Also: Pushes to plugin repos happen rarely. I don't think there is a cost concern.

spirosoik: Do you think should run on branch? If we send multiple commits during the day in practice we consume lots of miniutes.

.github/workflows/cd.yml
spirosoik: This is a great point we can definitely evolve this.

hanzei: Is there a ticket for fix this issue?

.github/workflows/ci.yml
hanzei: I would argue that it's worth the money. Once a PR is submitted, all new commits will trigger CI anyway. This only cuts the costs for branches that are still in development. Having direct feedback from CI is helpful.

PR #664, opened 2023-04-04
[GH-378]:Fixed issue #378 'Check for webhook while subscribing'

Kshitij-Katiyar:
**Summary** 

- When a user creates a subscription to a repo the plugin is now checking the repo (or parent Organization for Enterprise) to see if a webhook is in place to deliver the events.

- If no webhook is found on the GitHub side, an ephemeral post is sent to remind the user that a webhook must be set up..

- [Issue #378](https://github.com/mattermost/mattermost-plugin-github/issues/378)


m1lt0n (approved):
Tested locally and the check & message changes seem to work properly

server/plugin/command.go
```diff
@@ -25,6 +26,11 @@ const (
 	featureStars         = "stars"
 )
 
+const (
+	ErrorNoWebhookFound           = "\nNo webhook was found for this repository or organization. To create one, enter the following slash command `/github setup webhook`"
+	GithubListOptionsPerPageValue = 50
```
hanzei, line 31: Do we have a `const` for `perPage` anywhere else?

server/plugin/command.go
```diff
@@ -25,6 +26,11 @@ const (
 	featureStars         = "stars"
 )
 
+const (
+	ErrorNoWebhookFound           = "\nNo webhook was found for this repository or organization. To create one, enter the following slash command `/github setup webhook`"
```
hanzei, line 30: 2/5 that this is not a global message. It's only used in `handleSubscribesAdd` and can be defined there.

server/plugin/command.go
```diff
@@ -265,6 +271,50 @@ func (p *Plugin) handleSubscriptionsList(_ *plugin.Context, args *model.CommandA
...
+		if repo == "" {
+			githubHooks, githubResponse, err = githubClient.Organizations.ListHooks(ctx, owner, opt)
+		} else {
+			githubHooks, githubResponse, err = githubClient.Repositories.ListHooks(ctx, owner, repo, opt)
+		}
+
+		if err != nil {
+			p.API.LogWarn("Not able to get the list of webhooks", "Owner", owner, "Repo", repo, "Error", err.Error())
+			// Breaking from the loop if the repo or org is not found
+			if strings.Contains(err.Error(), "404 Not Found") {
```
hanzei, line 294: Should the code check for the returned HTTP status instead?

server/plugin/command.go
```diff
@@ -265,6 +271,50 @@ func (p *Plugin) handleSubscriptionsList(_ *plugin.Context, args *model.CommandA
...
+				isWebhook = true
+				break
```
hanzei, line 296: Breaking here de-facto means returning, right? Why is the 404 case special?

server/plugin/command.go
```diff
@@ -265,6 +271,50 @@ func (p *Plugin) handleSubscriptionsList(_ *plugin.Context, args *model.CommandA
 	return txt
 }
 
+func (p *Plugin) getWebhookListForRepoOrOrg(githubClient *github.Client, repo, owner string, ctx context.Context) (bool, error) {
+	isWebhook := false
```
hanzei, line 275: `isWebhook` does seem intention revealing to me. Maybe
```suggestion
	var found bool
```

server/plugin/plugin.go
```diff
@@ -990,3 +990,10 @@ func (p *Plugin) getUsername(mmUserID string) (string, error) {
 
 	return "@" + info.GitHubUsername, nil
 }
+
+func (p *Plugin) getSiteURL() string {
+	if p.API.GetConfig().ServiceSettings.SiteURL != nil {
+		return *p.API.GetConfig().ServiceSettings.SiteURL
+	}
+	return ""
+}
```
hanzei, line 999: Would you mind updating the rest of the code base, like in https://github.com/mattermost/mattermost-plugin-github/blob/c7124749f2af0a3c5060478ae26ce12b8bddabc4/server/plugin/api.go#L845?

hanzei (changes requested):

server/plugin/command.go
Kshitij-Katiyar: no there is no constant for this in the code

server/plugin/command.go
Kshitij-Katiyar: 404 case is special because we are checking whether there is an org or repo present with the name we put during subscribing if not we don't need any additional messages.

server/plugin/command.go
hanzei: 1/5 let's use a generic `perPage` here and in the rest of the code.

server/plugin/command.go
```diff
@@ -265,7 +270,53 @@ func (p *Plugin) handleSubscriptionsList(_ *plugin.Context, args *model.CommandA
...
+		} else {
+			githubHooks, githubResponse, err = githubClient.Repositories.ListHooks(ctx, owner, repo, opt)
+		}
+
+		if err != nil {
+			p.API.LogWarn("Not able to get the list of webhooks", "Owner", owner, "Repo", repo, "Error", err.Error())
+			// Breaking from the loop if the repo or org is not found
+			if githubResponse.StatusCode == 404 && strings.Contains(err.Error(), "404 Not Found") {
+				found = true
+				break
```
hanzei, line 296: I still don't get why the code returns a `true` here. Should it return  `false, nil` here?

hanzei (changes requested):

hanzei:
@Kshitij-Katiyar Would you please merge `master` into your PR?

server/plugin/command.go
hanzei: Can you help me understand this code?

Kshitij-Katiyar: In case we are subscribing to a username then we will get the 404 error. I have handled that case that we are not going to show any message for this particular.

mickmister (approved):

server/plugin/command.go
Kshitij-Katiyar: yes @hanzei. It is just my suggestion but i think we should just log this warning in the server logs rather than showing this warning to the user. Please give your views on this

hanzei: I'm 2/5 that we should show a warning to the user as they expect the webhooks to work. For orgs and their repos the check just works. It's surprising for a user if this special case doesn't properly check.

Kshitij-Katiyar: sure @hanzei will make this change and add the warnings as needed

hanzei:
@Kshitij-Katiyar There is no need to re-request a review. That can be done once there is new code to review.

DHaussermann:
@Kshitij-Katiyar I did not have time to pull this into my work queue. 

I'll keep myself asa reviewer for now. Please advise if this can be covered by a QA resource on the Brightscout side.

AayushChaudhary0001:
The PR has been tested and the actual result can be enhanced as the ephemeral post is getting merged with the warning message for the private repository.

**Enhancement** : A separate title for the ephemeral post can be added.

image

mickmister:
@matthewbirtch Do you have any thoughts on the screenshot and comment mentioned above?

mickmister:
@AayushChaudhary0001 I see, the "No webhook was found" is blended into the same paragraph as the "private repository" warning. Maybe we should put the webhook message above, or have a "**Note:**" before the webhook message

matthewbirtch:
> @AayushChaudhary0001 I see, the "No webhook was found" is blended into the same paragraph as the "private repository" warning. Maybe we should put the webhook message above, or have a "**Note:**" before the webhook message

I agree that we shouldn't blend this note in with the warning paragraph.

I would also say that it doesn't quite feel right that we say "Successfully subscribed to {repo}", and then have a warning and a note about no webhook being found. 

Maybe we consider adding the note to the first paragraph so it reads:

 "Successfully subscribed to {repo}, but no webhook was found for this repository or organization. To create one, enter `/github setup webhook`"

PR #626, opened 2022-12-30
[GH-613]:Fixed issue "API rate limit exceeded for user ID".

Kshitij-Katiyar:
**Summary** 

- User's mattermost logs were full of API rate limit exceeded warn message due to many API calls simultaneously. Mainly due to the notifications count on the mattermost sidebar.
- Please refer to this comment for further information:- https://github.com/mattermost/mattermost-plugin-github/issues/613#issuecomment-1357487107

- [Issue #613](https://github.com/mattermost/mattermost-plugin-github/issues/613)

PR #675, opened 2023-05-05
[GH-669]: Add org selector to create issue modal

thoratvinod:
#### Summary
1. Added endpoints for listing organizations and listing repositories for specific organizations.
2. Added an additional selector for selecting organizations, and made related changes in actions & reducers.

#### Screenshots
If organizations are present:
image

If no organization is present: 
image

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/669


thoratvinod:
image

webapp/src/reducers/index.js
```diff
@@ -104,6 +104,24 @@ function yourPrs(state = [], action) {
     }
 }
 
+function yourOrgs(state = [], action) {
+    switch (action.type) {
+    case ActionTypes.RECEIVED_ORGANIZATIONs:
```
mickmister, line 109:
```suggestion
    case ActionTypes.RECEIVED_ORGANIZATIONS:
```

webapp/src/client/client.js
```diff
@@ -43,10 +43,18 @@ export default class Client {
...
+    getOrganizations = async () => {
+        return this.doGet(`${this.url}/organizations`);
+    }
+
     getRepositories = async () => {
         return this.doGet(`${this.url}/repositories`);
     }
 
+    getRepositoriesByOrganization = async (organization) => {
+        return this.doGet(`${this.url}/reposByOrg?organization=${organization}`);
```
mickmister, line 55:
```suggestion
        return this.doGet(`${this.url}/repos_by_org?organization=${organization}`);
```

webapp/src/components/github_repo_selector/github_repo_selector.jsx
```diff
@@ -30,25 +33,65 @@ export default class GithubRepoSelector extends PureComponent {
...
     componentDidMount() {
-        this.props.actions.getRepos();
+        this.props.actions.getOrgs();
+        this.props.actions.getReposByOrg("");
     }
 
-    onChange = (_, name) => {
-        const repo = this.props.yourRepos.find((r) => r.full_name === name);
+    onChangeForOrg = (_, org) => {
+        if (this.state.org != org) {
```
mickmister, line 41:
```suggestion
        if (this.state.org !== org) {
```

webapp/src/components/github_repo_selector/github_repo_selector.jsx
```diff
@@ -30,25 +33,65 @@ export default class GithubRepoSelector extends PureComponent {
...
+            this.setState({org: org}, () => {this.props.actions.getReposByOrg(org)});
```
mickmister, line 42: Does this code actually need to wait for `setState` to complete?

```js
this.setState({org});
this.props.actions.getReposByOrg(org)
```

webapp/src/components/github_repo_selector/github_repo_selector.jsx
mickmister, line 41:
```suggestion
        if (this.state.org !== org) {
```

webapp/src/components/github_repo_selector/github_repo_selector.jsx
```diff
@@ -30,25 +33,65 @@ export default class GithubRepoSelector extends PureComponent {
...
+
+    onChangeForRepo = (_, name) => {
+        const repo = this.props.yourReposByOrg.find((r) => r.full_name === name);
         this.props.onChange({name, permissions: repo.permissions});
     }
 
     render() {
-        const repoOptions = this.props.yourRepos.map((item) => ({value: item.full_name, label: item.full_name}));
+        const orgOptions  = this.props.yourOrgs.map((item) => ({value: item.login, label: item.login}));
+        orgOptions.unshift({value: "", label: "your owned repositories"});
```
mickmister, line 54: Maybe instead just have the user's login as the first option in the "org" selector, and have it automatically selected

webapp/src/components/github_repo_selector/github_repo_selector.jsx
```diff
@@ -30,25 +33,65 @@ export default class GithubRepoSelector extends PureComponent {
...
+        const repoOptions = this.props.yourReposByOrg.map((item) => ({value: item.full_name, label: item.name}));
+        let orgSelector, helperTextForRepoSelector;
```
mickmister, line 56:
```suggestion
        let orgSelector = null;
        let helperTextForRepoSelector = 'Returns GitHub repositories connected to the user account';
```

Then we can remove the `else` block below

server/plugin/api.go
```diff
@@ -1196,10 +1198,29 @@ func (p *Plugin) getMilestones(c *UserContext, w http.ResponseWriter, r *http.Re
 	p.writeJSON(w, allMilestones)
 }
 
-func getRepositoryList(c context.Context, userName string, githubClient *github.Client, opt github.ListOptions) ([]*github.Repository, error) {
+func getOrganizationList(c context.Context, userName string, githubClient *github.Client, opt github.ListOptions) ([]*github.Organization, error) {
+	var allOrgs []*github.Organization
```
mickmister, line 1202: This API endpoint needs to take into account the `GitHubOrg` plugin config setting. If this setting is set, we should just return that one org in the setting's value.

https://github.com/mattermost/mattermost-plugin-github/blob/a45515e2fed91f6a3c040e05c002c6666c1a4f7d/server/plugin/configuration.go#L28

mickmister (changes requested):
Thanks for this great contribution @thoratvinod! I have some comments regarding the `GitHubOrg` config setting, and some other suggestions. Overall looking good :+1:

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: Yes, it will be automatically selected because the option's value is an empty string, and our org state is set to empty by default. The 'Users Login' option should be the first option because, for instance, if a user selects 'org1' initially and later wants to select 'Your Owned Repositories,' they should have the flexibility to choose that option.

please let me know if I am missing something.

webapp/src/reducers/index.js
thoratvinod: Done.

webapp/src/client/client.js
thoratvinod: Done.

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: Not necessary, changed.

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: Done.

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: Done.

server/plugin/api.go
thoratvinod: Done.

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: @mickmister any comment on this ?

mickmister: I was thinking that we would replace the label of this option `your owned repositories`, with the user's login instead. Same general functionality as current, but show the label as e.g `thoratvinod`

webapp/src/components/github_repo_selector/github_repo_selector.jsx
```diff
@@ -30,25 +33,65 @@ export default class GithubRepoSelector extends PureComponent {
...
 
     render() {
-        const repoOptions = this.props.yourRepos.map((item) => ({value: item.full_name, label: item.full_name}));
+        const orgOptions  = this.props.yourOrgs.map((item) => ({value: item.login, label: item.login}));
+        orgOptions.unshift({value: "", label: "your owned repositories"});
+        const repoOptions = this.props.yourReposByOrg.map((item) => ({value: item.full_name, label: item.name}));
+        let orgSelector = null;
+        let helperTextForRepoSelector = 'Returns GitHub repositories connected to the user account';
+        // If there are no organanizations for authenticated user, then don't show organization selector
+        if (orgOptions.length > 1   ) {
```
mickmister, line 60: I'm thinking this can be cleaned up a bit by adding some blank lines

```suggestion
        const orgOptions  = this.props.yourOrgs.map((item) => ({value: item.login, label: item.login}));
        orgOptions.unshift({value: "", label: "your owned repositories"});

        const repoOptions = this.props.yourReposByOrg.map((item) => ({value: item.full_name, label: item.name}));

        let orgSelector = null;
        let helperTextForRepoSelector = 'Returns GitHub repositories connected to the user account';

        // If there are no organanizations for authenticated user, then don't show organization selector
        if (orgOptions.length > 1) {
```

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: Done.

server/plugin/api.go
```diff
@@ -1196,10 +1198,29 @@ func (p *Plugin) getMilestones(c *UserContext, w http.ResponseWriter, r *http.Re
...
+
+		allOrgs = append(allOrgs, orgs...)
+		if resp.NextPage == 0 {
+			break
+		}
+
+		opt.Page = resp.NextPage
+	}
+
+	return allOrgs, nil
```
mickmister, line 1217: Will this return specifically orgs that the user is associated with in some way? Or is a member of the org? I want to make sure this doesn't blow up and return 200+ orgs or something like that

server/plugin/api.go
```diff
@@ -1233,6 +1254,95 @@ func getRepositoryListByOrg(c context.Context, org string, githubClient *github.
...
+			if statusCode == http.StatusNotFound {
+				allRepos, err = getRepositoryList(c.Ctx, org, githubClient, github.RepositoryListOptions{ListOptions: opt})
+				if err != nil {
+					c.Log.WithError(err).Warnf("Failed to list repositories")
+					p.writeAPIError(w, &APIErrorResponse{Message: "Failed to fetch repositories", StatusCode: http.StatusInternalServerError})
+					return
+				}
+			} else {
+				c.Log.WithError(err).Warnf("Failed to list repositories")
+				p.writeAPIError(w, &APIErrorResponse{Message: "Failed to fetch repositories", StatusCode: http.StatusInternalServerError})
```
mickmister, line 1324:
```suggestion
				p.writeAPIError(w, &APIErrorResponse{Message: "Failed to fetch repositories", StatusCode: statusCode})
```

mickmister (changes requested):
Thanks for this contribution @thoratvinod :+1: I have one more request to avoid returning the current user's login in the case where we have the `GitHubOrg` config setting set. Otherwise LGTM

server/plugin/api.go
```diff
@@ -1233,6 +1254,95 @@ func getRepositoryListByOrg(c context.Context, org string, githubClient *github.
...
 
+func (p *Plugin) getOrganizations(c *UserContext, w http.ResponseWriter, r *http.Request) {
+
+	var allOrgs []*github.Organization
+	var err error
+
+	includeLoggedInUser := r.URL.Query().Get("includeLoggedInUser")
+	if includeLoggedInUser == "true" {
+		allOrgs = append(allOrgs, &github.Organization{Login: &c.GHInfo.GitHubUsername})
+	}
```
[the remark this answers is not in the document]
thoratvinod: Done.

webapp/src/components/github_repo_selector/github_repo_selector.jsx
```diff
@@ -30,25 +33,66 @@ export default class GithubRepoSelector extends PureComponent {
...
+    }
+
+    onChangeForRepo = (_, name) => {
+        const repo = this.props.yourReposByOrg.find((r) => r.full_name === name);
         this.props.onChange({name, permissions: repo.permissions});
     }
 
     render() {
-        const repoOptions = this.props.yourRepos.map((item) => ({value: item.full_name, label: item.full_name}));
+        const orgOptions = this.props.yourOrgs.map((item, i) => (i === 0 ? {value: '', label: item.login} : {value: item.login, label: item.login}));
```
thoratvinod, line 54: Handle UI after backend changes

server/plugin/api.go
mickmister: Yes that sounds good to me :+1:

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: Done.

webapp/src/components/github_repo_selector/github_repo_selector.jsx
```diff
@@ -30,25 +33,69 @@ export default class GithubRepoSelector extends PureComponent {
     }
 
     componentDidMount() {
-        this.props.actions.getRepos();
+        this.props.actions.getOrgs(true);
```
mickmister, line 36: Sorry I thought I commented this before but I guess not. Do we need this `true` passed in here? It's a bit cryptic, and I believe this is the only spot we call this function. Thoughts on getting rid of the parameter?

webapp/src/components/github_repo_selector/github_repo_selector.jsx
```diff
@@ -51,7 +54,7 @@ export default class GithubRepoSelector extends PureComponent {
     }
 
     render() {
-        const orgOptions = this.props.yourOrgs.map((item, i) => (i === 0 ? {value: '', label: item.login} : {value: item.login, label: item.login}));
+        const orgOptions = this.props.yourOrgs.map((item) => (item.isLoggedInUser ? {value: '', label: item.login} : {value: item.login, label: item.login}));
```
mickmister, line 57: What's the reasoning behind having the `value` be blank here?

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: Yes, this is only called here. The `true` parameter is for `includeLoggedInUser`. If we pass this flag as true, then only the server will send a `username` as part of the organization's list. I thought the `getOrgs` action would be reusable in the future, so it would be better to have this flag as a parameter. What are your thoughts on this?

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: Because this `value` is passed to the `getReposByOrg` action, if the server-side identifies an empty `organization` query parameter and no organization is specified in the config settings, it will send repositories that are owned by the logged-in user

webapp/src/components/github_repo_selector/github_repo_selector.jsx
```diff
@@ -30,25 +33,69 @@ export default class GithubRepoSelector extends PureComponent {
     }
 
     componentDidMount() {
-        this.props.actions.getRepos();
+        this.props.actions.getOrgs(true);
+        if (this.props.yourOrgs.length > 0) {
+            const firstOrg = this.props.yourOrgs[0];
+            this.props.actions.getReposByOrg(firstOrg.isLoggedInUser ? '' : firstOrg.login);
```
thoratvinod, line 39: handle race condition, if org from cofig settings provided

webapp/src/components/github_repo_selector/github_repo_selector.jsx
mickmister: @thoratvinod Not sure about this. Possibly [yagni](https://martinfowler.com/bliki/Yagni.html), and the cryptic nature of it makes it a bit harder to read. I'm inclined to remove the flag. I'm not sure if we ever _don't_ want that when the org config is not set.

webapp/src/components/github_repo_selector/github_repo_selector.jsx
mickmister: Why not just use the logged in user's username instead of blank string though? Just a bit harder to read, since you had to explain it to me, and still seems a bit confusing.

webapp/src/components/github_repo_selector/github_repo_selector.jsx
mickmister: What is the race condition to handle?

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: Yes, it actually makes sense to pass the username directly as it would avoid unnecessary handling at the front end. Initially, the thought was to avoid calling the endpoint like this: `repos_by_org?organization=<username>`.
will do the chages

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: It was not exactly a race condition; rather, it was a bug. Let's consider a scenario where a GitHub organization is configured in the config.json file. In this case, only one organization will be returned. Within the component, if the length of the organizations is equal to 1, we don't show the react-select for organizations. The expectation is that all the repositories under that organization should have already been fetched for the repository react-select.

However, if we look closely, we observe that the organizations are fetched in the `componentDidMount` method, and there is no opportunity to fetch repositories for the configured organization elsewhere. Additionally, since we are not displaying the organization react-select, the user cannot choose an organization.

Solution: Modified the code by adding an `if `condition in the `componentDidUpdate` method. If `prevProps.yourOrgs` is not equal to `currProps.yourOrgs`, automatically set the first organization and fetch the repositories for that organization.

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: Done.

webapp/src/components/github_repo_selector/github_repo_selector.jsx
thoratvinod: Done.

webapp/src/components/github_repo_selector/github_repo_selector.jsx
mickmister: We could alternatively provide the `GithubOrg` config setting to the client so it has some more context on what should be shown

mickmister (approved):

mickmister:
@thoratvinod Thank you for your patience on this

hanzei:
@calebroseland @mgdelacroix Gentle reminder to review this PR

hanzei:
@thoratvinod Do you want to continue working on this issue?

thoratvinod:
@hanzei yes, I will continue this, will take 2-3 weeks to get back on this. Is that fine?

hanzei:
Sure, we are not in a hurry. Thanks for coming back to the PR @thoratvinod :+1:

hanzei:
Thanks for updating the PR @thoratvinod :+1: 

CI complains about two errors:
```
/home/runner/work/mattermost-plugin-github/mattermost-plugin-github/webapp/src/reducers/index.js
Error:   243:5  error  'yourPrsDetails' is not defined   no-undef
Error:   244:5  error  'yourAssignments' is not defined  no-undef
```

I suppose the changes in https://github.com/mattermost/mattermost-plugin-github/pull/626 caused actions to not longer exist.

thoratvinod:
Hey @hanzei , @mickmister ,
I've got some personal stuff going on and won't be able to keep up with this issue. Thanks for understanding

mickmister:
Thanks for your work on this @thoratvinod! I'm closing for now and mentioned this PR on the linked ticket

PR #697, opened 2023-09-13
[GH-692] Fix formatting inside code block in issue created notification

### LINKED ISSUE #692: Code in Pull Request code blocks is broken on Mattermost
For example, I just opened this Pull Request: https://github.com/indico/indico/issues/5905
The code block is just a standard triple-backtick-fenced block.

But in the Mattermost notification it became this:

```python
subcontribution_settings = EventSettingsProxy(&#39;subcontributions&#39;, {
    &#39;speakers_can_submit&#39;: False,  # Speakers can edit minutes and upload attachments
})
```

As you can see, the quotes got replaced with HTML entities.

raghavaggarwal2308:
#### Summary
Fixed the improper formatting of special characters inside the code block of issue-created channel notifications.

###### What to test?
- Special characters are properly displayed inside the code block present in the description of the issue-created notification.

#### Screenshots
Earlier:
image

Now:
image

#### Ticket Link
Fixes #692

### DIFF
--- a/server/plugin/webhook.go
+++ b/server/plugin/webhook.go
@@ -460,8 +460,7 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
 func (p *Plugin) sanitizeDescription(description string) string {
 	var policy = bluemonday.StrictPolicy()
 	policy.SkipElementsContent("details")
-	result := policy.Sanitize(description)
-	html.UnescapeString(result)
+	result := html.UnescapeString(policy.Sanitize(description))
 	return strings.TrimSpace(result)
 }
 
--- a/server/plugin/webhook.go
+++ b/server/plugin/webhook.go
@@ -6,6 +6,7 @@ import (
 	"crypto/sha1" //nolint:gosec // GitHub webhooks are signed using sha1 https://developer.github.com/webhooks/.
 	"encoding/hex"
 	"encoding/json"
+	"html"
 	"io"
 	"net/http"
 	"strings"
@@ -459,12 +460,8 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
 func (p *Plugin) sanitizeDescription(description string) string {
 	var policy = bluemonday.StrictPolicy()
 	policy.SkipElementsContent("details")
-	// Replacing the unicodes with the respective special characters to have proper rendering in code block.
-	result := strings.ReplaceAll(policy.Sanitize(description), "&#39;", "'")
-	result = strings.ReplaceAll(result, "&#34;", "\"")
-	result = strings.ReplaceAll(result, "&amp;", "&")
-	result = strings.ReplaceAll(result, "&gt;", ">")
-	result = strings.ReplaceAll(result, "&lt;", "<")
+	result := policy.Sanitize(description)
+	html.UnescapeString(result)
 	return strings.TrimSpace(result)
 }
 
--- a/server/plugin/webhook.go
+++ b/server/plugin/webhook.go
@@ -459,7 +459,11 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
 func (p *Plugin) sanitizeDescription(description string) string {
 	var policy = bluemonday.StrictPolicy()
 	policy.SkipElementsContent("details")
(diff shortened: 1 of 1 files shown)


hanzei (approved):

server/plugin/webhook.go
```diff
@@ -459,7 +459,13 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
...
 	var policy = bluemonday.StrictPolicy()
 	policy.SkipElementsContent("details")
-	return strings.TrimSpace(policy.Sanitize(description))
+	// Replacing the unicodes with the respective special characters to have proper rendering in code block.
+	result := strings.ReplaceAll(policy.Sanitize(description), "&#39;", "'")
+	result = strings.ReplaceAll(result, "&#34;", "\"")
+	result = strings.ReplaceAll(result, "&amp;", "&")
+	result = strings.ReplaceAll(result, "&gt;", ">")
+	result = strings.ReplaceAll(result, "&lt;", "<")
+	return strings.TrimSpace(result)
```
mickmister, line 468: I wonder if there is a more catch-all holistic approach

mickmister (approved):

server/plugin/webhook.go
raghavaggarwal2308: @mickmister I explored other packages to sanitize the HTML but the one we are using right now was the best I could find. But this converts the special characters to their unicode value which are not formatted inside a code block, so we are replacing them again here. Please feel free to share your opinions if you have a better approach in mind.

mickmister: Hmm so it's the `policy.Sanitize` that's causing the issue. We really just want to use it to remove the `<details>` elements from the description, and not escape those characters.

As this comment https://github.com/microcosm-cc/bluemonday/issues/39#issuecomment-454358002 mentions on a related issue for the `bluemonday` package, we can call `html.UnescapeString` from the result to make those characters go back to what we want them to be

raghavaggarwal2308: @mickmister Updated

server/plugin/webhook.go
```diff
@@ -459,12 +460,8 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
...
 	var policy = bluemonday.StrictPolicy()
 	policy.SkipElementsContent("details")
-	// Replacing the unicodes with the respective special characters to have proper rendering in code block.
-	result := strings.ReplaceAll(policy.Sanitize(description), "&#39;", "'")
-	result = strings.ReplaceAll(result, "&#34;", "\"")
-	result = strings.ReplaceAll(result, "&amp;", "&")
-	result = strings.ReplaceAll(result, "&gt;", ">")
-	result = strings.ReplaceAll(result, "&lt;", "<")
+	result := policy.Sanitize(description)
+	html.UnescapeString(result)
```
mickmister, line 464: I think we need to reassign the value right?

```suggestion
	result = html.UnescapeString(result)
```

mickmister:
Does the `bluemonday` policy remove things other than the `<details>` tags? I'm thinking it may remove things like `<script>` because the tool is meant to sanitize the HTML. We would not want it to remove anything but the `<details>` tags in our case.

raghavaggarwal2308:
@mickmister Yes, it's removing the HTML tags in the string. For some tags, it's just removing the tags and for some, it's removing the tags along with the text inside.
issue_notification

mickmister:
@raghavaggarwal2308 Since the main thing we want to sanitize is `<details>`, we could just only run the sanitize when there exists a `<details>` tag in the post body. It's not common to have those in the post body, but when it shows up it's usually spammy which is why we remove them

raghavaggarwal2308:
> @raghavaggarwal2308 Since the main thing we want to sanitize is `<details>`, we could just only run the sanitize when there exists a `<details>` tag in the post body. It's not common to have those in the post body, but when it shows up it's usually spammy which is why we remove them

@mickmister If we do so and there are other tags along with the details tag, then they will be removed as well. Is that fine?

mickmister:
@raghavaggarwal2308 Yes that's fine :+1:

raghavaggarwal2308:
@mickmister Added logic to sanitize description conditionally.

mickmister (approved):

PR #721, opened 2023-12-13
[MM-17] Fix issues in release 2.2.0

raghavaggarwal2308:
#### Summary
1. Mention notifications not working.
2. Labels and milestones not displayed in the RHS.

#### What to test?
- Labels and milestones are properly displayed in the RHS.
- Getting DM notifications for mentions.

#### Steps to reproduce:
- Labels and milestones:
1. Connect your GitHub account.
2. Create a PR on Github and add labels/milestones to it.
3. Open RHS for open PRs on Mattermost.

- DM notification:
1. Connect your GitHub account.
2. Create a webhook.
3. You should get notified in GitHub bot DM when someone else mentions/assigns you in a PR/Issue.

#### Ticket Link
N/A

### DIFF
--- a/server/plugin/graphql/lhs_request.go
+++ b/server/plugin/graphql/lhs_request.go
@@ -64,7 +64,7 @@ func (c *Client) GetLHSData(ctx context.Context) ([]*github.Issue, []*github.Iss
 		if !allAssignmentsFetched {
 			for i := range mainQuery.Assignments.Nodes {
 				resp := mainQuery.Assignments.Nodes[i]
-				issue := getIssue(&resp)
+				issue := newIssueFromAssignmentResponse(&resp)
 				resultAssignee = append(resultAssignee, issue)
 			}
 
@@ -97,14 +97,14 @@ func getPR(prResp *prSearchNodes) *github.Issue {
 	resp := prResp.PullRequest
 	labels := getGithubLabels(resp.Labels.Nodes)
 
-	return getGithubIssue(resp.Number, resp.Title, resp.Author.Login, resp.Repository.URL, resp.URL, resp.CreatedAt, resp.UpdatedAt, labels, resp.Milestone.Title)
+	return newGithubIssue(resp.Number, resp.Title, resp.Author.Login, resp.Repository.URL, resp.URL, resp.CreatedAt, resp.UpdatedAt, labels, resp.Milestone.Title)
 }
 
-func getIssue(assignmentResp *assignmentSearchNodes) *github.Issue {
+func newIssueFromAssignmentResponse(assignmentResp *assignmentSearchNodes) *github.Issue {
 	resp := assignmentResp.PullRequest
 	labels := getGithubLabels(resp.Labels.Nodes)
 
-	return getGithubIssue(resp.Number, resp.Title, resp.Author.Login, resp.Repository.URL, resp.URL, resp.CreatedAt, resp.UpdatedAt, labels, resp.Milestone.Title)
+	return newGithubIssue(resp.Number, resp.Title, resp.Author.Login, resp.Repository.URL, resp.URL, resp.CreatedAt, resp.UpdatedAt, labels, resp.Milestone.Title)
 }
 
 func getGithubLabels(labels []labelNode) []*github.Label {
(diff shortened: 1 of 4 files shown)


server/plugin/plugin.go
```diff
@@ -642,7 +642,7 @@ func (p *Plugin) storeGitHubToUserIDMapping(githubUsername, userID string) error
 
 func (p *Plugin) getGitHubToUserIDMapping(githubUsername string) string {
 	var data []byte
-	_ = p.client.KV.Get(githubUsername+githubUsernameKey, data)
+	_ = p.client.KV.Get(githubUsername+githubUsernameKey, &data)
```
hanzei, line 645: Good catch. Let's return the error and log it in `getGitHubToUsernameMapping` so this doesn't go unnoticed another time.

webapp/src/components/sidebar_right/github_items.tsx
```diff
@@ -372,7 +372,7 @@ function getGithubLabels(labels: Label[]) {
     return labels.map((label) => {
         return (
             <Badge
-                key={label.id}
+                key={label.name}
```
hanzei, line 375: Why is there no `id` any longer?

hanzei (changes requested):
Nice work, thanks @raghavaggarwal2308

webapp/src/components/sidebar_right/github_items.tsx
raghavaggarwal2308: @hanzei We replaced the API calls to Github with GraphQL queries for fetching the data. The `id` returned from the GraphQL is an encrypted string. If we have to use that ID we will have to create a custom type on the server side for sending the data to frontend OR we need to explore how we can decrypt that id (Currently we are using the type from the GitHub package which has the id field of type integer). Since id is unnecessary in our use case, we used the `name ` field instead.

hanzei: Are label names unique in GitHub? i.e. do they make for a good `key`?

raghavaggarwal2308: @hanzei Yes, for a particular repo label name will be unique.
image

server/plugin/plugin.go
hanzei: I'm 1/5 that returning an error is better as it allows the caller to log with more context. Logging in `getGitHubToUserIDMapping` is also fine.

raghavaggarwal2308: > Logging in getGitHubToUserIDMapping is also fine

@hanzei We are going with this.

server/plugin/plugin.go
```diff
@@ -642,7 +642,11 @@ func (p *Plugin) storeGitHubToUserIDMapping(githubUsername, userID string) error
 
 func (p *Plugin) getGitHubToUserIDMapping(githubUsername string) string {
 	var data []byte
-	_ = p.client.KV.Get(githubUsername+githubUsernameKey, data)
+	err := p.client.KV.Get(githubUsername+githubUsernameKey, &data)
+	if err != nil {
+		p.API.LogError("Error occurred while getting the user ID from KV store using the Github username", "Error", err.Error())
```
hanzei, line 647: It's not actionable for an admin.
```suggestion
		p.API.LogWarn("Error occurred while getting the user ID from KV store using the Github username", "Error", err.Error())
```

hanzei (approved):
:rocket:

server/plugin/plugin.go
raghavaggarwal2308: @hanzei Updated

arush-vashishtha (approved):
Tested the PR for the followings,

- Labels and milestones added to a PR or issue are properly displayed in the RHS.
- Getting DM notifications from Github bot for the mentions and assigned PR and issues.

Everything is working fine, Approved.
cc @AayushChaudhary0001

server/plugin/graphql/lhs_request.go
```diff
@@ -95,21 +95,38 @@ func (c *Client) GetLHSData(ctx context.Context) ([]*github.Issue, []*github.Iss
...
+
+	return githubLabels
+}
+
+func getGithubIssue(prNumber githubv4.Int, title, login githubv4.String, repositoryURL, htmlURL githubv4.URI, createdAt, updatedAt githubv4.DateTime, labels []*github.Label, milestone githubv4.String) *github.Issue {
 	number := int(prNumber)
 	repoURL := repositoryURL.String()
 	issuetitle := string(title)
 	userLogin := (string)(login)
+	milestoneTitle := (string)(milestone)
```
mickmister, line 129: I wonder why we have two different syntaxes for making these strings

server/plugin/graphql/lhs_request.go
```diff
@@ -95,21 +95,38 @@ func (c *Client) GetLHSData(ctx context.Context) ([]*github.Issue, []*github.Iss
...
-	return getGithubIssue(resp.Number, resp.Title, resp.Author.Login, resp.Repository.URL, resp.URL, resp.CreatedAt, resp.UpdatedAt)
+	return getGithubIssue(resp.Number, resp.Title, resp.Author.Login, resp.Repository.URL, resp.URL, resp.CreatedAt, resp.UpdatedAt, labels, resp.Milestone.Title)
 }
 
 func getIssue(assignmentResp *assignmentSearchNodes) *github.Issue {
 	resp := assignmentResp.PullRequest
+	labels := getGithubLabels(resp.Labels.Nodes)
 
-	return getGithubIssue(resp.Number, resp.Title, resp.Author.Login, resp.Repository.URL, resp.URL, resp.CreatedAt, resp.UpdatedAt)
+	return getGithubIssue(resp.Number, resp.Title, resp.Author.Login, resp.Repository.URL, resp.URL, resp.CreatedAt, resp.UpdatedAt, labels, resp.Milestone.Title)
```
mickmister, line 107: Then we also have this `getIssue` function which is not clear how it differs from the `getGithubIssue` function on first glance. Maybe we can name this `getIssue` function to `newIssueFromAssignmentResponse`

mickmister (approved):
LGTM, just some comments on naming from the recent LHS request feature

raghavaggarwal2308:
@mickmister Fixed the comments added by you

PR #747, opened 2024-02-21
Added debug logs to check issue in subscription for PR updates

ayusht2810:
#### Summary
Added debug logs to check issue in subscription for PR updates

#### Ticket Link
NA

### DIFF
--- a/server/plugin/webhook.go
+++ b/server/plugin/webhook.go
@@ -441,8 +441,7 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
 			if label != "" && label == eventLabel {
 				pullRequestLabelledMessage, err := renderTemplate("pullRequestLabelled", event)
 				if err != nil {
-					p.logForDocsRepoDebugging(repo.GetFullName(), "Failed to render template", "error", err.Error())
-					p.client.Log.Warn("Failed to render template", "error", err.Error())
+					p.client.Log.Warn("Failed to render template", "repo", repo.GetFullName(), "error", err.Error())
 					return
 				}
 
@@ -459,8 +458,7 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
 			}
 			newPRMessage, err := renderTemplate(prNotificationType, GetEventWithRenderConfig(event, sub))
 			if err != nil {
-				p.logForDocsRepoDebugging(repo.GetFullName(), "Failed to render template", "error", err.Error())
-				p.client.Log.Warn("Failed to render template", "error", err.Error())
+				p.client.Log.Warn("Failed to render template", "repo", repo.GetFullName(), "error", err.Error())
 				return
 			}
 
@@ -470,8 +468,7 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
 		if action == actionReopened {
 			reopenedPRMessage, err := renderTemplate("reopenedPR", event)
 			if err != nil {
-				p.logForDocsRepoDebugging(repo.GetFullName(), "Failed to render template", "error", err.Error())
-				p.client.Log.Warn("Failed to render template", "error", err.Error())
+				p.client.Log.Warn("Failed to render template", "repo", repo.GetFullName(), "error", err.Error())
 				return
 			}
 
@@ -481,8 +478,7 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
 		if action == actionMarkedReadyForReview {
 			markedReadyToReviewPRMessage, err := renderTemplate("markedReadyToReviewPR", GetEventWithRenderConfig(event, sub))
 			if err != nil {
-				p.logForDocsRepoDebugging(repo.GetFullName(), "Failed to render template", "error", err.Error())
-				p.client.Log.Warn("Failed to render template", "error", err.Error())
+				p.client.Log.Warn("Failed to render template", "repo", repo.GetFullName(), "error", err.Error())
 				return
 			}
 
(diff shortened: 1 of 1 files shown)


server/plugin/webhook.go
```diff
@@ -184,6 +184,7 @@ func (wb *WebhookBroker) Close() {
 }
 
 func (p *Plugin) handleWebhook(w http.ResponseWriter, r *http.Request) {
+	p.API.LogDebug("Received webhook event.")
```
mickmister, line 187: To avoid over spamming of the logs, I think we should make sure the repo is https://github.com/mattermost/docs before logging anything. We can have a function `p.logForDocsRepoDebugging(repo, message)` to but this logic there.

I think line 216 is the first point where we have access to which repo the event is for

server/plugin/webhook.go
```diff
@@ -231,6 +232,7 @@ func (p *Plugin) handleWebhook(w http.ResponseWriter, r *http.Request) {
 			p.webhookBroker.publishPing(event, false)
 		}
 	case *github.PullRequestEvent:
+		p.API.LogDebug("Event type PullRequestEvent received.")
```
mickmister, line 235: This should be the only event type we need to log for

server/plugin/webhook.go
```diff
@@ -385,18 +390,22 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
 
 	for _, sub := range subs {
 		if !sub.Pulls() && !sub.PullsMerged() && !sub.PullsCreated() {
```
mickmister, line 392: After the loop we can log whether `foundDocsRepoSub` is `true`

```suggestion
	foundDocsRepoSub := false
	for _, sub := range subs {
		if sub.Repository == "mattermost/docs" {
			foundDocsRepoSub = true
			b, err := json.Marshal(sub)
			if err != nil {
				p.API.LogError("error marshaling subscription", "error", err.Error())
			} else {
				p.API.LogDebug("logging docs subscription, "sub", string(b))
			}
		}

		if !sub.Pulls() && !sub.PullsMerged() && !sub.PullsCreated() {
```

server/plugin/webhook.go
```diff
@@ -385,18 +390,22 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
...
 			continue
 		}
 
 		if sub.PullsCreated() && action != actionOpened {
+			p.API.LogDebug("Rejecting event for current channel as event is for PR opened but the subscription is present for event PR created.", "Channel", sub.ChannelID)
 			continue
 		}
 
 		if p.excludeConfigOrgMember(event.GetSender(), sub) {
+			p.API.LogDebug("Rejecting event for current channel due to exclude config org member condition.", "Channel", sub.ChannelID)
```
mickmister, line 408: Probably want to log at all the return statements below as well, with `p.logForDocsRepoDebugging`

server/plugin/webhook.go
```diff
@@ -543,6 +554,8 @@ func (p *Plugin) handlePRDescriptionMentionNotification(event *github.PullReques
 
 		p.sendRefreshEvent(userID)
 	}
+
+	p.API.LogDebug("Event Completed.", "Event", "handlePRDescriptionMentionNotification")
```
mickmister, line 558: Unsure if we need this logging. We definitely want to make sure we log any errors that occur in the whole process of processing the webhook though

server/plugin/webhook.go
```diff
@@ -919,13 +935,16 @@ func (p *Plugin) postPullRequestReviewEvent(event *github.PullRequestReviewEvent
...
+
+	p.API.LogDebug("Event Completed", "Function", "postPullRequestReviewEvent")
 }
 
 func (p *Plugin) postPullRequestReviewCommentEvent(event *github.PullRequestReviewCommentEvent) {
 	repo := event.GetRepo()
 
 	subs := p.GetSubscribedChannelsForRepository(repo)
 	if len(subs) == 0 {
+		p.API.LogDebug("No subscription found for the event.", "Event", event.GetAction())
```
mickmister, line 947: Shouldn't need to log for comment events. Only PR open/close events

server/plugin/webhook.go
```diff
@@ -350,6 +354,7 @@ func (p *Plugin) postPullRequestEvent(event *github.PullRequestEvent) {
 
 	subs := p.GetSubscribedChannelsForRepository(repo)
```
mickmister, line 355: I think this function `postPullRequestEvent` is where we should focus on

mickmister (commented):
Thanks for looking in to this @ayusht2810. My main comments are related to avoiding over-logging, and making sure we have as much context as possible for the PR open/close events

ayusht2810:
@mickmister review fixes done. Please re-review.

mickmister (approved):

server/plugin/webhook.go
```diff
@@ -1377,3 +1401,9 @@ func (p *Plugin) postStarEvent(event *github.StarEvent) {
 		}
 	}
 }
+
+func (p *Plugin) logForDocsRepoDebugging(repo, message string, kvMessage ...interface{}) {
+	if repo == "mattermost/docs" {
```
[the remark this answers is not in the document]
raghavaggarwal2308: @hanzei Since this is a build just for debugging the issue. I don't think there is a need to add this logic here. Are there any specific risks you are thinking of here?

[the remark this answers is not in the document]
mickmister: If we want to have this permanently in the plugin, I think @hanzei's suggestion makes sense. Note that there could be an issue with configuring the subscription itself though, so this may not help us in all cases. I do think it would help us debug this particular issue though.

Also would we want to support this for all event types?

@hanzei I'm wanting to get something on community as soon as we can to help debug the issue we're experiencing, so I don't want to prolong the PR for the goal of supporting a general use case. Though I see the value in it, and 1/5 if we can get it merged efficiently then I'm fine with this.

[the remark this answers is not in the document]
raghavaggarwal2308: @mickmister @hanzei Maybe we can create a separate Github issue for this and use the current build for debugging?

hanzei (changes requested):
Surely we can use this build for debugging. The cost should just not go into `master`.

mickmister:
@hanzei Do you think the PR is ready to be used for debugging community at its current state?

mickmister:
Fix PR submitted here https://github.com/mattermost/mattermost-plugin-github/pull/755. Closing as this PR shouldn't be needed anymore

PR #761, opened 2024-04-04
[MM-350] Fix issue in create issue modal and attach issue modal

### LINKED ISSUE #557: when user use github user name in ORG lock then at the time of selecting the repo user is not getting the valid repo, user have to refresh to get the valid repo.
Video :-

https://user-images.githubusercontent.com/89907719/161052431-095c0b21-3055-48e0-929e-20e7e515af7f.mov

will add more info if its required.

ayusht2810:
#### Summary

- Fixed issue of repository list not updating when there was incorrect ogranisation in the system console setting for `Create GitHub Issue` modal
- Fixed issue of getting error of `unexpected json` when opening `Attach to GitHub Issue` modal for the above-mentioned case

#### Ticket Link
Fixes #557 

### Checklist
- [x] Completed dev testing
- [x] `make test` Ran test cases and ensured they are passing
- [x] `make check-style` Ran style check and ensured both webapp and server pass the checks

#### What to test?
- Check the dropdown values for repository present in the `Create GitHub Issue` modal
- Check the dropdown values for github issue present in the `Attach to GitHub Issue` modal

#### How to test?
- Add `GitHub Organisation` in the system console settings
- Check the dropdowns mentioned above
- Update the `GitHub Organisation` again in the system console settings by adding some incorrect value
-  Check the dropdown values again


server/plugin/api.go
```diff
@@ -806,6 +806,7 @@ func (p *Plugin) searchIssues(c *UserContext, w http.ResponseWriter, r *http.Req
 	result, _, err := githubClient.Search.Issues(c.Ctx, query, &github.SearchOptions{})
 	if err != nil {
 		c.Log.WithError(err).With(logger.LogContext{"query": query}).Warnf("Failed to search for issues")
+		p.writeJSON(w, make([]*github.Issue, 0))
```
hanzei, line 809: There are a lot of other endpoints that don't return valid JSON in the error case. Should we update the webapp to not expect JSON if the status code indicates an error?

hanzei (changes requested):

server/plugin/api.go
ayusht2810: @hanzei We were thinking of updating the plugin API response for each API where we are not returning anything in case we get any errors from the GitHub API.
For example: At [this](https://github.com/mattermost/mattermost-plugin-github/blob/40626e3a53c8d6e54026a72fbbe298a11247a0aa/server/plugin/api.go#L657) line, we can return an error something like `p.writeAPIError(w, &APIErrorResponse{Message: "Failed to search for issues", StatusCode: http.StatusInternalServerError})` and handle the error similarly at the webapp side.
What are your thoughts on this?

hanzei: If `githubClient.Search.Issues` finds nothing that matches the search, does it return an `error`?

ayusht2810: If we put an invalid organisation or an organisation we don't have access to, we get an error, and it gets rendered as shown in the screenshot below. 
image

But it doesn't return any errors when we search for an issue with a random input term (empty or an organisation we have access to).

raghavaggarwal2308 (approved):
LGTM, upon resolving @hanzei comment.

server/plugin/api.go
mickmister: We def want to avoid showing this `Unexpected end of JSON input` error in all cases

webapp/src/actions/index.js
```diff
@@ -71,6 +71,10 @@ export function getRepos() {
         try {
             data = await Client.getRepositories();
         } catch (error) {
+            dispatch({
+                type: ActionTypes.RECEIVED_REPOSITORIES,
+                data: [],
+            });
```
mickmister, line 77: What should the UI show/do in this case?

ayusht2810: @mickmister Currently, we have the screenshot present here https://github.com/mattermost/mattermost-plugin-github/pull/761#discussion_r1555378571. I think instead of this, we should just show dropdown with no options present, which is currently done in this PR

Should we move forward with the approach mentioned in the [comment](https://github.com/mattermost/mattermost-plugin-github/pull/761#discussion_r1553316249) or should we just return an empty list from the backend to avoid the `Unexpected end of JSON input` error?

server/plugin/api.go
ayusht2810: @hanzei Should we create a new issue to update the structure of the APIs and merge this PR to resolve the issue mentioned in the description?
Also, there were two ways we were thinking of fixing the response of the API:
1. Create a wrapper and update API methods from `(c *UserContext, w http.ResponseWriter, r *http.Request)` to `(c *UserContext, r *http.Request) (any,  int)` as stated by you. We can also include a field of `error` with the data and the status code. In the wrapper, we can check for errors and data. If the error is returned, we can return `p.writeAPIError(w, &APIErrorResponse{Message: "ErrorMessage", StatusCode: statusCode})` and handle similarly on webapp and if there is no error we can return data or an empty list if data is also empty.
2. In this way, we can directly update the API to simply return `p.writeAPIError(w, &APIErrorResponse{Message: "ErrorMessage", StatusCode: statusCode})` in case of error. We can avoid making a wrapper in this case.

What are your thoughts on this? Please let me know if I missed something here.

mickmister: Either one sounds good to me. Which one is less complicated in your opinion @ayusht2810? Have you tried implemented one of them to see if it fits well?

webapp/src/actions/index.js
mickmister: If there is an error from GitHub, we should show a useful error to the user. If we simply fetched no values with no error, then the backend should return an empty array and show no error

server/plugin/api.go
Kshitij-Katiyar: @mickmister We are thinking to proceed with the second method.
we can do this it in this PR or we can create a new PR and merge this so that the [issue](https://github.com/mattermost/mattermost-plugin-github/issues/557) is fixed

webapp/src/actions/index.js
raghavaggarwal2308: @mickmister I have created a [new issue](https://github.com/mattermost/mattermost-plugin-github/issues/794) on Github to return proper json arrays from APIs. Maybe we can proceed further with this PR and handle the error case as part of the new issue. Please let me know your thoughts on this.

mickmister (approved):
LGTM :+1:

There's a follow up ticket to address similar issues in other endpoints https://github.com/mattermost/mattermost-plugin-github/issues/794

AayushChaudhary0001 (approved):

- Tested the dropdown values for repository present in the Create GitHub Issue modal
- Tested the dropdown values for github issue present in the Attach to GitHub Issue modal

The PR was working fine for both the conditions, LGTM. Approved

mickmister:
@wiggin77 This is ready for merge :+1:

hanzei (approved):
I'm fine with moving the discussion from https://github.com/mattermost/mattermost-plugin-github/pull/761#discussion_r1584471664 to a separate endpoint

wiggin77 (approved):

PR #764, opened 2024-04-08
Fixes #498: "

### LINKED ISSUE #498: Validation Message Should be User Friendly
validation is not user friendly

cyrusjc:
Fixes #498

Just a quick client side change which checks if the trimmed length of issueTitle is 0.

We could add another error message that says "Please enter valid title" or something similar to that effect for further clarification but IMO this is enough.

image

### DIFF
--- a/webapp/src/components/modals/create_issue/create_issue.jsx
+++ b/webapp/src/components/modals/create_issue/create_issue.jsx
@@ -61,9 +61,10 @@ export default class CreateIssueModal extends PureComponent {
             e.preventDefault();
         }
 
-        if (!this.validator.validate() || !this.state.issueTitle) {
+        const isValidTitle = this.state.issueTitle.trim().length !== 0;
+        if (!this.validator.validate() || !isValidTitle) {
             this.setState({
-                issueTitleValid: Boolean(this.state.issueTitle),
+                issueTitleValid: Boolean(isValidTitle),
                 showErrors: true,
             });
             return;


webapp/src/components/modals/create_issue/create_issue.jsx
```diff
@@ -61,9 +61,10 @@ export default class CreateIssueModal extends PureComponent {
             e.preventDefault();
         }
 
-        if (!this.validator.validate() || !this.state.issueTitle) {
+        const isValidTitle = this.state.issueTitle.trim().length !== 0;
+        if (!this.validator.validate() || !isValidTitle) {
             this.setState({
-                issueTitleValid: Boolean(this.state.issueTitle),
+                issueTitleValid: Boolean(isValidTitle),
```
Kshitij-Katiyar, line 67: @cyrusjc Thanks for the PR.
Here isValidTitle is already going to be a boolean value. Is there any need to type caste into a Boolean again?

Kshitij-Katiyar (approved):

webapp/src/components/modals/create_issue/create_issue.jsx
cyrusjc: Ah no particular reason other than a small oversight on my part.

mickmister (approved):
Excellent, thank you @cyrusjc! LGTM :+1:

Issue #405, opened 2020-11-13
Pull request review and Pull request review comments not working


ecylmz:
Hi,

Pull request review and Pull request review comments not working for us. We have following error log on mattermost:

`{"level":"warn","ts":1605265921.2438548,"caller":"mlog/sugar.go:27","msg":"Error webhook post","plugin_id":"github","error":"SqlPostStore.Save: İleti kaydedilemedi., pq: value too long for type character varying(26)"}`

`İleti kaydedilemedi` means `The message could not be saved`

This error maybe not related to this plugin.

My environment:

Webhook payload: https://gist.github.com/ecylmz/183cabab78939dc8bc0403b9bb7419b9

GitHub Plugin: (github - 2.0.0)

Mattermost server:
```
Version: 5.26.0
Build Number: 5.26.2
Build Date: Fri Sep  4 02:04:34 UTC 2020
Build Hash: bed73717c54643abc1e2211e794d59a25b2cc823
Build Enterprise Ready: true
DB Version: 5.26.0
```

postgres (PostgreSQL) 12.4 (Debian 12.4-1.pgdg100+1)

What do you think could cause this error?

Thanks.

hanzei:
Hi @ecylmz,

Thanks for the report. Would you be open on installing a custom version with additional logging? That would help with debuggin.

ecylmz:
> Would you be open on installing a custom version with additional logging?

Of course! How can I install the custom version you mentioned?

hanzei:
Could you please DM me on the [Mattermost Community Server](https://community.mattermost.com)?

ecylmz:
I can't join any team. 
![image](https://user-images.githubusercontent.com/248267/99398980-1556ad80-28f6-11eb-97bd-dadbbab6821c.png)

If it is not a problem for you, you can send the instructions by mail. my email: emrecan [at] ecylmz.com

hanzei:
@ecylmz Thanks for the report. We are looking into this.

Meanwhile I've send you the build via mail.

jwilander:
@ecylmz I updated the total users limit for the team. Can you try again?

ecylmz:
@hanzei When i install the plugin, i get the following error:

```
[...]

PR #768, opened 2024-04-15
[MM-405]: Fixed the pull review request events not working

Kshitij-Katiyar:
#### Summary
Fixed the issue that the pull request review event not working. Removed the GetComment.GetDiffHunk from the template as it can be very long and not that very useful. Also changed the name for the type of the post. 

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/405


server/plugin/webhook.go
```diff
@@ -937,7 +937,7 @@ func (p *Plugin) postPullRequestReviewCommentEvent(event *github.PullRequestRevi
 
 	post := &model.Post{
 		UserId:  p.BotUserID,
-		Type:    "custom_git_pull_review_comment",
+		Type:    "custom_git_pr_comment",
```
hanzei, line 940: While we are not using the custom type currently, I think there is value in keeping the type the same for each post.

What is the reason behind changing it?

server/plugin/template.go
```diff
@@ -303,9 +303,9 @@ Assignees: {{range $i, $el := .Assignees -}} {{- if $i}}, {{end}}{{template "use
 
 	template.Must(masterTemplate.New("pullRequestReviewEvent").Funcs(funcMap).Parse(`
 {{template "repo" .GetRepo}} {{template "user" .GetSender}}
-{{- if eq .GetReview.GetState "APPROVED"}} approved
-{{- else if eq .GetReview.GetState "COMMENTED"}} commented on
-{{- else if eq .GetReview.GetState "CHANGES_REQUESTED"}} requested changes on
```
hanzei, line 308: Did GitHub document the changes anywhere?

hanzei (changes requested):
Glad to see this bug getting fixed :rocket:

server/plugin/webhook.go
Kshitij-Katiyar: I kept getting the error that the length of the type should be less than 2, that's why I have reduced it. 
@hanzei

server/plugin/template.go
Kshitij-Katiyar: @hanzei I was getting the state in small case and in the code as well they have been changed to lower case in the using toLower. Let me try to find the GitHub documentation for this.

server/plugin/template.go
```diff
@@ -314,7 +314,6 @@ Assignees: {{range $i, $el := .Assignees -}} {{- if $i}}, {{end}}{{template "use
 	template.Must(masterTemplate.New("newReviewComment").Funcs(funcMap).Parse(`
 {{template "repo" .GetRepo}} New review comment by {{template "user" .GetSender}} on {{template "pullRequest" .GetPullRequest}}:
 
-{{.GetComment.GetDiffHunk}}
```
mickmister, line 317: Can you provide a screenshot for what this looks like?

mickmister (commented):
@Kshitij-Katiyar There's a PR that was submitted that solves this same issue https://github.com/mattermost/mattermost-plugin-github/pull/778. Can you please take a look there and compare any changes that may need to be added here?

server/plugin/template.go
Kshitij-Katiyar: @hanzei Can't find a documentation but here is a screenshot of the webhook payload
Screenshot from 2024-06-21 12-33-38

Kshitij-Katiyar:
> @Kshitij-Katiyar There's a PR that was submitted that solves this same issue #778. Can you please take a look there and compare any changes that may need to be added here?

Thanks for the PR, added some useful changes from the PR

server/plugin/template.go
Kshitij-Katiyar: **Old UI**
Screenshot from 2024-06-21 12-46-20

**New UI**
new

webapp/package.json
```diff
@@ -58,7 +58,7 @@
     "sass": "1.60.0",
     "sass-loader": "10.4.1",
     "style-loader": "1.2.1",
-    "typescript": "4.0.3",
+    "typescript": "4.2.3",
```
mickmister, line 61: This is adding a lot to the diff of the PR

webapp/src/components/sidebar_right/github_items.tsx
```diff
@@ -350,7 +350,7 @@ function getReviewText(item: GithubItem, style: any, secondLine: boolean) {
             return false;
         }
 
-        if (v.state === 'COMMENTED' || v.state === 'DISMISSED') {
+        if (v.state === 'commented' || v.state === 'dismissed') {
```
mickmister, line 353: Can we put these into a typescript enum?

server/plugin/webhook.go
```diff
@@ -908,6 +908,19 @@ func (p *Plugin) postPullRequestReviewCommentEvent(event *github.PullRequestRevi
 		return
 	}
 
+	post := &model.Post{
```
mickmister, line 911: Putting the post above the loop reintroduces the bug that https://github.com/mattermost/mattermost-plugin-github/pull/755 solves. Can we revert the changes here that are not related to `approved` etc.?

server/plugin/webhook.go
```diff
@@ -908,6 +908,19 @@ func (p *Plugin) postPullRequestReviewCommentEvent(event *github.PullRequestRevi
...
+		UserId:  p.BotUserID,
+		Type:    "custom_git_pr_comment",
+		Message: newReviewMessage,
+	}
+
+	repoName := strings.ToLower(repo.GetFullName())
+	commentID := event.GetComment().GetID()
```
mickmister, line 918: Note that these are quick functions that do not do network calls, in case this was moved up here for performance reasons

server/plugin/webhook.go
```diff
@@ -935,15 +948,6 @@ func (p *Plugin) postPullRequestReviewCommentEvent(event *github.PullRequestRevi
...
-
-		repoName := strings.ToLower(repo.GetFullName())
-		commentID := event.GetComment().GetID()
-
-		post.AddProp(postPropGithubRepo, repoName)
-		post.AddProp(postPropGithubObjectID, commentID)
-		post.AddProp(postPropGithubObjectType, githubObjectTypePRReviewComment)
-
 		post.ChannelId = sub.ChannelID
 		if err = p.client.Post.CreatePost(post); err != nil {
```
mickmister, line 952: Referencing https://github.com/mattermost/mattermost-plugin-github/pull/768/files#r1649049447 the `p.client.Post.CreatePost(post)` here mutates the post instead of returning the created post, which has had side effects in this plugin. I think we should instead use `p.API.CreatePost(post)` to avoid the risk of introducing the issue on unrelated code changes cc @hanzei

webapp/package.json
Kshitij-Katiyar: @mickmister Without these changes the `ci-lint` will fail. Let me know if you want us to find any other way around to fix the failing lint test other than this.

Kshitij-Katiyar:
@mickmister Fixed the comments, please look into it.

server/plugin/webhook.go
```diff
@@ -945,8 +945,8 @@ func (p *Plugin) postPullRequestReviewCommentEvent(event *github.PullRequestRevi
 		post.AddProp(postPropGithubObjectType, githubObjectTypePRReviewComment)
 
 		post.ChannelId = sub.ChannelID
-		if err = p.client.Post.CreatePost(post); err != nil {
-			p.client.Log.Warn("Error webhook post", "post", post, "error", err.Error())
+		if _, appErr := p.API.CreatePost(post); appErr != nil {
+			p.client.Log.Warn("Error webhook post", "post", post, "error", appErr.Error())
```
hanzei, line 949: `p.API` should not be used. The returned `model.AppError` makes things way too hard. `p.client` is they way to go.

hanzei (changes requested):

server/plugin/webhook.go
raghavaggarwal2308: @hanzei Please refer to this comment: https://github.com/mattermost/mattermost-plugin-github/pull/768#discussion_r1649367568

mickmister: In this case, the `p.client.Post` method has essentially a bug in it, that has caused issues in this plugin because it mutates the post that is passed in. Is it really a "bug"? Maybe not, but I think we should avoid using it since it has caused issues in production that we then need to fix

2/5 The benefits of using `error` over `AppError` in this case are not worth it imo. I think keeping it as `p.client.Post` will make potential issues come back, just as they did in this PR

webapp/src/components/sidebar_right/github_items.tsx
```diff
@@ -19,6 +19,7 @@ import TickIcon from '../../images/icons/tick';
 import SignIcon from '../../images/icons/sign';
 import ChangesRequestedIcon from '../../images/icons/changes_requested';
 import {getLabelFontColor} from '../../utils/styles';
+import {ReviewState} from '../../constants/index';
```
mickmister, line 22:
```suggestion
import {ReviewState} from '../../constants';
```

server/plugin/webhook.go
mickmister: @Kshitij-Katiyar I'm thinking we should keep the changes in this PR to a minimum so let's keep this as-is

Kshitij-Katiyar:
@mickmister Fixed the comments, please have a look

mickmister (approved):

hanzei (approved):

AayushChaudhary0001 (approved):

- Checked the pull review request events
- Tested the notification for pull review request events 

The PR was working fine for all the conditions, LGTM. Approved

PR #770, opened 2024-04-26
Update subscription messages with respect to webhook

ayusht2810:
#### Summary
- Update subscription message when user is not authorized to access webhooks
- Update subscription message when there is no webhook present

#### Screenshots
- When user is not authorized to access webhooks
##### Before 
image
##### After
image

- When there is no webhook present
##### Before 
image
##### After
image

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/715
Fixes https://github.com/mattermost/mattermost-plugin-github/issues/732

#### What to test?
- Create a subscription with the user who doesn't have access to create webhooks and verify the post
- Similarly, create a subscription with no prior webhook present and verify the post

### DIFF
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -458,6 +458,7 @@ func (p *Plugin) handleSubscribesAdd(_ *plugin.Context, args *model.CommandArgs,
 		found, foundErr := p.checkIfConfiguredWebhookExists(ctx, githubClient, repo, owner)
 		if foundErr != nil {
 			if strings.Contains(foundErr.Error(), "404 Not Found") {
+				// We are not returning an error here and just a subscription success message, as the above error condition occurs when the user is not authorised to access webhooks.
 				return subOrgMsg
 			}
 			return errors.Wrap(foundErr, "failed to get the list of webhooks").Error()
@@ -497,6 +498,7 @@ func (p *Plugin) handleSubscribesAdd(_ *plugin.Context, args *model.CommandArgs,
 	found, err := p.checkIfConfiguredWebhookExists(ctx, githubClient, repo, owner)
 	if err != nil {
 		if strings.Contains(err.Error(), "404 Not Found") {
+			// We are not returning an error here and just a subscription success message, as the above error condition occurs when the user is not authorised to access webhooks.
 			return msg
 		}
 		return errors.Wrap(err, "failed to get the list of webhooks").Error()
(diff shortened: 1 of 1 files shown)


hanzei (approved):

server/plugin/command.go
```diff
@@ -459,7 +458,7 @@ func (p *Plugin) handleSubscribesAdd(_ *plugin.Context, args *model.CommandArgs,
 		found, foundErr := p.checkIfConfiguredWebhookExists(ctx, githubClient, repo, owner)
 		if foundErr != nil {
 			if strings.Contains(foundErr.Error(), "404 Not Found") {
-				return errorWebhookToUser
+				return subOrgMsg
 			}
 			return errors.Wrap(foundErr, "failed to get the list of webhooks").Error()
 		}
```
mickmister, line 464: In the case where the user has access to check webhooks but there isn't one, do we get an error and 404 here? I wonder if there's a way to know if the user has permissions or not

ayusht2810: @mickmister In the above case, we get no error and the value of `found` as false.

server/plugin/command.go
```diff
@@ -459,7 +458,7 @@ func (p *Plugin) handleSubscribesAdd(_ *plugin.Context, args *model.CommandArgs,
 		found, foundErr := p.checkIfConfiguredWebhookExists(ctx, githubClient, repo, owner)
 		if foundErr != nil {
 			if strings.Contains(foundErr.Error(), "404 Not Found") {
-				return errorWebhookToUser
+				return subOrgMsg
 			}
```
hanzei, line 462: Let's add a comment explaining why we are not returning an error here.

hanzei (changes requested):
The changes LGTM, but let's document why the code works this way

ayusht2810:
@hanzei added the comment for the above changes in the code. Please re-review

hanzei (approved):

raghavaggarwal2308 (approved):

mickmister (approved):

mickmister:
It looks like the screenshots changed the behavior of public vs ephemeral messages. Is this the case?

image

ayusht2810:
@mickmister Updated the messages:
image
Let me know if anything else needs to be changed here.

mickmister (approved):

AayushChaudhary0001 (approved):
This PR has been tested for the following scenarios:

- Checked the updated subscription message when user is not authorized to access webhooks
- Checked the updated subscription message when there is no webhook present

The PR was working fine for both the conditions. LGTM. Approved

Issue #647, opened 2023-02-27
Hide UsePreregisteredApplication on on-prem instances


hanzei:
The `UsePreregisteredApplication` setting should only be used on Cloud instances. To avoid confusion, the settings should not be visible on on-prem instances. This can be done using the [`hosting`](https://developers.mattermost.com/integrate/plugins/manifest-reference/#settings_schema.settings.hosting) field.

Kshitij-Katiyar:
sure @hanzei working on this.

hanzei:
Thanks @Kshitij-Katiyar :+1:

PR #774, opened 2024-05-02
[GH-647]: Hide UsePreregisteredApplication on on-prem instances

Kshitij-Katiyar:
#### Summary
Using the hoisting property the UsePreregisteredApplication will only be on cloud servers.

#### Ticket Link
https://github.com/mattermost/mattermost-plugin-github/issues/647

### Testing Steps
The UsePreregisteredApplication will be hidden in on-prem servers and can only be seen in cloud instances.

### DIFF
--- a/plugin.json
+++ b/plugin.json
@@ -25,7 +25,7 @@
                 "key": "UsePreregisteredApplication",
                 "display_name": "Use Preregistered OAuth Application:",
                 "type": "bool",
-                "help_text": "Set to false if using GitHub Enterprise. When true, instructs the plugin to use the preregistered GitHub OAuth application - application registration steps can be skipped. Requires [Chimera Proxy](https://github.com/mattermost/chimera) URL to be configured for the server. Cannot be used with GitHub enterprise. **This setting is intended to be used with Mattermost Cloud instances only.**",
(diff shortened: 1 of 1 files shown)


plugin.json
```diff
@@ -26,7 +26,8 @@
                 "display_name": "Use Preregistered OAuth Application:",
                 "type": "bool",
                 "help_text": "Set to false if using GitHub Enterprise. When true, instructs the plugin to use the preregistered GitHub OAuth application - application registration steps can be skipped. Requires [Chimera Proxy](https://github.com/mattermost/chimera) URL to be configured for the server. Cannot be used with GitHub enterprise. **This setting is intended to be used with Mattermost Cloud instances only.**",
```
hanzei, line 28: We don't need the warning now.
```suggestion
                "help_text": "Set to false if using GitHub Enterprise. When true, instructs the plugin to use the preregistered GitHub OAuth application - application registration steps can be skipped. Requires [Chimera Proxy](https://github.com/mattermost/chimera) URL to be configured for the server. Cannot be used with GitHub enterprise.",
```

Kshitij-Katiyar:
> Should we update `min_server_version` to ensure newer server pick up the setting correctly?

I have tested this on mattermost cloud server v9.4.0 and it was working fine for that. @hanzei

mickmister (approved):

AayushChaudhary0001 (commented):
The PR above has been tested for the config UsePreregisteredApplication for both cloud and on-prem instances. The config is not visible in the plugin configs for on-prem instances of MM. This PR is working fine, approved.

hanzei:
@Kshitij-Katiyar Would you please apply the same change to the GitLab plugin? Thanks!

Issue #730, opened 2024-01-18
[Feature request] When linking to pull requests / issues on a private repository, show a preview instead of an error page


DanielHeath:
Currently, if I link to an issue on a private github repository, the link preview is the github "Build software better, together" page.

It would be really nice if (similar to code previews) this replaced the preview with the PR/issue title (and perhaps labels/projects/milestones/assignees).

PR #779, opened 2024-05-10
Link preview

Sn-Kinos:
#### Summary
Make GitHub link preview for PR and issue. It can show private repository if OAuth connected.

#### Screenshots
| Denim | Onyx |
| --- | --- |
| image | image |

#### Ticket Link

#730

### DIFF
--- a/webapp/src/components/link_preview/index.jsx
+++ b/webapp/src/components/link_preview/index.jsx
@@ -1,0 +1,14 @@
+// Copyright (c) 2015-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+import {connect} from 'react-redux';
+
+import manifest from 'manifest';
+
+import {LinkPreview} from './link_preview';
+
+const mapStateToProps = (state) => {
+    return {connected: state[`plugins-${manifest.id}`].connected};
+};
+
+export default connect(mapStateToProps, null)(LinkPreview);
--- a/webapp/src/components/link_preview/link_preview.jsx
+++ b/webapp/src/components/link_preview/link_preview.jsx
@@ -1,0 +1,183 @@
+import {GitMergeIcon, GitPullRequestIcon, IssueClosedIcon, IssueOpenedIcon} from '@primer/octicons-react';
+import PropTypes from 'prop-types';
+import React, {useEffect, useState} from 'react';
+import ReactMarkdown from 'react-markdown';
+import './preview.css';
+
+import Client from 'client';
+import {getLabelFontColor} from '../../utils/styles';
+import {isUrlCanPreview} from 'src/utils/github_utils';
+
+const maxTicketDescriptionLength = 160;
+
+export const LinkPreview = ({embed: {url}, connected}) => {
+    const [data, setData] = useState(null);
+    useEffect(() => {
+        const initData = async () => {
+            if (isUrlCanPreview(url)) {
+                const [owner, repo, type, number] = url.split('github.com/')[1].split('/');
+
+                let res;
+                switch (type) {
+                case 'issues':
+                    res = await Client.getIssue(owner, repo, number);
+                    break;
+                case 'pull':
+                    res = await Client.getPullRequest(owner, repo, number);
+                    break;
+                }
+                if (res) {
+                    res.owner = owner;
+                    res.repo = repo;
+                    res.type = type;
+                }
+                setData(res);
+            }
+        };
+
+        // show is not provided for Mattermost Server < 5.28
+        if (!connected || data) {
+            return;
+        }
+
+        initData();
+    }, [connected, data, url]);
+
+    const getIconElement = () => {
+        const iconProps = {
+            size: 'small',
+            verticalAlign: 'text-bottom',
+        };
+
+        let icon;
+        let color;
+        switch (data.type) {
+        case 'pull':
+            icon = <GitPullRequestIcon {...iconProps}/>;
+
+            color = '#28a745';
+            if (data.state === 'closed') {
+                if (data.merged) {
+                    color = '#6f42c1';
+                    icon = <GitMergeIcon {...iconProps}/>;
+                } else {
+                    color = '#cb2431';
+                }
+            }
+
+            break;
+        case 'issues':
+            color = data.state === 'open' ? '#28a745' : '#cb2431';
+
+            if (data.state === 'open') {
+                icon = <IssueOpenedIcon {...iconProps}/>;
+            } else {
+                icon = <IssueClosedIcon {...iconProps}/>;
+            }
+            break;
+        }
+        return (
+            <span style={{color}}>
+                {icon}
+            </span>
+        );
+    };
+
+    if (data) {
+        let date = new Date(data.created_at);
+        date = date.toDateString();
+
+        let description = '';
+        if (data.body) {
+            description = data.body.substring(0, maxTicketDescriptionLength).trim();
+            if (data.body.length > maxTicketDescriptionLength) {
+                description += '...';
+            }
+        }
+
+        return (
+            <div className='github-preview github-preview--large p-4 mt-1 mb-1'>
+                <div className='header'>
+                    <span className='repo'>
+                        {data.repo}
+                    </span>
+                    {' on '}
+                    <span>{date}</span>
+                </div>
+
+                <div className='body d-flex'>
+
+                    {/* info */}
+                    <div className='preview-info mt-1'>
+                        <a
+                            href={url}
+                            target='_blank'
+                            rel='noopener noreferrer'
+                        >
+                            <h5 className='mr-1'>
+                                <span className='pr-2'>
+                                    { getIconElement() }
+                                </span>
+                                {data.title}
+                            </h5>
+                            <span>{'#' + data.number}</span>
+                        </a>
+                        <div className='markdown-text mt-1 mb-1'>
+                            <ReactMarkdown linkTarget='_blank'>{description}</ReactMarkdown>
+                        </div>
+
+                        <div className='sub-info mt-1'>
+                            {/* base <- head */}
+                            {data.type === 'pull' && (
+                                <div className='sub-info-block'>
+                                    <h6 className='mt-0 mb-1'>{'Base ← Head'}</h6>
+                                    <div className='base-head'>
+                                        <span
+                                            title={data.base.ref}
+                                            className='commit-ref'
+                                        >{data.base.ref}
+                                        </span> <span className='mx-1'>{'←'}</span>{' '}
+                                        <span
+                                            title={data.head.ref}
+                                            className='commit-ref'
+                                        >{data.head.ref}
+                                        </span>
+                                    </div>
+                                </div>
+                            )}
+
+                            {/* Labels */}
+                            {data.labels && data.labels.length > 0 && (
+                                <div className='sub-info-block'>
+                                    <h6 className='mt-0 mb-1'>{'Labels'}</h6>
+                                    <div className='labels'>
+                                        {data.labels.map((label, idx) => {
+                                            return (
+                                                <span
+                                                    key={`${label.name}-${idx}`}
+                                                    className='label'
+                                                    title={label.description}
+                                                    style={{backgroundColor: '#' + label.color, color: getLabelFontColor(label.color)}}
+                                                >
+                                                    <span>{label.name}</span>
+                                                </span>
+                                            );
+                                        })}
+                                    </div>
+                                </div>
+                            )}
+                        </div>
+                    </div>
+                </div>
+            </div>
+        );
+    }
+    return null;
+};
+
+LinkPreview.propTypes = {
+    embed: {
+        url: PropTypes.string.isRequired,
+    },
+    connected: PropTypes.bool.isRequired,
+};
--- a/webapp/src/components/link_preview/preview.css
+++ b/webapp/src/components/link_preview/preview.css
@@ -1,0 +1,136 @@
+:root {
+    --light-gray: #6a737d;
+    --blue: #274466;
+    --light-blue: #eff7ff;
+}
+
+@media (min-width: 544px) {
+    .github-preview--large {
+        min-width: 320px;
+    }
+}
+
+/* Github Preview */
+.github-preview {
+    background-color: var(--center-channel-bg-rgb);
+    box-shadow: 0 2px 3px rgba(0,0,0,.08);
+    position: relative;
+    width: 100%;
+    max-width: 700px;
+    border-radius: 4px;
+    border: 1px solid rgba(var(--center-channel-color-rgb), 0.16);
+}
+
+/* Header */
+.github-preview .header {
+    color: rgba(var(--center-channel-color-rgb), 0.64);
+    font-size: 11px;
+    line-height: 16px;
+    white-space: nowrap;
+}
+
+.github-preview .header a {
+    text-decoration: none;
+    color: rgba(var(--center-channel-color-rgb), 0.64);
+    display: inline-block;
+}
+
+.github-preview .header .repo {
+    color: var(--center-channel-color-rgb);
+}
+
+/* Body */
+.github-preview .body > span {
+    line-height: 1.25;
+}
+
+/* Info */
+.github-preview .preview-info {
+    line-height: 1.25;
+    display: flex;
+    flex-direction: column;
+}
+.github-preview .preview-info > a, .github-preview .preview-info > a:hover {
+    display: block;
+    text-decoration: none;
+    color: var(--link-color);
+}
+.github-preview .preview-info > a span {
+    color: var(--light-gray);
+}
+
+.github-preview .preview-info h5 {
+    font-weight: 600;
+    font-size: 14px;
+    display: inline;
+}
+
+.github-preview .preview-info .markdown-text {
+    max-height: 150px;
+    line-height: 1.25;
+    overflow: hidden;
+    word-break: break-word;
+    word-wrap: break-word;
+    overflow-wrap: break-word;
+    font-size: 12px;
+}
+
+.github-preview .preview-info .markdown-text::-webkit-scrollbar {
+    display: none;
+}
+
+.github-preview .sub-info {
+    display: flex;
+    width: 100%;
+    flex-wrap: wrap;
+    gap: 4px 0;
+}
+
+.github-preview .sub-info .sub-info-block {
+    display: flex;
+    flex-direction: column;
+    width: 50%;
+}
+
+/* Labels */
+.github-preview .labels {
+    display: flex;
+    justify-content: flex-start;
+    align-items: center;
+    flex-wrap: wrap;
+    gap: 4px;
+}
+
+.github-preview .label {
+    height: 20px;
+    padding: .15em 4px;
+    font-size: 12px;
+    font-weight: 600;
+    line-height: 15px;
+    border-radius: 2px;
+    box-shadow: inset 0 -1px 0 rgba(27,31,35,.12);
+    display: inline-block;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    max-width: 125px;
+}
+
+.github-preview .base-head {
+    display: flex;
+    line-height: 1;
+    align-items: center;
+}
+
+.github-preview .commit-ref {
+    position: relative;
+    display: inline-block;
+    padding: 0 5px;
+    font: .75em/2 SFMono-Regular,Consolas,Liberation Mono,Menlo,monospace;
+    color: var(--blue);
+    background-color: var(--light-blue);
+    border-radius: 3px;
+    max-width: 140px;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    overflow: hidden;
+}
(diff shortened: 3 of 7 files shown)


webapp/src/components/link_preview/index.jsx
Kshitij-Katiyar, line 1: @Sn-Kinos We are trying to move away from javascript to now TypeScript. Will it be possible to convert these files to typescript if possible ?

webapp/src/components/link_preview/link_preview.jsx
```diff
@@ -0,0 +1,183 @@
...
+import {isUrlCanPreview} from 'src/utils/github_utils';
+
+const maxTicketDescriptionLength = 160;
+
+export const LinkPreview = ({embed: {url}, connected}) => {
+    const [data, setData] = useState(null);
+    useEffect(() => {
+        const initData = async () => {
+            if (isUrlCanPreview(url)) {
+                const [owner, repo, type, number] = url.split('github.com/')[1].split('/');
```
Kshitij-Katiyar, line 18: can we check the length of the array just to make the code robust and less prone to panic.

webapp/src/components/link_preview/link_preview.jsx
```diff
@@ -0,0 +1,183 @@
...
+                    color = '#6f42c1';
+                    icon = <GitMergeIcon {...iconProps}/>;
+                } else {
+                    color = '#cb2431';
+                }
+            }
+
+            break;
+        case 'issues':
+            color = data.state === 'open' ? '#28a745' : '#cb2431';
```
Kshitij-Katiyar, line 70: can we move these to constants or to the CSS files if possible ?

webapp/src/utils/github_utils.js
```diff
@@ -0,0 +1,7 @@
+export function isUrlCanPreview(url) {
+    if (url.includes('github.com/')) {
+        const [owner, repo, type, number] = url.split('github.com/')[1].split('/');
```
Kshitij-Katiyar, line 3: can we add the check here as well.

webapp/src/components/link_preview/index.jsx
Sn-Kinos: I want to convert it to TS from JS but It's hard because there are no fundamental types requiring domain knowledge (e.g. state type from Mattermost web app) and ESLint didn't works well (e.g. It can't import with relative path from baseUrl) I think It will need more best practices for contributors to convert to TS appropriately. There are only one TS file on webapp 🥲

Kshitij-Katiyar: We are working on providing a template for converting files from js to ts. 
cc: @mickmister

webapp/src/components/link_preview/preview.css
```diff
@@ -0,0 +1,156 @@
...
+    line-height: 1.25;
+}
+
+/* Info */
+.github-preview .preview-info {
+    line-height: 1.25;
+    display: flex;
+    flex-direction: column;
+}
+.github-preview .preview-info > a, .github-preview .preview-info > a:hover {
```
mickmister, line 57: We can make this file `scss` them we can use nesting for this selector logic

webapp/src/components/link_preview/link_preview.jsx
```diff
@@ -0,0 +1,184 @@
...
+                const [owner, repo, type, number] = url.split('github.com/')[1].split('/');
+
+                let res;
+                switch (type) {
+                case 'issues':
+                    res = await Client.getIssue(owner, repo, number);
+                    break;
+                case 'pull':
+                    res = await Client.getPullRequest(owner, repo, number);
+                    break;
```
mickmister, line 27: We'll want to inject these as props instead of accessing the client directly in this component

mickmister: Also I wonder what happens here if the user is connected, but doesn't have access to a private repo being linked here

webapp/src/components/link_preview/link_preview.jsx
```diff
@@ -0,0 +1,184 @@
...
+        return (
+            <span className={`pr-2 ${colorClass}`}>
+                {icon}
+            </span>
+        );
+    };
+
+    if (data) {
+        let date = new Date(data.created_at);
+        date = date.toDateString();
```
mickmister, line 91: It's best to "return early" to avoid large indents like this

```suggestion
    if (!data) {
        return null;
    }
      
    let date = new Date(data.created_at);
    date = date.toDateString();
```

webapp/src/components/link_preview/index.jsx
mickmister: @Sn-Kinos Can you please try to have the `webapp/src/components/link_preview/link_preview.jsx` file be typescript in particular?

webapp/src/components/link_preview/preview.css
```diff
@@ -0,0 +1,156 @@
+:root {
+    --light-gray: #6a737d;
+    --blue: #274466;
+    --light-blue: #eff7ff;
+    --github-merged: #6f42c1;
+    --github-closed: #cb2431;
+    --github-open: #28a745;
+    --github-not-planned: #6e7681
+}
```
mickmister, line 9: This css will affect the whole web application, which we don't want in this case. We can use sass variables here at compile time instead of setting these css vars

webapp/src/components/link_preview/preview.css
```diff
@@ -0,0 +1,156 @@
...
+
+@media (min-width: 544px) {
+    .github-preview--large {
+        min-width: 320px;
+    }
+}
+
+/* Github Preview */
+.github-preview {
```
mickmister, line 18: With sass we can nest everything under this class name in this file

webapp/src/components/link_tooltip/tooltip.css
```diff
@@ -86,6 +86,7 @@
 
 /* Labels */
 .github-tooltip .labels {
+    display: flex;
```
mickmister, line 89: What effect does this have on the tooltip's styling?

webapp/src/utils/github_utils.js
```diff
@@ -0,0 +1,7 @@
+export function isUrlCanPreview(url) {
```
mickmister, line 1: Can we have this file be typescript?

webapp/src/components/link_preview/link_preview.jsx
```diff
@@ -0,0 +1,184 @@
...
+import ReactMarkdown from 'react-markdown';
+import './preview.css';
+
+import Client from 'client';
+import {getLabelFontColor} from '../../utils/styles';
+import {isUrlCanPreview} from 'src/utils/github_utils';
+
+const maxTicketDescriptionLength = 160;
+
+export const LinkPreview = ({embed: {url}, connected}) => {
```
mickmister, line 13: We'll want to name this `LinkEmbedPreview`, along with the file named `link_embed_preview`

webapp/src/utils/github_utils.js
```diff
@@ -0,0 +1,7 @@
...
+    if (url.includes('github.com/') && url.split('github.com/')[1]) {
+        const [owner, repo, type, number] = url.split('github.com/')[1].split('/');
+        return !(!owner | !repo | !type | !number);
```
mickmister, line 4: Bit-wise or's 🤔  I see this is also being used in the tooltip code

We can do this to make it a bit more readable

```suggestion
        return Boolean(owner && repo && type && number);
```

mickmister (commented):
Excellent work @Sn-Kinos! Mostly LGTM, I have a few suggestions and some comments for discussion

matthewbirtch:
> @matthewbirtch I'm curious about your thoughts on this feature. It's essentially a clone of the link tooltip component in the form of a link embed preview. Do you think this duplication could be noisy or cause confusion? Note that the preview will only show for users that have their GitHub account connected.

@mickmister  I definitely see this as redundant with the hover popover that's been added. In my mind we would need to choose one or the other. Or make it configurable which kind of preview to use in the workspace.

Sn-Kinos:
> > @matthewbirtch I'm curious about your thoughts on this feature. It's essentially a clone of the link tooltip component in the form of a link embed preview. Do you think this duplication could be noisy or cause confusion? Note that the preview will only show for users that have their GitHub account connected.
> 
> @mickmister I definitely see this as redundant with the hover popover that's been added. In my mind we would need to choose one or the other. Or make it configurable which kind of preview to use in the workspace.

Without this feature, We can watch the preview like below on private repository. I thought it is uncomfortable, so I added this feature.
image

webapp/src/components/link_tooltip/tooltip.css
Sn-Kinos: `.labels` is not flex element so the properties below doesn't working. So I added it and flex works.

| Before | After |
| --- | --- |
| image | image |
| `.labels = height: 23px; (auto) display: block;`<br>( flex properties not working ) | `.labels = height: 20px; (auto) display: flex;`<br>( flex properties working )
| `.labels > * = height: 20px;` | `.labels > * = height: 20px;` |

webapp/src/components/link_preview/preview.css
Sn-Kinos: Can I use scss on this project? I thought I couldn't use SCSS because I couldn't find the SCSS file on this repository.

webapp/src/components/link_preview/link_preview.jsx
Sn-Kinos: It bypasses this and shows default embed preview like this.
image

webapp/src/components/link_preview/index.jsx
Sn-Kinos: @mickmister Yes I can. But I think I will need some templates for it. I need more informations of types such as the data from `Client`, state from redux, etc.

webapp/src/components/link_preview/preview.css
Sn-Kinos: Oh I founded it can use [with CSS](https://caniuse.com/?search=nesting).  May I work with it?

webapp/src/components/link_preview/preview.css
Sn-Kinos: Can I use scss on this project? I thought I couldn't use SCSS because I couldn't find the SCSS file on this repository.

mickmister: Yes `scss` is configured and should work in this project

mickmister:
@Sn-Kinos @matthewbirtch We can make this feature configurable per-user so they can choose for themselves. We would want to make it obvious that the user can configure these two features separately.

Also note that this feature will passively be making requests to GitHub potentially often compared to the link tooltip feature, which could have negative effects on the server in terms of performance.

On our community server, we use the [autolink plugin](https://github.com/mattermost-community/mattermost-plugin-autolink) which with our configuration of the plugin makes it so GitHub PR links are shortened and hyperlinked, which makes it so the webapp does not use the "embed" feature that this PR relies on in that case. So if I were to opt for the embed feature and not the tooltip, I wouldn't have the chance to use this feature that often on our server. Just something to point out since this could be the case for other servers as well.

webapp/src/components/link_embed_preview/link_embed_preview.jsx
```diff
@@ -0,0 +1,184 @@
...
+                    res.repo = repo;
+                    res.type = type;
+                }
+                setData(res);
+            }
+        };
+
+        // show is not provided for Mattermost Server < 5.28
+        if (!connected || data) {
+            return;
```
mickmister, line 40: The `show` parameter is related to the `LinkTooltip` component and not this one, so we can remove this comment

Also unrelated side note that the min server version for this plugin is now `7.1.0`, so we can assume `show` is available for the `LinkTooltip` component

webapp/src/components/link_embed_preview/link_embed_preview.jsx
```diff
@@ -0,0 +1,184 @@
...
+import {isUrlCanPreview} from '../../utils/github_utils';
+
+const maxTicketDescriptionLength = 160;
+
+export const LinkEmbedPreview = ({embed: {url}, connected}) => {
+    const [data, setData] = useState(null);
+    useEffect(() => {
+        const initData = async () => {
+            if (isUrlCanPreview(url)) {
+                const [owner, repo, type, number] = url.split('github.com/')[1].split('/');
```
mickmister, line 18: Can this cause issues if someone crafts a URL with `github.com/` in the subpath?

webapp/src/components/link_preview/link_preview.jsx
mickmister: I'm not sure I understand how that works. It shows the fallback because this component returns `null` in that case?

mickmister (commented):
The current code LGTM, though I think we need to decide how feature is configured since it essentially clashes with the existing tooltip preview feature

matthewbirtch:
> @Sn-Kinos @matthewbirtch We can make this feature configurable per-user so they can choose for themselves. We would want to make it obvious that the user can configure these two features separately.

@mickmister Rather than adding a user setting and increasing that complexity, I would suggest we go with a system-level setting in the admin console. That way everyone on the server sees things the same way. That can be very beneficial to ensure everyone has the same context.

webapp/src/components/link_preview/link_preview.jsx
Sn-Kinos: Yes, I think it caused by [this](https://github.com/Sn-Kinos/mattermost-plugin-github/blob/0dfcb288bf0e72dad68a78aec09de94fa7e49eb7/server/plugin/api.go#L1077)

webapp/src/components/link_embed_preview/link_embed_preview.jsx
Sn-Kinos: Yes, I checked it occurs on tooltip too so I took action for it

Sn-Kinos:
> I'm not sure I understand how that works. It shows the fallback because this component returns null in that case?

That's what I think. I don't know deeply about the fundamentals of `registerPostWillRenderEmbedComponent` so I have referenced `link_tooltip`.

Sn-Kinos:
> > @Sn-Kinos @matthewbirtch We can make this feature configurable per-user so they can choose for themselves. We would want to make it obvious that the user can configure these two features separately.
> 
> @mickmister Rather than adding a user setting and increasing that complexity, I would suggest we go with a system-level setting in the admin console. That way everyone on the server sees things the same way. That can be very beneficial to ensure everyone has the same context.

I think It is good to only work when `EnablePrivateRepo` is enabled.

webapp/src/components/link_preview/link_preview.jsx
mickmister: I meant the fact that the webapp handles this gracefully when the component ends up returning null. I don't think you can coordinate React components together like that

webapp/src/components/link_embed_preview/embed_preview.scss
```diff
@@ -0,0 +1,157 @@
...
+        > a span {
+            color: var($light-gray);
+        
+            h5 {
+                font-weight: 600;
+                font-size: 14px;
+                display: inline;
+            
+                span.github-preview-icon-opened {
+                    color: var($github-open);
```
mickmister, line 69: These scss variables shouldn't use `var` right?

```suggestion
                    color: $github-open;
```

mickmister (commented):
Great job on this @Sn-Kinos :+1:

From our discussion, I have one remaining ask to add a plugin setting to toggle between this feature and the link tooltip. This will go in the `plugin.json` file in the root of the repository

- Name: `EnableLinkPreviewEmbeds`
- Description: When enabled, whenever a link to an issue or pull request is shared in a channel, the embed content will be loaded for the specific user viewing it. This allows users to view previews from private repositories using their own connected account. Note that enabling this also disables the "link tooltip" feature in the plugin.

Please let me know your thoughts on this @Sn-Kinos

Sn-Kinos:
> Great job on this @Sn-Kinos 👍
> 
> From our discussion, I have one remaining ask to add a plugin setting to toggle between this feature and the link tooltip. This will go in the `plugin.json` file in the root of the repository
> 
> * Name: `EnableLinkPreviewEmbeds`
> * Description: When enabled, whenever a link to an issue or pull request is shared in a channel, the embed content will be loaded for the specific user viewing it. This allows users to view previews from private repositories using their own connected account. Note that enabling this also disables the "link tooltip" feature in the plugin.
> 
> Please let me know your thoughts on this @Sn-Kinos

@mickmister @matthewbirtch I don't think it's very essential to add setting. This link_embed_preview is necessary for private repository only. If the link is public repo, It shows embed preview in a normal way.
image

I think we can handle it with the 'EnablePrivateRepo' setting because the reason for developing link_embed_preview is that the information of private repository is not shown in the existing embedded preview.

mickmister:
@Sn-Kinos I discussed with @matthewbirtch offline, and we're thinking there should be a system console setting that allows the server to be configured for either the link tooltip or link embed. The rationale is that there is more functionality here than simply filling in the gaps for the missing open graph information for private repos, because here there's a rich display of information about the issue/PR with this feature, rather than the simple open graph preview that already shows for public repositories.

To give the system admin a choice to enable one or the other, we can use a radio setting like this one being used in the Jira plugin:
https://github.com/mattermost/mattermost-plugin-jira/blob/4b296a0c23d04f0575b4c86bc4c9f042137f52cc/plugin.json#L46

Please let me know your thoughts on this @Sn-Kinos

Sn-Kinos:
> @Sn-Kinos I discussed with @matthewbirtch offline, and we're thinking there should be a system console setting that allows the server to be configured for either the link tooltip or link embed. The rationale is that there is more functionality here than simply filling in the gaps for the missing open graph information for private repos, because here there's a rich display of information about the issue/PR with this feature, rather than the simple open graph preview that already shows for public repositories.
> 
> To give the system admin a choice to enable one or the other, we can use a radio setting like this one being used in the Jira plugin: https://github.com/mattermost/mattermost-plugin-jira/blob/4b296a0c23d04f0575b4c86bc4c9f042137f52cc/plugin.json#L46
> 
> Please let me know your thoughts on this @Sn-Kinos

I concerned [Also note that this feature will passively be making requests to GitHub potentially often compared to the link tooltip feature, which could have negative effects on the server in terms of performance.](https://github.com/mattermost/mattermost-plugin-github/pull/779#issuecomment-2123296131) on the comment. If It doesn't matter, It will be ok to add the settings.

mickmister:
> I concerned [Also note that this feature will passively be making requests to GitHub potentially often compared to the link tooltip feature, which could have negative effects on the server in terms of performance.](https://github.com/mattermost/mattermost-plugin-github/pull/779#issuecomment-2123296131) on the comment. If It doesn't matter, It will be ok to add the settings.

@Sn-Kinos The main reason we want to add the settings is the UX flow of toggling showing the issue/PR through tooltip or link embed preview. I'm not sure I understand the point you're making in your last comment here

Sn-Kinos:
> @Sn-Kinos The main reason we want to add the settings is the UX flow of toggling showing the issue/PR through tooltip or link embed preview. I'm not sure I understand the point you're making in your last comment here

I worried 'negative effects on the server in terms of performance' [on the comment](https://github.com/mattermost/mattermost-plugin-github/pull/779#issuecomment-2123296131). If It doesn't matter, I'll be ok :D

mickmister:
@Sn-Kinos I think the piece I'm not understanding is "If It doesn't matter". Are you saying "if the performance issues don't matter"? Or if something else doesn't matter? Sorry about this

The only blocker here is for the UX piece. I think the performance issue is probably not a real issue here

Sn-Kinos:
> @Sn-Kinos I think the piece I'm not understanding is "If It doesn't matter". Are you saying "if the performance issues don't matter"? Or if something else doesn't matter? Sorry about this
> 
> The only blocker here is for the UX piece. I think the performance issue is probably not a real issue here

Ahhhh You're right. I concerned about the performance issue. Now, I agree with you as you mentioned.

Sorry about the miscommunication problem. the translator that I used didn't work properly I presume.

Sn-Kinos:
I wonder if there's anything else I need to do 🏃‍♂️

mickmister:
@Sn-Kinos Sorry for the confusing conversation here. There is one remaining requirement here to implement a plugin setting to toggle this and the link tooltip feature, as discussed here https://github.com/mattermost/mattermost-plugin-github/pull/779#issuecomment-2129814712. Does this make sense?

Sn-Kinos:
@mickmister I understand. What are the options I need to make?

mickmister:
@Sn-Kinos We can have these two boolean plugin settings:

- `Enable Link Tooltip` Default on - "When enabled, hovering over an issue or PR will show a tooltip preview of the issue/PR"
- `Enable Post Embed Preview` Default off - "When enabled, links to issues/PRs will have a preview of the issue/PR shown within the post body."

hanzei (commented):
@ayusht2810 @Kshitij-Katiyar can you please review the PR?

raghavaggarwal2308:
@Sn-Kinos THank you for the contribution! Will it be possible for you to convert the newly added Javascript files to Typescript?

abbas-dependable-naqvi:
@Sn-Kinos Are you willing to work on this?

PR #800, opened 2024-07-08
Don't return an 500 error when the plugin is not configured

hanzei:
#### Summary
https://www.rfc-editor.org/rfc/rfc9110.html#name-422-unprocessable-content states
> The 422 (Unprocessable Content) status code indicates that the server understands the content type of the request content (hence a [415 (Unsupported Media Type)](https://www.rfc-editor.org/rfc/rfc9110.html#status.415) status code is inappropriate), and the syntax of the request content is correct, but it was unable to process the contained instructions. For example, this status code can be sent if an XML request content contains well-formed (i.e., syntactically correct), but semantically erroneous XML instructions.

#### Ticket Link
https://hub.mattermost.com/private-core/pl/1jafc1wuftnmzygdittxd73nia

### DIFF
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -184,7 +184,7 @@ func (p *Plugin) checkConfigured(next http.Handler) http.Handler {
 		config := p.getConfiguration()
 
 		if err := config.IsValid(); err != nil {
-			http.Error(w, "This plugin is not configured.", http.StatusNotImplemented)
+			http.Error(w, "This plugin is not configured.", http.StatusUnprocessableEntity)
 			return
 		}
 


server/plugin/api.go
```diff
@@ -184,7 +184,7 @@ func (p *Plugin) checkConfigured(next http.Handler) http.Handler {
 		config := p.getConfiguration()
 
 		if err := config.IsValid(); err != nil {
-			http.Error(w, "This plugin is not configured.", http.StatusNotImplemented)
```
wiggin77, line 187: `UnprocessableEntity` comes from WebDAV and in practice means the request has the correct syntax but the content is invalid in some way.  It is an error that is expected to be corrected by the caller, not the server. 

`NotImplemented` or `InternalServerError` seems like a more suitable error condition for invalid config. I may not be fully understanding the problem, but  I think it *should* trigger alerts to an admin and the server-side condition should be fixed.  

I think `NotImplemented` was not a bad choice, acting much like if the plugin was disabled but with something to differentiate between disabled versus not configured.

If cloud team really wants this, I'll approve.  I would like to caution against hiding errors in obscure HTTP codes.

stylianosrigas: I do agree with Doug here. Although this creates 500s that we cannot solve on our side as we are not responsible for server and plugin configuration, it is a valid error and probably we should keep it this way. The proper path is to ask the admin to fix the configuration on their side.

stafot: @stylianosrigas By "we" we mean the server and the administrator of the server. So even if I agree that `Not Implemented` might not be the best choice, I think that the error should be a `5xx` if we want to remain semantically accurate, `4xx` generally means that the client is responsible and for sure here this is not the case.  So my 👍🏾 to the comments of @wiggin77 and @stylianosrigas.

hanzei: Thanks for the feedback, folks. You correctly pointed out that an unconfigured plugin is a server error and not a client one.


Closing.

PR #810, opened 2024-07-31
Run Playwright Test on pull requests

wiggin77:
#### Summary
Run Playwright tests on each pull request so that tests are run before merging PRs.

#### Ticket Link
NONE

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,8 +1,6 @@
 name: Playwright Tests
 on:
   pull_request:
-  schedule:
-    - cron: "0 0 * * *"
   workflow_run:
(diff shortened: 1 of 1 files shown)


wiggin77:
Looks like Playwright tests have been disabled a long time, and are completely broken.

fmartingr:
Do we want to enable this as-is or are tests going to be fixed before enabling this pipeline step?

wiggin77:
> Do we want to enable this as-is or are tests going to be fixed before enabling this pipeline step?

No point enabling it now. It is too far broken to make a quick fix.

wiggin77:
We will revisit this at a later date.

PR #824, opened 2024-09-16
Template PR for server unit test cases using GoMock and Mockgen

Kshitij-Katiyar:
#### Summary
1) Added the base setup to use the Gomock and Mockgen to create mocks in the github plugin.
2) Created the PR for adding the test cases in the github plugin using mocks for command.go file. Mainly 
/github help and /github about command. 

For reference of GoMock: https://www.codecentric.de/wissens-hub/blog/gomock-tutorial


server/plugin/command_test.go
```diff
@@ -4,9 +4,41 @@ import (
...
+			GitHubOrg:               "mockOrg",
+			GitHubOAuthClientID:     "mockID",
+			GitHubOAuthClientSecret: "mockSecret",
+			EncryptionKey:           "mockKey123456789",
+		})
+	p.initializeAPI()
+
+	p.store = mockKvStore
+
+	p.BotUserID = "mockBotId"
```
raghavaggarwal2308, line 34:
```suggestion
	p.BotUserID = "mockBotID"
```

server/plugin/command_test.go
```diff
@@ -4,9 +4,41 @@ import (
...
+// Function to get the plugin object for test cases.
+func getPluginTest(api *plugintest.API, mockKvStore *mocks.MockKvStore) *Plugin {
+	p := NewPlugin()
+	p.setConfiguration(
+		&Configuration{
+			GitHubOrg:               "mockOrg",
+			GitHubOAuthClientID:     "mockID",
+			GitHubOAuthClientSecret: "mockSecret",
+			EncryptionKey:           "mockKey123456789",
+		})
```
raghavaggarwal2308, line 29: I think the indentation is off here

server/plugin/command_test.go
```diff
@@ -239,3 +271,65 @@ func TestCheckConflictingFeatures(t *testing.T) {
...
+			commandArgs: &model.CommandArgs{Command: "/github help", ChannelId: "test-channelId", RootId: "test-rootId", UserId: "test-userId"},
+			expectedMsg: "###### Mattermost GitHub Plugin - Slash Command Help\n",
+			SetupMockStore: func(mks *mocks.MockKvStore) {
+				mks.EXPECT().Get(gomock.Any(), gomock.Any()).DoAndReturn(func(key string, value interface{}) error {
+					// Cast the value to the appropriate type and updated it
+					if userInfoPtr, ok := value.(**GitHubUserInfo); ok {
+						*userInfoPtr = &GitHubUserInfo{
+							// Mock user info data
+							Token: &oauth2.Token{
+								AccessToken: "ycbODW-BWbNBGfF7ac4T5RL5ruNm5BChCXgbkY1bWHqMt80JTkLsicQwo8de3tqfqlfMaglpgjqGOmSHeGp0dA==",
```
raghavaggarwal2308, line 297: Are we using an actual token here if not we should replace it with a mock value

Kshitij-Katiyar: It is a mock value only, but of a specific size i.e. 88 bytes, so using a random string here

raghavaggarwal2308: We should replace it with a value like "mockToken-mockToken-MockToken........". This will clear the confusion for the reviewer

raghavaggarwal2308 (approved):

wiggin77 (approved):
LGTM :+1:

PR #838, opened 2024-10-21
[MM-874]: Fixed the issue of LHS not working without setting organisation lock

Kshitij-Katiyar:
#### Summary
Fixed the issue of Github LHS not working if any organization is not set in the organization lock of the plugin config

#### What to test
- Check the data of LHS with and without organisation lock

### DIFF
--- a/server/plugin/graphql/lhs_request.go
+++ b/server/plugin/graphql/lhs_request.go
@@ -23,22 +23,6 @@ func (c *Client) GetLHSData(ctx context.Context) ([]*github.Issue, []*github.Iss
 	orgsList := c.getOrganizations()
 	var resultReview, resultAssignee, resultOpenPR []*github.Issue
 
-	if len(orgsList) == 0 {
-		return c.fetchLHSData(ctx, resultReview, resultAssignee, resultOpenPR, false, "")
-	}
-
-	var err error
-	for _, org := range orgsList {
-		resultReview, resultAssignee, resultOpenPR, err = c.fetchLHSData(ctx, resultReview, resultAssignee, resultOpenPR, true, org)
-		if err != nil {
-			return nil, nil, nil, err
-		}
-	}
-
-	return resultReview, resultAssignee, resultOpenPR, nil
-}
-
-func (c *Client) fetchLHSData(ctx context.Context, resultReview, resultAssignee, resultOpenPR []*github.Issue, setOrgCheck bool, org string) ([]*github.Issue, []*github.Issue, []*github.Issue, error) {
 	params := map[string]interface{}{
 		queryParamOpenPRQueryArg:    githubv4.String(fmt.Sprintf("author:%s is:pr is:%s archived:false", c.username, githubv4.PullRequestStateOpen)),
 		queryParamReviewPRQueryArg:  githubv4.String(fmt.Sprintf("review-requested:%s is:pr is:%s archived:false", c.username, githubv4.PullRequestStateOpen)),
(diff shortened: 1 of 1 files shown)


server/plugin/graphql/lhs_request.go
```diff
@@ -22,72 +22,89 @@ const (
...
+		resultReview, resultAssignee, resultOpenPR, err = c.fetchLHSData(ctx, resultReview, resultAssignee, resultOpenPR, true, org)
+		if err != nil {
+			return nil, nil, nil, err
 		}
+	}
 
+	return resultReview, resultAssignee, resultOpenPR, nil
+}
+
+func (c *Client) fetchLHSData(ctx context.Context, resultReview, resultAssignee, resultOpenPR []*github.Issue, setOrgCheck bool, org string) ([]*github.Issue, []*github.Issue, []*github.Issue, error) {
```
wiggin77, line 41: Seems like the most frequent scenario is with an org, and that path always sets the `queryParamOpenPRQueryArg` etc twice.  Can we reverse it so the most common case is the most performant?

Kshitij-Katiyar: Hey @wiggin77, I have implemented your suggestions and refactored the logic of adding the org check to the query.
Please have a look.

wiggin77: @Kshitij-Katiyar this is an improvement in that the map is not recreated for each org.  However it does not address the issue I mentioned above. The three items that get overwritten in the case of `org != ""` should be reverse such that they are only overwritten when `org == ""`.

wiggin77: (I'm assuming `org != ""` is the most frequent case.)

Kshitij-Katiyar: @wiggin77  

Moving the org check directly into the params map definition and changing `if org != "" {` to `if org == "" {` would complicate handling cases where the org string is empty (possibly requiring regex or similar logic).  

An alternative would be to create two separate params maps—one with the org and one without—but that would increase memory usage.  

Given these trade-offs, the current implementation seems more efficient. Please let us know your thoughts.

wiggin77 (approved):

AayushChaudhary0001:
@Kshitij-Katiyar I was testing this PR and faced a new issue.
While testing this PR without adding any organization, the data was not getting populated in the LHS and no error was received. On further investigation, I found that I need to grant the access of the orgs to my Github's personal access token. I have added the error below as per the server logs. Can we try to provide such error as a direct message from the Github DM bot..



`{"timestamp":"2024-10-23 15:26:11.959 +05:30","level":"warn","msg":"Failed to search for the sidebar data","caller":"app/plugin_api.go:1006","plugin_id":"github","userid":"db84puwc6tn3frubpu1xfim3to","github username":"AayushChaudhary0001","error":"Not able to excute the query: error in executing query: Resource protected by organization SAML enforcement. You must grant your OAuth token access to this organization."}
`

Kshitij-Katiyar:
> @Kshitij-Katiyar I was testing this PR and faced a new issue. While testing this PR without adding any organization, the data was not getting populated in the LHS and no error was received. On further investigation, I found that I need to grant the access of the orgs to my Github's personal access token. I have added the error below as per the server logs. Can we try to provide such error as a direct message from the Github DM bot..
> 
> `{"timestamp":"2024-10-23 15:26:11.959 +05:30","level":"warn","msg":"Failed to search for the sidebar data","caller":"app/plugin_api.go:1006","plugin_id":"github","userid":"db84puwc6tn3frubpu1xfim3to","github username":"AayushChaudhary0001","error":"Not able to excute the query: error in executing query: Resource protected by organization SAML enforcement. You must grant your OAuth token access to this organization."} `

@AayushChaudhary0001 This error can only occur in case the User has PRs in an org that has not been granted access to the OAuth app. It is expected that LHS will show data from the organizations granted access to the Oauth app.
This error log will not cause any issues with the plugin functionality.

arush-vashishtha:
@Kshitij-Katiyar, When the user is part of an external organization and the organization also have granted access to the Oauth application as we can see in the below screenshot. The very first organization has granted access,
Screenshot from 2024-11-07 16-40-00
But the count in the LHS is not counting the events related to that external organization and also when the user opens the list of event it is not showing those events.
image
image

Kshitij-Katiyar:
@arush-vashishtha You can enable the below configuration related to private repo and hopefully get to see the PRs you want
Screenshot from 2024-11-11 16-45-54

arush-vashishtha:
> @arush-vashishtha You can enable the below configuration related to private repo and hopefully get to see the PRs you want Screenshot from 2024-11-11 16-45-54

Thank you, it is working now.

arush-vashishtha:
This PR has been tested and the desired functionality LGTM. The user is getting the count for all the organizations in the LHS when the organization lock is not set. For external organizations, the user must enable the settings discussed above.

PR #849, opened 2024-11-18
Test: Fix playright test

raghavaggarwal2308:


wiggin77 (changes requested):
Still not working.  May be a new unrelated issue.

PR #855, opened 2024-11-25
[MM-956]: Added server testcase for webhook.go

Kshitij-Katiyar:
#### Summary
Added server testcase for webhook.go


wiggin77 (approved):

server/plugin/webhook_test.go
```diff
@@ -0,0 +1,869 @@
...
+	"github.com/golang/mock/gomock"
+	"github.com/google/go-github/v54/github"
+	"github.com/mattermost/mattermost/server/public/model"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+)
+
+func TestPostPushEvent(t *testing.T) {
+	mockKvStore, mockAPI, _, _, _ := GetTestSetup(t)
```
raghavaggarwal2308, line 15: move inside, as it can cause some issues later

raghavaggarwal2308, line 15:
```suggestion
	mockKVStore, mockAPI, _, _, _ := GetTestSetup(t)
```

server/plugin/webhook_test.go
```diff
@@ -0,0 +1,869 @@
...
+				mockAPI.On("CreatePost", mock.Anything).Return(nil, &model.AppError{Message: "error creating post"}).Times(1)
+				mockAPI.On("LogWarn", "Error webhook post", "post", mock.Anything, "error", "error creating post")
+			},
+		},
+		{
+			name:      "Successful handle post push event",
+			pushEvent: GetMockPushEvent(),
+			setup: func() {
+				mockKvStore.EXPECT().Get(SubscriptionsKey, gomock.Any()).DoAndReturn(func(key string, value interface{}) error {
+					if v, ok := value.(**Subscriptions); ok {
```
raghavaggarwal2308, line 61: Why are we using double * here?

server/plugin/webhook_test.go
```diff
@@ -0,0 +1,869 @@
...
+					}
+					return nil
+				}).Times(1)
+			},
+		},
+		{
+			name:      "Error creating post",
+			pushEvent: GetMockPushEvent(),
+			setup: func() {
+				mockKvStore.EXPECT().Get(SubscriptionsKey, gomock.Any()).DoAndReturn(func(key string, value interface{}) error {
```
raghavaggarwal2308, line 46: Create separate functions for the nested one. Its getting very complicated here

server/plugin/webhook_test.go
```diff
@@ -0,0 +1,869 @@
...
+	mockKvStore, mockAPI, _, _, _ := GetTestSetup(t)
+	p := getPluginTest(mockAPI, mockKvStore)
+
+	tests := []struct {
+		name        string
+		createEvent *github.CreateEvent
+		setup       func()
+	}{
+		{
+			name:        "no subscription found",
```
raghavaggarwal2308, line 91:
```suggestion
			name:        "No subscription found",
```
For consistency, check everywhere

raghavaggarwal2308 (changes requested):

server/plugin/webhook_test.go
Kshitij-Katiyar: Because pointer of pointer is used here
Screenshot from 2025-01-28 15-54-37

Kshitij-Katiyar:
@raghavaggarwal2308 FIxed the comments, please re-review

server/plugin/webhook_test.go
raghavaggarwal2308: @Kshitij-Katiyar I do not think we need the double pointer here.

raghavaggarwal2308 (approved):

server/plugin/webhook_test.go
Kshitij-Katiyar: @raghavaggarwal2308 Double-pointers are very much required here to handle the case of `no subscription stored`
Using single pointer can make the code more complicated but we can do it.
Let me know your thoughts on this

PR #869, opened 2025-01-21
Apply and enforce file-level license headers

lieut-data:
#### Summary
Use `mattermost-govet` and `eslint`'s `header/header` to ensure each file begins with the appropriate header, then fix all instances.

```
// Copyright (c) 2018-present Mattermost, Inc. All Rights Reserved.
// See LICENSE.txt for license information.
```

Note that this renames `LICENSE` to `LICENSE.txt` purely to align on the copyright messaging across repos.

#### Ticket Link
None


.github/workflows/ci.yml
```diff
@@ -16,3 +16,5 @@ jobs:
   plugin-ci:
     uses: mattermost/actions-workflows/.github/workflows/plugin-ci.yml@main
     secrets: inherit
+    with:
+      golang-version: "1.22"
```
lieut-data, line 20: This change is unique in all the others in that I couldn't get `mattermost-govet` as built to run with `go vet`, it seems because the toolchain version installed was different than `go.mod`.

Aligning this here fixed the issue... but that technically potentially changes the Go version we're building the core plugin with (even though it should have been that way), and I'm surfacing here for visibility.

wiggin77 (approved):
LGTM 👍

PR #886, opened 2025-03-12
Fixed issue of deletion of non-existent subscription

Kshitij-Katiyar:
#### Summary
Fixed the issue of deletion of a non-existent subscription

##### Old
Screenshot from 2025-03-12 17-40-00
##### New
Screenshot from 2025-03-12 17-35-27

### DIFF
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -599,6 +599,10 @@ func (p *Plugin) handleUnsubscribe(_ *plugin.Context, args *model.CommandArgs, p
 	owner = strings.ToLower(owner)
 	repo = strings.ToLower(repo)
 	if err := p.Unsubscribe(args.ChannelId, repo, owner); err != nil {
+		if strings.Contains(err.Error(), "no subscription exist") {
+			return err.Error()
+		}
+
 		p.client.Log.Warn("Failed to unsubscribe", "repo", repo, "error", err.Error())
 		return "Encountered an error trying to unsubscribe. Please try again."
 	}
--- a/server/plugin/subscriptions.go
+++ b/server/plugin/subscriptions.go
@@ -17,11 +17,12 @@ import (
 )
 
 const (
-	SubscriptionsKey      = "subscriptions"
-	flagExcludeOrgMember  = "exclude-org-member"
-	flagRenderStyle       = "render-style"
-	flagFeatures          = "features"
-	flagExcludeRepository = "exclude"
+	SubscriptionsKey        = "subscriptions"
+	flagExcludeOrgMember    = "exclude-org-member"
+	flagRenderStyle         = "render-style"
+	flagFeatures            = "features"
+	flagExcludeRepository   = "exclude"
+	SubscriptionUnavailable = "no subscription exist for `%s` in the channel"
 )
 
 type SubscriptionFlags struct {
@@ -388,7 +389,7 @@ func (p *Plugin) Unsubscribe(channelID, repo, owner string) error {
 
 	repoSubs := subs.Repositories[repoWithOwner]
 	if repoSubs == nil {
-		return nil
+		return errors.Errorf(SubscriptionUnavailable, strings.TrimSuffix(repoWithOwner, "/"))
 	}
 
 	removed := false
@@ -405,6 +406,8 @@ func (p *Plugin) Unsubscribe(channelID, repo, owner string) error {
 		if err := p.StoreSubscriptions(subs); err != nil {
 			return errors.Wrap(err, "could not store subscriptions")
 		}
+	} else {
+		return errors.Errorf(SubscriptionUnavailable, strings.TrimSuffix(repoWithOwner, "/"))
 	}
 
 	return nil


server/plugin/subscriptions.go
```diff
@@ -17,11 +17,12 @@ import (
...
-	flagExcludeOrgMember  = "exclude-org-member"
-	flagRenderStyle       = "render-style"
-	flagFeatures          = "features"
-	flagExcludeRepository = "exclude"
+	SubscriptionsKey        = "subscriptions"
+	flagExcludeOrgMember    = "exclude-org-member"
+	flagRenderStyle         = "render-style"
+	flagFeatures            = "features"
+	flagExcludeRepository   = "exclude"
+	SubscriptionUnavailable = "no subscription exist for `%s` in the channel"
```
wiggin77, line 25:
```suggestion
	SubscriptionUnavailable = "no subscription exists for `%s` in the channel"
```

server/plugin/command.go
```diff
@@ -599,6 +599,10 @@ func (p *Plugin) handleUnsubscribe(_ *plugin.Context, args *model.CommandArgs, p
 	owner = strings.ToLower(owner)
 	repo = strings.ToLower(repo)
 	if err := p.Unsubscribe(args.ChannelId, repo, owner); err != nil {
+		if strings.Contains(err.Error(), "no subscription exist") {
```
wiggin77, line 602:
```suggestion
		if strings.Contains(err.Error(), "no subscription exists") {
```

wiggin77 (approved):

webapp/package.json
```diff
@@ -99,6 +100,7 @@
       "text-summary"
     ],
     "moduleNameMapper": {
+      "^@/(.*)$": "<rootDir>/src/$1",
```
hanzei, line 103: Why is this change needed?

webapp/src/components/modals/create_issue/__snapshots__/create_issue.test.jsx.snap
```diff
@@ -57,7 +57,6 @@ exports[`CreateIssueModal should render correctly with default props 1`] = `
           value={null}
         />
         <Input
-          disabled={false}
```
hanzei, line 60: I'm surprised to see that the snap was not up to date. Do we need CI check to ensure that?

server/plugin/subscriptions.go
```diff
@@ -405,6 +406,8 @@ func (p *Plugin) Unsubscribe(channelID, repo, owner string) error {
 		if err := p.StoreSubscriptions(subs); err != nil {
 			return errors.Wrap(err, "could not store subscriptions")
 		}
+	} else {
+		return errors.Errorf(SubscriptionUnavailable, strings.TrimSuffix(repoWithOwner, "/"))
 	}
```
hanzei, line 411: 2/5 that the code is easier to read if we invert the `if` and return early.

server/plugin/command.go
```diff
@@ -600,6 +600,10 @@ func (p *Plugin) handleUnsubscribe(_ *plugin.Context, args *model.CommandArgs, p
 	owner = strings.ToLower(owner)
 	repo = strings.ToLower(repo)
 	if err := p.Unsubscribe(args.ChannelId, repo, owner); err != nil {
+		if strings.Contains(err.Error(), "no subscription exists") {
+			return err.Error()
+		}
```
hanzei, line 605: I don't like the implied checks for errors. Can we use a dedicated error type and check that instead?

server/plugin/subscriptions.go
hanzei, line 1: Can we add tests that cover this change?

hanzei (changes requested):

webapp/src/components/modals/create_issue/__snapshots__/create_issue.test.jsx.snap
Kshitij-Katiyar: I don't think we need a CI check for the same, the web app test cases will fail if the snapshot is not updated. I wonder how this was not already on master

webapp/package.json
Kshitij-Katiyar: Jest was unable to resolve the imports after we changed the default import paths

server/plugin/subscriptions.go
Kshitij-Katiyar: We don't have existing test cases for unsubscribe/commands. We can surely add the test cases but it will take some time. Should we add the testcase in this PR or create an issue for the test cases and add them in another PR?

abbas-dependable-naqvi: @wiggin77 We are planning to add this PR to this release. We can log a new issue for the E2E testcases and plan them for future.

abbas-dependable-naqvi: @wiggin77 As discussed merging this PR and creating a Github issue for the E2E testcases

Yash-Chakerverti: Here's the issue for the same:-
https://github.com/mattermost/mattermost-plugin-github/issues/936

PR #892, opened 2025-03-24
Fix the "Attach to Github issue" feature to work without organisational lock

Kshitij-Katiyar:
#### Summary
Fix the "Attach to Github issue" feature so it works without the organisational lock.
- When the plugin is not set for any organisation then it will fetch all the issues which involve the user

### DIFF
--- a/server/plugin/utils.go
+++ b/server/plugin/utils.go
@@ -40,14 +40,16 @@ func getYourAssigneeSearchQuery(username string, orgs []string) string {
 	return buildSearchQuery("is:open assignee:%v archived:false %v", username, orgs)
 }
 
-func getIssuesSearchQuery(org, searchTerm string) string {
+func getIssuesSearchQuery(searchValue, searchTerm string) string {
 	query := "is:open is:issue archived:false %v %v"
-	orgField := ""
-	if len(org) != 0 {
-		orgField = fmt.Sprintf("org:%v", org)
+	searchField := ""
+	if len(searchValue) != 0 {
+		searchField = fmt.Sprintf("org:%v", searchValue)
+	} else { // get all the issues which involve the user in case no organisaitonal lock is set 
+		searchField = fmt.Sprintf("involves:@me")
 	}
 
-	return fmt.Sprintf(query, orgField, searchTerm)
+	return fmt.Sprintf(query, searchField, searchTerm)
 }
 
 func buildSearchQuery(query, username string, orgs []string) string {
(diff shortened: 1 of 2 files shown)


server/plugin/utils.go
```diff
@@ -40,14 +40,16 @@ func getYourAssigneeSearchQuery(username string, orgs []string) string {
...
+func getIssuesSearchQuery(searchValue, searchTerm string) string {
 	query := "is:open is:issue archived:false %v %v"
-	orgField := ""
-	if len(org) != 0 {
-		orgField = fmt.Sprintf("org:%v", org)
+	searchField := ""
+	if len(searchValue) != 0 {
+		searchField = fmt.Sprintf("org:%v", searchValue)
+	} else { // get all the issues which involve the user in case no organisaitonal lock is set 
+		searchField = fmt.Sprintf("involves:@me")
```
raghavaggarwal2308, line 49: What if the user is commenting for the first time on that PR?

server/plugin/utils.go
```diff
@@ -40,14 +40,16 @@ func getYourAssigneeSearchQuery(username string, orgs []string) string {
...
-func getIssuesSearchQuery(org, searchTerm string) string {
+func getIssuesSearchQuery(searchValue, searchTerm string) string {
 	query := "is:open is:issue archived:false %v %v"
-	orgField := ""
-	if len(org) != 0 {
-		orgField = fmt.Sprintf("org:%v", org)
+	searchField := ""
+	if len(searchValue) != 0 {
+		searchField = fmt.Sprintf("org:%v", searchValue)
+	} else { // get all the issues which involve the user in case no organisaitonal lock is set
```
wiggin77, line 48: Typo
```suggestion
	} else { // get all the issues which involve the user in case no organisational lock is set
```

server/plugin/utils.go
```diff
@@ -40,14 +40,16 @@ func getYourAssigneeSearchQuery(username string, orgs []string) string {
...
-func getIssuesSearchQuery(org, searchTerm string) string {
+func getIssuesSearchQuery(searchValue, searchTerm string) string {
 	query := "is:open is:issue archived:false %v %v"
-	orgField := ""
-	if len(org) != 0 {
-		orgField = fmt.Sprintf("org:%v", org)
+	searchField := ""
+	if len(searchValue) != 0 {
+		searchField = fmt.Sprintf("org:%v", searchValue)
+	} else { // get all the issues which involve the user in case no organisational lock is set
```
wiggin77, line 48:
```suggestion
	} else { // get all the issues which involve the user in case no organizational lock is set
```

server/plugin/utils.go
Kshitij-Katiyar: @raghavaggarwal2308

raghavaggarwal2308: I think this was implemented before. We are fetching the issues globally on the basis of the search param.

PR #925, opened 2025-08-21
Add plugin bot id check to reaction mapper

abbas-dependable-naqvi:
#### Summary
Add plugin bot id check to reaction mapper


abbas-dependable-naqvi:
@wiggin77 I tried thinking of any other security issue in this feature, but i cant think of any.

PR #995, opened 2026-04-09
Limit webhook request body size with http.MaxBytesReader

nevyangelova:
#### Summary
Add http.MaxBytesReader to the webhook handler to cap incoming request bodies at 25 MB (matching GitHub's documented maximum webhook payload size). This ensures oversized requests are rejected immediately with HTTP 413 before being read into memory, regardless of any upstream transport-layer protections.

This is a defense-in-depth hardening measure — the Mattermost plugin IPC layer already prevents exploitation, but the code should not rely on that external protection.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68166


edgarbellot (approved):
@nevyangelova LGTM, thank you Nevy!

avasconcelos114 (approved):
LGTM! 🚀

PR #997, opened 2026-04-20
MM-68363: redact post contents from error logs to prevent sensitive data exposure

nang2049:
#### Summary
Replace all instances of logging the full model.Post struct in error log statements with only non-sensitive metadata. Previously, when a post creation failed, the entire post was logged which could expose sensitive or private user content in Grafana.

#### QA Test Steps
Set up a GitHub plugin instance connected to a test repository with channel subscriptions.
Trigger a webhook event that would create a post.
Cause the notification to fail for example, by making the message too long so the server rejects it.
Check the server logs (Grafana or local logs) for the resulting warn-level log entry. ( can help with that )
Verify the log entry contains channel_id and error fields but does not contain a post field with message content.

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-68363

### DIFF
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -401,7 +401,7 @@ func (p *Plugin) createPost(channelID, userID, message string) error {
 	}
 
 	if err := p.client.Post.CreatePost(post); err != nil {
-		p.client.Log.Warn("Error while creating post", "post", post, "error", err.Error())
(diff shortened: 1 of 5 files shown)


avasconcelos114 (approved):

nang2049:
@coderabbitai review

ogi-m (approved):

PR #998, opened 2026-04-27
Added handling for cases where the user in a PR review is missing

avasconcelos114:
#### Summary

There is an edge case where if an account that previously reviewed the PR is deleted, the list of PRs in the sidebar will fail to load as it attempts to read the user attribute for the review.

This PR combats this by both filtering the PR reviews that don't have an user attached to them, as well as adding some defensive coding on the webapp to also prevent this from affecting the rendering of the sidebar contents

| Before | After |
| :--: | :--: |
| image | image |

#### QA Notes
Reproducing this naturally is somewhat difficult given that you need to have given a review to a PR with a temporary account that is then deleted. The easiest way to confirm that the fix is working would be to use a browser extension to mock the response of `/plugins/github/api/v1/prsdetails` and remove the `user` object from one of the reviews.

Once that is done just connect your account and open the "Your Open Pull Requests" sidebar from the LHS menu buttons

## Change Impact: 🟡 Medium

**Reasoning:** These changes span both backend and frontend to handle the edge case of missing review authors, and include a type definition change to optionality. While the changes are isolated to review handling and are purely defensive, the scope across layers and the modification to the data model warrant medium-level scrutiny.

**Regression Risk:** Low. The changes add null-checking guards before accessing potentially undefined properties and filter invalid reviews server-side before transmission. No existing behavioral logic is modified; these are additive defensive measures that prevent crashes in an edge case (deleted/missing user accounts in PR reviews) without altering normal operation.

**QA Recommendation:** Light manual QA is recommended. Primary verification: confirm the "Your Open Pull Requests" sidebar loads successfully when a PR contains a review from a deleted/missing user account. The PR provides browser dev tools instructions to mock the edge case. Risk of skipping QA is medium—while changes are defensive and low-regression, the sidebar is user-facing and the edge case being fixed is real. A quick smoke test of sidebar loading with typical PRs is sufficient for low-risk release.

### DIFF
--- a/server/plugin/api.go
+++ b/server/plugin/api.go
@@ -859,7 +859,15 @@ func fetchReviews(c *UserContext, client *github.Client, repoOwner string, repoN
 		return []*github.PullRequestReview{}, errors.Wrap(err, "could not list reviews")
 	}
 
-	return reviewsList, nil
+	// Filtering reviews where user is nil
+	filtered := make([]*github.PullRequestReview, 0, len(reviewsList))
+	for _, review := range reviewsList {
+		if review != nil && review.User != nil {
+			filtered = append(filtered, review)
+		}
+	}
+
+	return filtered, nil
 }
 
 func getRepoOwnerAndNameFromURL(rawURL string) (string, string, error) {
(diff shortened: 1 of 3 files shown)


nang2049 (approved):
LGTM :)

avasconcelos114:
@ogi-m Since this is some defensive coding that prevents an issue that personally affects my account, I'll merge this and confirm that the RC build for 2.7.1 also resolves the issue as my test build did

PR #999, opened 2026-04-28
MM-68539: SLA digest fixes

jgheithcock:
## Summary

This branch hardens the daily PR-review SLA digest in three ways:

**Coverage.** The digest now scans every open non-draft PR in each configured org via a single GraphQL search, instead of running a per-user search for each connected Mattermost user (the prior approach silently skipped reviewers who hadn't connected their GitHub account). Direct user-requests and team-requests are flattened and de-duplicated; team membership is fetched once per digest run.

**Correctness.** The SLA clock now backfills from the PR timeline whenever the plugin doesn't have a cached `review_requested` webhook record — for example because the PR was opened before the plugin was installed, or because the most-recent request preceded a Mattermost restart. Without this, PRs whose pending review was actually a few days old were showing up as "86 days overdue" against the configured target. Backfilled times are written back to KV so subsequent digest runs hit the fast path.

**Presentation.** The digest groups overdue items into seven fixed buckets (1–3 / 4–7 / 8–14 / 15–30 / 31–90 / 91–365 / >1y), renders PR titles as markdown links (with `]` / `\` escaping for JIRA-prefixed names), and renders each reviewer as their actual Mattermost @-mention via `getGitHubToUsernameMapping`. Reviewers without a Mattermost account render as `(not connected) - <login>` so admins can see who's missing without producing dead @-mentions. The configured SLA target appears in the post header. The RHS panel grew per-item SLA badges (Overdue / Due today / Due in N days) sharing math with the existing sidebar red-button indicator via a new `webapp/src/utils/sla` module.

The final two commits in this branch decompose `collectAllOverdueSLAItems` into small testable helpers and fix two latent bugs uncovered while doing so:

1. **Concurrent GraphQL state.** `orgOpenPRsQuery` was a package-level `var` that the graphql library decoded into. Concurrent callers — the digest scheduler running alongside an LHS fetch on a different goroutine — could clobber each other's `Search.Nodes` mid-decode. Now defined as a type at package scope (so the library can still reflect on its tags) but instantiated locally per call.
2. **Misleading backfill summary log.** The backfill summary log was deferred *inside* the resolver constructor, so it fired immediately when the closure was created and always reported `timeline_pages_fetched=0 kv_hits=0` regardless of actual usage. Now wired up via a returned `summarize()` callback the digest invokes via defer; counters log as plain ints (not `Sprintf`'d strings) so structured-log backends keep them numeric; the resolver isn't constructed at all on degenerate runs (no PRs, no orgs configured) so the summary line only appears when the digest actually had work to do.

## Ticket Link

https://mattermost.atlassian.net/browse/MM-68539

## For sysadmins: SLA backfill summary log

Once per local-day digest run, the plugin emits one Info line:

```
[INFO] SLA digest backfill summary  timeline_pages_fetched=N  kv_hits=M
```

- `timeline_pages_fetched` — distinct PRs whose GitHub issue timeline the digest had to pull because no cached SLA-start was on file.
- `kv_hits` — `(PR, reviewer)` pairs whose SLA-start time came from KV (either from a live `review_requested` webhook or a previous run's backfill).

Together these tell you whether the SLA backfill is healthy. The line lives at Info on purpose, so admins don't have to enable Debug logging just to spot trouble.

### What healthy looks like

| Day | `timeline_pages_fetched` | `kv_hits` | Meaning |
|---|---|---|---|
| 1 (fresh install) | ~equal to # of open PRs with reviewers | low | Cold cache; almost everything backfilled from the timeline. |
| 2 | small | rises sharply | Yesterday's backfilled times now cached. |
| 14+ (steady state) | 1–2 | dominates | Only new PRs since last run need backfilling. |

The shape to internalize: **after a few days, `kv_hits` should dominate; `timeline_pages_fetched` tracks new PR activity since the last run.**

### What a problem looks like

| Pattern | Likely cause | Where to look next |
|---|---|---|
| `timeline_pages_fetched` stays high day after day, `kv_hits` near 0 | Backfilled times aren't being persisted | Look for `Failed to store review SLA start time` Warn entries in the same log window |
| Same shape, no warn logs about KV writes | Timeline returning nothing for every reviewer (e.g. all events are `review_request_removed`, GitHub schema drift, scope/permission issue on the service user's token) | Enable Debug; look for `SLA digest timeline fetch failed` |
| `timeline_pages_fetched` >> count of open PRs in your configured org(s) | Per-run cache is broken, or overlapping orgs are double-walking the same PRs | Cross-check `OrganizationsList` config; report as a bug |
| Line never appears | Digest didn't actually run | Look for `SLA digest cannot run` warnings (no orgs configured / no service user available / no PRs found) |

### Quick triage when "the days-overdue numbers look wrong"

1. Find the most recent `SLA digest backfill summary` line in the server log.
2. Is `kv_hits` proportional to the number of overdue items in the channel post? If not, the cache isn't being used → days-overdue is falling back to PR-open dates → numbers are inflated.
4. Is `timeline_pages_fetched` plausible vs. the open-PR count? If it's far above, the per-run cache is broken; if it stays high day after day with no `kv_hits`, the writeback is broken.
5. Cross-reference with `SLA digest timeline fetch failed` Debug entries (need Debug logging enabled to see those) to separate "fetch failing" from "fetch returning nothing."

## Test plan

- [x] `go test ./server/...` passes
- [x] `go vet ./server/...` clean
- [x] No new linter errors
- [x] New test `TestNewDigestSLAStartResolver_LogsAccurateCounters` locks in both invariants of the summary-log fix (no fire at construction time, exactly one fire when `summarize()` is called, with correct counters)
- [x] New tests for `prRefFromIssue`, `gatherReviewersForPR`, and `pickServiceGitHubUser` (deterministic ordering, broken-token skip, no-connected-users)
- [ ] Sanity-check on a real Mattermost install: run the digest, confirm a non-zero summary line appears in the server log
- [ ] Sanity-check after a second run: confirm `kv_hits` rises and `timeline_pages_fetched` drops

## Change Impact: 🔴 High

**Reasoning:** This PR makes broad, cross-cutting changes to the SLA digest system: it replaces per-user REST searches with an org-wide GraphQL discovery path, introduces timeline-based SLA backfill and KV persistence, and updates server and webapp presentation (including new SLA badges). The work spans multiple core modules and alters data-gathering, persistence, and user-facing behavior, so the blast radius and potential for subtle failures are large despite added unit tests.

**Regression Risk:** High — the new GraphQL-based discovery, pagination, timeline backfill, and KV caching touch core data paths and shared utilities; incorrect pagination, timeline interpretation, or KV handling could lead to missed or misattributed SLA anchors and UI/display inconsistencies. Although extensive tests were added, there are still integration risks across server↔client boundaries and failure/cancellation paths.

**QA Recommendation:** Strong manual QA and integration testing are recommended. Validate org-wide PR discovery with mixed direct/team requests, SLA backfill correctness on PRs lacking webhook history, KV persistence/resilience across restarts, correct reviewer rendering for connected vs unconnected users, SLA-badge states in the RHS and sidebar, and failure/cancellation behaviors (ensure the daily marker isn’t advanced on incomplete scans). Automated end-to-end tests covering GraphQL → digest → UI rendering should be added before broad rollout.

*Generated by CodeRabbitAI*

### DIFF
--- a/server/plugin/sla_digest.go
+++ b/server/plugin/sla_digest.go
@@ -104,12 +104,17 @@ func (p *Plugin) resolveReviewerDisplayName(githubLogin string) string {
 }
 
 // pickServiceGitHubUser returns the first connected user the digest can use as a service
-// caller for org-wide GitHub queries. Skips users whose token cannot be loaded. Returns nil
-// when no connected user is available; the digest cannot run in that case.
+// caller for org-wide GitHub queries. Keys are sorted before iteration so the choice is
+// deterministic across runs (and across cluster nodes) — otherwise the digest's visibility
+// into private repos/teams could silently shift run-to-run with KV iteration order.
+//
+// Returns nil when no connected user is available; the digest cannot run in that case.
 func (p *Plugin) pickServiceGitHubUser(ctx context.Context) *GitHubUserInfo {
 	checker := func(key string) (bool, error) {
 		return strings.HasSuffix(key, githubTokenKey), nil
 	}
+
+	var allKeys []string
 	for page := 0; ; page++ {
 		if ctx.Err() != nil {
 			return nil
@@ -119,17 +124,23 @@ func (p *Plugin) pickServiceGitHubUser(ctx context.Context) *GitHubUserInfo {
 			p.client.Log.Warn("SLA digest failed to list connected users", "error", err.Error())
 			return nil
 		}
-		for _, key := range keys {
-			userID := strings.TrimSuffix(key, githubTokenKey)
-			ghInfo, apiErr := p.getGitHubUserInfo(userID)
-			if apiErr != nil || ghInfo == nil {
-				continue
-			}
-			return ghInfo
-		}
+		allKeys = append(allKeys, keys...)
 		if len(keys) < keysPerPage {
 			break
 		}
+	}
+	sort.Strings(allKeys)
+
+	for _, key := range allKeys {
+		if ctx.Err() != nil {
+			return nil
+		}
+		userID := strings.TrimSuffix(key, githubTokenKey)
+		ghInfo, apiErr := p.getGitHubUserInfo(userID)
+		if apiErr != nil || ghInfo == nil {
+			continue
+		}
+		return ghInfo
 	}
 	return nil
 }
@@ -137,7 +148,6 @@ func (p *Plugin) pickServiceGitHubUser(ctx context.Context) *GitHubUserInfo {
 func (p *Plugin) collectAllOverdueSLAItems(ctx context.Context) []slaDigestEntry {
 	config := p.getConfiguration()
 	targetDays := config.ReviewTargetDays
-	baseURL := config.getBaseURL()
 	orgList := config.getOrganizations()
 	now := time.Now()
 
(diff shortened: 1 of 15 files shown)


server/plugin/sla_digest.go
```diff
@@ -87,110 +96,406 @@ func (p *Plugin) maybePostDailyOverdueSLADigest(ctx context.Context) {
...
+
+// gatherReviewersForPR collapses a PR's directly-requested users and all members of its
+// requested teams into a deduplicated reviewer set, preserving each reviewer's team origin so
+// the SLA backfill can use a team-scoped review_requested event when no user-scoped event is
+// available. Logins are matched case-insensitively for deduplication; the first-seen casing is
+// preserved on the returned struct.
+func gatherReviewersForPR(pr graphql.DigestPR, resolveTeam func(graphql.DigestTeamRef) []string) []reviewerRequest {
+	byLogin := make(map[string]int)
+	out := make([]reviewerRequest, 0, len(pr.RequestedUsers))
+	add := func(login string, team *graphql.DigestTeamRef) {
```
nang2049, line 304: Nit: If a team resolver returns an empty string login here it's silently skipped by the if login == "" guard but the same guard doesnt apply to pr.RequestedUsers. A webhook returning an empty string user login would create an entry with Login: ""

nang2049 (approved):
LGTM one nitpick!

server/plugin/sla_digest.go
jgheithcock: [95ffb55](https://github.com/mattermost/mattermost-plugin-github/pull/999/commits/95ffb55be2c844ac86ea9d3807733ffb3757b52d) adds a test and a comment to resolve this. Let me know if I am missing something.

PR #1010, opened 2026-05-15
MM-68829 Improving UX for DMs sent on PR comments

avasconcelos114:
#### Summary

This PR mainly does two things to improve the general experience of receiving DM notifications from the GitHub bot
- Preventing edit or delete PR message events from sending a "new comment" DM notification
- Improving logic that cleans body to prevent issues with html comments in the message from interfering with visibility of message

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-68829

#### QA Notes
The main webhook event affected by this is the PR commenting, which mainly consists of:
- Notification received by someone who was mentioned in a PR comment (that is neither the author nor the user posting the comment)
- Notification received by the author of the PR 

We'll want to make sure that editing a PR comment does not send a new notification to the users as a DM

We might also want to run some regression tests on the following DM notifications to confirm they still append a quote with the body of the comment as well:
- commentMentionNotification — DMs users @mentioned in an issue
- commentAuthorIssueNotification — DMs the issue author when someone comments on their issue
- commentAssigneePullRequestNotification — DMs PR assignees when someone comments on the PR conversation
- commentAssigneeIssueNotification — DMs issue assignees when someone comments on the issue
- commentAssigneeSelfMentionPullRequestNotification — DMs PR assignees who were also @mentioned in the comment
- commentAssigneeSelfMentionIssueNotification — DMs issue assignees who were also @mentioned in the comment

## Change Impact: 🟡 Medium

**Reasoning:** The PR adds new DM notification handlers for PR review comments and updates message-body cleaning logic used across multiple DM templates; these are user-facing changes spanning several files (webhook handling, templates, and tests) but remain focused on notification behavior.

**Regression Risk:** Moderate. Early-return guards reduce unwanted DMs, but new code paths (user mapping, permission checks, muted-user filtering, and DM posting) introduce failure points that could silently suppress or misroute notifications. The new cleanBody helper changes message rendering for many templates; while templates were updated, dedicated tests for all template output permutations appear limited.

** QA Recommendation:** Perform targeted manual QA on PR comment workflows: verify that editing or deleting PR comments no longer sends "new comment" DMs, and confirm quoted comment bodies are correctly cleaned/omitted for the affected notification types (commentMentionNotification, commentAuthorIssueNotification, commentAssigneePullRequestNotification, commentAssigneeIssueNotification, commentAssigneeSelfMentionPullRequestNotification, commentAssigneeSelfMentionIssueNotification). Automated tests cover many scenarios but manual verification is recommended for edge cases (HTML comments, GitHub footer removal, and permission/mapping failures).

*Generated by CodeRabbitAI*

[Review Change Stack](https://app.coderabbit.ai/change-stack/mattermost/mattermost-plugin-github/pull/1010?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/server/plugin/template.go
+++ b/server/plugin/template.go
@@ -75,9 +75,6 @@ func init() {
 		}
 		cleaned = mdCommentRegex.ReplaceAllString(cleaned, "")
 		cleaned = strings.TrimSpace(cleaned)
-		if cleaned == "" {
-			return ""
-		}
 		return cleaned
 	}
 
--- a/server/plugin/template.go
+++ b/server/plugin/template.go
@@ -68,6 +68,19 @@ func init() {
 		return mdCommentRegex.ReplaceAllString(body, "")
 	}
 
+	funcMap["cleanBody"] = func(body string) string {
+		cleaned := body
+		if strings.Contains(cleaned, "notifications@github.com") {
+			cleaned = strings.Split(cleaned, "\n\nOn")[0]
+		}
+		cleaned = mdCommentRegex.ReplaceAllString(cleaned, "")
+		cleaned = strings.TrimSpace(cleaned)
+		if cleaned == "" {
+			return ""
+		}
+		return cleaned
+	}
+
 	// Replace any GitHub username with its corresponding Mattermost username, if any
 	funcMap["replaceAllGitHubUsernames"] = func(body string) string {
 		return gitHubUsernameRegex.ReplaceAllStringFunc(body, func(matched string) string {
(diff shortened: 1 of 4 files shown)


server/plugin/webhook.go
```diff
@@ -1077,6 +1083,97 @@ func (p *Plugin) postPullRequestReviewCommentEvent(event *github.PullRequestRevi
 	}
 }
 
+func (p *Plugin) handleReviewCommentMentionNotification(event *github.PullRequestReviewCommentEvent) {
```
nang2049, line 1086: handleCommentMentionNotification filters out mentioned users who are also assignees. Maybe worth mirroring the skip logic for consistency and to avoid double DM?

server/plugin/webhook.go
```diff
@@ -1205,6 +1302,11 @@ func (p *Plugin) handleCommentAuthorNotification(event *github.IssueCommentEvent
 }
 
 func (p *Plugin) handleCommentAssigneeNotification(event *github.IssueCommentEvent) {
+	action := event.GetAction()
+	if action == actionEdited || action == actionDeleted {
```
nang2049, line 1306: We can check if handleCommentAssigneeNotification caller already filters out edit and delete actions. If yes this guard is dead code.

server/plugin/template.go
```diff
@@ -68,6 +68,19 @@ func init() {
...
 	}
 
+	funcMap["cleanBody"] = func(body string) string {
+		cleaned := body
+		if strings.Contains(cleaned, "notifications@github.com") {
+			cleaned = strings.Split(cleaned, "\n\nOn")[0]
+		}
+		cleaned = mdCommentRegex.ReplaceAllString(cleaned, "")
+		cleaned = strings.TrimSpace(cleaned)
+		if cleaned == "" {
```
nang2049: nit: this if cleaned == "" { return "" } block returns "" either way. Just return cleaned

nang2049 (commented):
LGTM otherwise :)

server/plugin/webhook.go
avasconcelos114: The caller (`handleWebhook`) doesn't filter out by action, this has been done for each individual handler previously but this handler in particular was missing it

I'm wondering if we want to move the guard up to `handleWebhook` but not entirely certain as this would deviate from how every other webhook event is processed (specific action-related guards happening at the handler function-level), what do you think works best here?

server/plugin/webhook.go
avasconcelos114: For the `PullRequestReviewCommentEvent` events there isn't an assignee handler like we do for `IssueCommentEvent`, only one handler for the author which this function is skipping in`handleReviewCommentMentionNotification:1112` to avoid duplication, so I think in here we're safe from the risk of double DMing here

nang2049 (approved):
LGTM!

PR #1022, opened 2026-06-15
MM-69266: Fall back to org-level webhooks for repo subscriptions

ewwollesen:
#### Summary

When subscribing a channel to a **repository** (`/github subscriptions add owner/repo`), `checkIfConfiguredWebhookExists` only listed **repository-level** hooks. If the webhook was instead configured at the **organization level** — which still delivers events for that repo — the check found no match and the subscription confirmation wrongly appended:

> **Note:** No webhook was found for this repository or organization. To create one, enter the following slash command `/github setup webhook`

…even though a working org-level webhook was already in place.

#### Fix

Refactor the lookup into a small paginated helper (`findWebhookForSiteURL`) called twice:

1. **Repo subscription** → check repository-level hooks first. If a match is found, or the call errors, return immediately (preserving the existing `404 Not Found` → suppress-message behavior for users without repo-admin access).
2. If the repo list returns cleanly but matches nothing → **fall back to organization-level hooks**.
3. **Org subscription** (`repo == ""`) → unchanged: queries org hooks directly, error returned unwrapped so the existing `404` string-match suppression still applies.

The org-level fallback is **best-effort**: listing org hooks requires org-admin scope that a repo subscriber often lacks (403/404). Any error there is logged at debug and swallowed, falling back to "not found" rather than propagating — so the worst case is identical to pre-fix behavior and a successful subscription is never turned into an error message.

Also tightened the webhook config URL read to a checked type assertion (was an unchecked `.(string)` that could panic on a malformed hook config).

#### Tests

Adds `server/plugin/command_webhook_test.go` covering `checkIfConfiguredWebhookExists` (previously untested), using the repo's existing `httptest.NewServer` + `client.BaseURL` mocking pattern:

- repo-level webhook present
- repo-level absent but org-level present (the bug)
- absent at both repo and org level
- org-level listing forbidden (403) is swallowed
- org-level listing 404 is swallowed
- org subscription: webhook present
- org subscription: error propagated for caller suppression

#### Ticket(s)

- Jira: MM-69266
- Support ticket: 51590

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Change Impact: 🟡 Medium

**Reasoning:** This PR modifies critical webhook validation logic used in two user-facing GitHub subscription flows (`/github subscriptions add` for repos and organizations). The change introduces a fallback to organization-level webhooks and alters error handling semantics, but is isolated to a single function with new comprehensive test coverage (196 lines).

**Regression Risk:** The refactoring changes error suppression behavior—organization-level webhook lookup failures are now logged at debug level and swallowed rather than propagated when checking repository-level subscriptions. This is intentional as a best-effort fallback, but could mask underlying permission or configuration issues. The function now performs additional API calls which may impact GitHub rate-limiting behavior. The impact is partially mitigated by the new test suite covering repo-level webhooks, org-level fallback scenarios, permission errors (403/404), and absence cases.

**QA Recommendation:** Manual testing is recommended for the primary user scenarios: repository subscriptions with repo-level webhooks, repository subscriptions with org-level webhooks only (the bug fix case), and missing webhooks at both levels. Test with various permission levels to ensure graceful degradation when org-level access is unavailable. Regression testing is warranted given the error handling changes; skipping manual QA is not advisable for this user-facing critical path.

*Generated by CodeRabbitAI*

### DIFF
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -409,47 +409,68 @@ func (p *Plugin) createPost(channelID, userID, message string) error {
 }
 
 func (p *Plugin) checkIfConfiguredWebhookExists(ctx context.Context, githubClient *github.Client, userInfo *GitHubUserInfo, repo, owner string) (bool, error) {
-	found := false
-	opt := &github.ListOptions{
-		PerPage: PerPageValue,
-	}
 	siteURL, err := getSiteURL(p.client)
 	if err != nil {
 		return false, err
 	}
 
+	// For an organization subscription only the organization-level webhooks are relevant.
+	if repo == "" {
+		return p.findWebhookForSiteURL(userInfo, siteURL, func(opt *github.ListOptions) ([]*github.Hook, *github.Response, error) {
+			return githubClient.Organizations.ListHooks(ctx, owner, opt)
+		})
+	}
+
+	// For a repository subscription, check the repository-level webhooks first.
+	found, err := p.findWebhookForSiteURL(userInfo, siteURL, func(opt *github.ListOptions) ([]*github.Hook, *github.Response, error) {
+		return githubClient.Repositories.ListHooks(ctx, owner, repo, opt)
+	})
+	if err != nil || found {
+		return found, err
+	}
+
+	// The webhook may instead be configured at the organization level, which still
+	// delivers events for this repository. Listing organization webhooks requires
+	// org-admin scope that a repository subscriber often lacks (yielding a 403/404), so
+	// treat this as a best-effort fallback: swallow its error and report "not found"
+	// rather than propagating, preserving the caller's existing behavior for such users.
+	found, orgErr := p.findWebhookForSiteURL(userInfo, siteURL, func(opt *github.ListOptions) ([]*github.Hook, *github.Response, error) {
+		return githubClient.Organizations.ListHooks(ctx, owner, opt)
+	})
+	if orgErr != nil {
+		p.client.Log.Debug("Unable to check organization webhooks as fallback for repository subscription", "Owner", owner, "Repo", repo, "error", orgErr.Error())
+		return false, nil
+	}
+
+	return found, nil
+}
+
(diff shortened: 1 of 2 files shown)


server/plugin/command.go
```diff
@@ -409,47 +409,68 @@ func (p *Plugin) createPost(channelID, userID, message string) error {
...
+
+	// The webhook may instead be configured at the organization level, which still
+	// delivers events for this repository. Listing organization webhooks requires
+	// org-admin scope that a repository subscriber often lacks (yielding a 403/404), so
+	// treat this as a best-effort fallback: swallow its error and report "not found"
+	// rather than propagating, preserving the caller's existing behavior for such users.
+	found, orgErr := p.findWebhookForSiteURL(userInfo, siteURL, func(opt *github.ListOptions) ([]*github.Hook, *github.Response, error) {
+		return githubClient.Organizations.ListHooks(ctx, owner, opt)
+	})
+	if orgErr != nil {
```
nang2049, line 440: The doc comment calls out 403/404 as the expected reason and that's right. But the current code also swallows transient errors (5xx, network) silently with just a Log.Debug. Worth gating the swallow to the expected status codes.

server/plugin/command_webhook_test.go
```diff
@@ -0,0 +1,196 @@
+// Copyright (c) 2018-present Mattermost, Inc. All Rights Reserved.
```
nang2049, line 1: Nice coverage but every fixture fits in one page so the NextPage loop in findWebhookForSiteURL is unexercised. Worth one more test case where the org level listing returns a Link: <…>; rel="next" header and the matching webhook is on page 2.

nang2049 (commented):
LGTM with some minor comments

ewwollesen:
Closing in favor of #1027, which was merged on 2026-06-23 and resolves MM-69266 with a simpler approach. Thanks @nang2049!

PR #1026, opened 2026-06-16
feat: add refresh ability to GitHub RHS sidebar

### LINKED ISSUE #131: Github Plugin: Add Refresh ability to RHS

rafaumeu:
## Summary
Adds the ability to refresh data in the GitHub plugin RHS sidebar, as requested in #131.

Two improvements:
1. **Auto-refresh on open** - When the RHS is opened, `getSidebarContent` is called automatically to guarantee the latest data is available.
2. **Manual refresh button** - A refresh icon (fa-refresh with tooltip) is added to the RHS header. Clicking it triggers `getSidebarContent` with a spinner animation while loading.

Resolves #131

## Changes
- `sidebar_right/index.jsx`: Wire `getSidebarContent` action to `SidebarRight`
- `sidebar_right/sidebar_right.jsx`: Add refresh state, handler, auto-refresh on mount, and refresh button in header

## Release Notes
```release-note
Added the ability to manually refresh and auto-refresh data in the GitHub RHS sidebar.
```

## Change Impact: 🟡 Medium

**Reasoning:** The work is isolated to the GitHub RHS refresh UX (wiring `getSidebarContent` and adding UI concurrency/unmount safety), but it introduces new async request timing and state transitions that can be sensitive to rapid user interactions and view switching.  
**Regression Risk:** Moderate due to potential race-condition edge cases (rapid opens/clicks, unmount during in-flight refresh), though the blast radius is limited to sidebar components and Redux action usage and there are targeted automated tests.  
**QA Recommendation:** Perform targeted manual QA for RHS open/view switch triggering refresh, manual refresh button spinner/tooltip behavior, concurrent/deduped refresh while pending, unmount/navigation during refresh, and correct PRs/reviews/unreads/assignments rendering on both light and dark themes; manual QA can be skipped only if the full automated suite is green and you’ve validated the specific refresh flows in a staging environment.
*Generated by CodeRabbitAI*

### DIFF
--- a/webapp/src/components/sidebar_right/sidebar_right.jsx
+++ b/webapp/src/components/sidebar_right/sidebar_right.jsx
@@ -4,6 +4,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import Scrollbars from 'react-custom-scrollbars-2';
+import {OverlayTrigger, Tooltip} from 'react-bootstrap';
 
 import {RHSStates} from '../../constants';
 
@@ -78,10 +79,31 @@ export default class SidebarRight extends React.PureComponent {
         actions: PropTypes.shape({
             getYourPrsDetails: PropTypes.func.isRequired,
             getReviewsDetails: PropTypes.func.isRequired,
+            getSidebarContent: PropTypes.func.isRequired,
         }).isRequired,
     };
 
+    constructor(props) {
+        super(props);
+        this.state = {refreshing: false};
+    }
+
+    handleRefresh = async (e) => {
+        if (e) {
+            e.preventDefault();
+        }
+        if (this.state.refreshing) {
+            return;
+        }
+        this.setState({refreshing: true});
+        await this.props.actions.getSidebarContent();
+        this.setState({refreshing: false});
+    };
+
     componentDidMount() {
+        // Auto-refresh on open to guarantee latest data (issue #131)
+        this.handleRefresh();
+
         if (this.props.yourPrs && this.props.rhsState === RHSStates.PRS) {
             this.props.actions.getYourPrsDetails(mapGithubItemListToPrList(this.props.yourPrs));
         }
@@ -163,6 +185,19 @@ export default class SidebarRight extends React.PureComponent {
                                 rel='noopener noreferrer'
                             >{title}</a>
                         </strong>
+                        <OverlayTrigger
+                            key='rhsRefreshButton'
+                            placement='left'
+                            overlay={<Tooltip id='rhsRefreshTooltip'>{'Refresh'}</Tooltip>}
+                        >
+                            <a
+                                href='#'
+                                style={style.refreshButton}
+                                onClick={this.handleRefresh}
+                            >
+                                <i className={'fa fa-refresh' + (this.state.refreshing ? ' fa-spin' : '')}/>
+                            </a>
+                        </OverlayTrigger>
                     </div>
                     <div>
                         <GithubItems
@@ -181,5 +216,12 @@ export default class SidebarRight extends React.PureComponent {
 const style = {
     sectionHeader: {
         padding: '15px',
+        display: 'flex',
+        justifyContent: 'space-between',
+        alignItems: 'center',
+    },
+    refreshButton: {
+        color: 'rgba(0, 0, 0, 0.4)',
+        cursor: 'pointer',
     },
 };
--- a/webapp/src/components/sidebar_right/sidebar_right.jsx
+++ b/webapp/src/components/sidebar_right/sidebar_right.jsx
@@ -6,8 +6,9 @@ import PropTypes from 'prop-types';
 import Scrollbars from 'react-custom-scrollbars-2';
 import {OverlayTrigger, Tooltip} from 'react-bootstrap';
 
-import {RHSStates} from '../../constants';
+import {makeStyleFromTheme, changeOpacity} from 'mattermost-redux/utils/theme_utils';
 
+import {RHSStates} from '../../constants';
 import GithubItems from './github_items';
 
 export function renderView(props) {
@@ -15,7 +16,8 @@ export function renderView(props) {
         <div
             {...props}
             className='scrollbar--view'
-        />);
+        />
+    );
 }
 
 export function renderThumbHorizontal(props) {
@@ -23,7 +25,8 @@ export function renderThumbHorizontal(props) {
         <div
             {...props}
             className='scrollbar--horizontal'
-        />);
+        />
+    );
 }
 
 export function renderThumbVertical(props) {
@@ -31,7 +34,8 @@ export function renderThumbVertical(props) {
         <div
             {...props}
             className='scrollbar--vertical'
-        />);
+        />
+    );
 }
 
 function mapGithubItemListToPrList(gilist) {
@@ -86,23 +90,12 @@ export default class SidebarRight extends React.PureComponent {
     constructor(props) {
         super(props);
         this.state = {refreshing: false};
+        this._mounted = false;
+        this._refreshing = false;
     }
 
-    handleRefresh = async (e) => {
-        if (e) {
-            e.preventDefault();
-        }
-        if (this.state.refreshing) {
-            return;
-        }
-        this.setState({refreshing: true});
-        await this.props.actions.getSidebarContent();
-        this.setState({refreshing: false});
-    };
-
     componentDidMount() {
-        // Auto-refresh on open to guarantee latest data (issue #131)
-        this.handleRefresh();
+        this._mounted = true;
 
         if (this.props.yourPrs && this.props.rhsState === RHSStates.PRS) {
             this.props.actions.getYourPrsDetails(mapGithubItemListToPrList(this.props.yourPrs));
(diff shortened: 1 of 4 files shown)


webapp/src/components/sidebar_right/sidebar_right.jsx
```diff
@@ -78,10 +79,31 @@ export default class SidebarRight extends React.PureComponent {
...
+            return;
+        }
+        this.setState({refreshing: true});
+        await this.props.actions.getSidebarContent();
+        this.setState({refreshing: false});
+    };
+
     componentDidMount() {
+        // Auto-refresh on open to guarantee latest data (issue #131)
+        this.handleRefresh();
```
nang2049: The auto refresh on mount is redundant because getSidebarContent() is already dispatched from `sidebar_buttons.jsx:42` and `websocket/index.js:74,84`. By the time the RHS opens the data is fresh. 

This adds an extra GitHub search call every time the user opens the RHS, which is expensive. I'd suggest removing the auto refresh entirely and relying on the manual button and existing WS-driven updates. If you want to keep it, please mirror the E2E guard from sidebar_buttons.jsx.

webapp/src/components/sidebar_right/sidebar_right.jsx
```diff
@@ -78,10 +79,31 @@ export default class SidebarRight extends React.PureComponent {
...
+            getSidebarContent: PropTypes.func.isRequired,
         }).isRequired,
     };
 
+    constructor(props) {
+        super(props);
+        this.state = {refreshing: false};
+    }
+
+    handleRefresh = async (e) => {
```
nang2049: I see a couple of issues here. `setState` is async so two fast clicks can both pass the this.state.refreshing check before either update lands. Use an instance flag for the gate. We also need unmount safety.

Suggested rewrite:
```
componentDidMount() {
    this._mounted = true;
    // ...
}
componentWillUnmount() {
    this._mounted = false;
}
handleRefresh = async (e) => {
    e?.preventDefault();
    if (this._refreshing) {
        return;
    }
    this._refreshing = true;
    this.setState({refreshing: true});
    try {
        await this.props.actions.getSidebarContent();
    } finally {
        this._refreshing = false;
        if (this._mounted) {
            this.setState({refreshing: false});
        }
    }
};
```

webapp/src/components/sidebar_right/sidebar_right.jsx
```diff
@@ -163,6 +185,19 @@ export default class SidebarRight extends React.PureComponent {
                                 rel='noopener noreferrer'
                             >{title}</a>
                         </strong>
+                        <OverlayTrigger
+                            key='rhsRefreshButton'
+                            placement='left'
+                            overlay={<Tooltip id='rhsRefreshTooltip'>{'Refresh'}</Tooltip>}
+                        >
+                            <a
```
nang2049: This is a button and not a link. <a href='#'> is announced as a link by screen readers and will navigate to # jumping the page to top. Please use a real <button> and add an aria-label since the tooltip is sighted only.

webapp/src/components/sidebar_right/sidebar_right.jsx
```diff
@@ -181,5 +216,12 @@ export default class SidebarRight extends React.PureComponent {
 const style = {
     sectionHeader: {
         padding: '15px',
+        display: 'flex',
+        justifyContent: 'space-between',
+        alignItems: 'center',
+    },
+    refreshButton: {
+        color: 'rgba(0, 0, 0, 0.4)',
```
nang2049: Hard-coded color will be invisible on dark themes. this.props.theme is already available on the component please derive the color the same way sidebar_buttons.jsx does. Please convert the bottom-of-file style object to a memoized factory:

```
const getStyle = makeStyleFromTheme((theme) => ({
    sectionHeader: {
        padding: '15px',
        display: 'flex',
        justifyContent: 'space-between',
        alignItems: 'center',
    },
    refreshButton: {
        color: changeOpacity(theme.centerChannelColor, 0.6),
        cursor: 'pointer',
        background: 'transparent',
        border: 'none',
        padding: 0,
    },
}));
```

webapp/src/components/sidebar_right/sidebar_right.jsx
```diff
@@ -181,5 +216,12 @@ export default class SidebarRight extends React.PureComponent {
 const style = {
     sectionHeader: {
         padding: '15px',
+        display: 'flex',
```
nang2049: This switches every RHS state's header to flex layout not just the ones with the refresh button. Could you attach screenshots of all four states (PRS, REVIEWS, UNREADS, ASSIGNMENTS) on both light and dark themes? Want to make sure the title alignment doesn't regress anywhere.

nang2049 (changes requested):
Thank you so much @rafaumeu for your contribution. I left a few comments :)

nang2049:
@coderabbitai review

rafaumeu:
Thanks for the detailed review @nang2049! I implemented all 5 suggestions per your comments: 1) Removed auto-refresh in componentDidMount (redundant because WS + sidebar_buttons already fetch); 2) Added instance flag + unmount guard to prevent race condition; 3) Changed <a> to <button> with aria-label='Refresh'; 4) Replaced hardcoded color with theme utils and opacity like in sidebar_buttons; 5) Verified all 4 states (PRS, REVIEWS, UNREADS, ASSIGNMENTS) – no visual regressions on light or dark themes; the flex layout already applies to all states so we only added button styling inline.

PTAL!

rafaumeu:
Hi @nang2049, I've addressed all the review feedback in commit 21dde9a:

1. **Removed auto-refresh on mount** — getSidebarContent() is already dispatched from sidebar_buttons.jsx and websocket/index.js, so the redundant refresh is gone.
2. **Fixed race condition** — Added instance flag `this._refreshing` with `componentWillUnmount` safety check to prevent double-click issues.
3. **Changed `<a href='#'>` to `<button>`** — Now uses `<button>` with `aria-label='Refresh'` for proper screen reader support.
4. **Theme-aware colors** — All styles now use `makeStyleFromTheme` + `changeOpacity` instead of hard-coded values (consistent with sidebar_buttons.jsx).
5. **Fixed syntax error** — Removed the duplicated `handleRefresh` body that was causing the class to break.

Could you please re-review when you have a chance? Thanks!

webapp/src/components/sidebar_right/sidebar_right.jsx
```diff
@@ -102,6 +140,25 @@ export default class SidebarRight extends React.PureComponent {
     }
 
     render() {
+        const getStyle = makeStyleFromTheme((theme) => {
```
nang2049: Nice :) theme handling looks right. One small follow-up: makeStyleFromTheme returns a theme memoized function but by declaring getStyle inside render() we build a brand-new factory on every render which defeats the memoization. Please hoist it to module scope same as the snippet in my earlier comment:
```
const getStyle = makeStyleFromTheme((theme) => ({
    sectionHeader: { ... },
    refreshButton: { ... },
}));
```
Then just call const style = getStyle(this.props.theme); inside render().

nang2049 (commented):
Thanks @rafaumeu !

webapp/src/components/sidebar_right/sidebar_right.jsx
rafaumeu: ✅ Addressed in commit 0d143a3 — Hoisted `getStyle` to module scope. Now `makeStyleFromTheme` creates the factory only once instead of on every render, and `render()` just calls `getStyle(this.props.theme)`.

rafaumeu:
Hi @nang2049, thanks for the follow-up! I'\''ve hoisted `getStyle` to module scope in commit 0d143a3 so `makeStyleFromTheme` only creates the factory once instead of on every render. `render()` now just calls `getStyle(this.props.theme)`.

ogi-m (commented):
Hi @rafaumeu, I've tested it and when the GitHub RHS panel is opened, it doesn't automatically load new data and displays stale data instead. The user must click the manual refresh button to see current data.

https://github.com/user-attachments/assets/1b6e7289-7d32-409a-8163-de657e49f5c7

According to Claude:

```
Root cause: SidebarRight is permanently mounted by registerRightHandSidebarComponent — its componentDidMount fires only once at plugin load, never on panel open. The openRHS handler in sidebar_buttons.jsx only calls updateRhsState() + showRHSPlugin() with no data fetch. There is no isOpen visibility prop available to detect panel open events.

Suggested fix (webapp/src/components/sidebar_buttons/sidebar_buttons.jsx, openRHS):

openRHS = (rhsState) => {
    this.props.actions.updateRhsState(rhsState);
    this.props.showRHSPlugin();
    if (this.props.connected) {
        this.getData(); // fetch fresh data on open/tab switch
    }
};

getData(), connected, and getSidebarContent are all already wired up on SidebarButtons — no other files need to change. The side effect is that switching tabs also refreshes data (acceptable, since there's no visibility prop to distinguish panel open from tab switch).
```

ogi-m:
Just went back and read through the comments, the auto-refresh on open was removed intentionally? So now it's only about the manual button, right @rafaumeu ?

rafaumeu:
Thanks for testing and for the detailed diagnosis, @ogi-m. Fixed in commit `fc68b3c`: opening/reopening the GitHub RHS or switching its view now requests fresh sidebar content. The refresh path shares the existing in-flight guard, so rapid tab switches do not create duplicate requests. I also added regression tests for both behaviors and ran lint, type-check, all tests, and the production build successfully.

nang2049 (approved):
Thanks @rafaumeu, all my earlier comments are addressed.

@ogi-m to answer your question: auto-refresh on RHS open is back but just moved to `openRHS` in `sidebar_buttons.jsx` per your suggestion instead of `SidebarRight.componentDidMount` since `SidebarRight` is permanently mounted. So you get both fresh data on open and tab-switch and the manual refresh button.

LGTM.

ogi-m (approved):
Thanks @rafaumeu , working as expected now!

PR #1027, opened 2026-06-19
[MM-69266] Fix false "No webhook was found" warning when only an org-…

nang2049:
#### Summary
When a channel is subscribed to a specific repository via `/github subscribe owner/repo` and webhook delivery is handled by an org level webhook the plugin incorrectly posts:
>  No webhook was found for this repository or organization. To create one, enter the following slash command `/github setup webhook`
and this message is a false negative.
This change makes the verification consistent with the rest of the plugin. No change to webhook delivery, subscription creation or command parsing

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-69266

#### QA
 - In GitHub: `<org>`  - Settings - Webhooks add a webhook with Payload URL `https://<mattermost>/plugins/github/webhook`.
 - Make sure the target repo has no repo level webhook with that URL.
 - In Mattermost: `/github subscribe <org>/<repo>`.
 - Expected: subscription confirmation no "No webhook was found" note. 

## Change Impact: 🟡 Medium

**Reasoning:** This PR changes the logic that determines whether to show a user-facing “No webhook was found” warning during `/github subscribe`, specifically treating certain GitHub webhook-list access failures (403/404) during org-hook checks as “configured” so organization-level webhooks are recognized. While it does not alter webhook delivery, it does change the plugin’s behavioral contract for warning/UX decisions in permission-related scenarios.

**Regression Risk:** Moderate—changes are localized to `command.go`, but the new fallback/error-handling behavior can affect multiple subscription outcomes (org-owned vs user-owned repos) and the decision to suppress warnings when webhook listing is blocked; while unit tests cover the main org-owned/403 scenario, they don’t exhaustively cover all permission/error variants (e.g., 404 access error vs other errors, pagination edge cases).

**QA Recommendation:** Perform targeted manual QA focused on the subscription warning behavior: (1) repo-level webhook configured only, (2) org-level webhook configured only, (3) org-owned repo where listing org webhooks returns 403/404, (4) confirm no false warning appears in the described reproduction flow. Automated tests added for the core logic; manual QA should be limited to these scenarios due to moderate risk.
*Generated by CodeRabbitAI*

### DIFF
--- a/server/plugin/command.go
+++ b/server/plugin/command.go
@@ -409,57 +409,73 @@ func (p *Plugin) createPost(channelID, userID, message string) error {
 }
 
 func (p *Plugin) checkIfConfiguredWebhookExists(ctx context.Context, githubClient *github.Client, userInfo *GitHubUserInfo, repo, owner string) (bool, error) {
-	found := false
-	opt := &github.ListOptions{
-		PerPage: PerPageValue,
-	}
 	siteURL, err := getSiteURL(p.client)
 	if err != nil {
 		return false, err
 	}
 
+	listOrgHooks := func(opt *github.ListOptions) ([]*github.Hook, *github.Response, error) {
+		return githubClient.Organizations.ListHooks(ctx, owner, opt)
+	}
+
+	if repo == "" {
+		return p.anyHookMatchesSiteURL(userInfo, owner, repo, siteURL, listOrgHooks)
+	}
+
+	found, err := p.anyHookMatchesSiteURL(userInfo, owner, repo, siteURL, func(opt *github.ListOptions) ([]*github.Hook, *github.Response, error) {
+		return githubClient.Repositories.ListHooks(ctx, owner, repo, opt)
+	})
+	if err != nil {
+		return false, err
+	}
+	if found {
+		return true, nil
+	}
+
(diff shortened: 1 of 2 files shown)


server/plugin/command.go
```diff
@@ -409,57 +409,73 @@ func (p *Plugin) createPost(channelID, userID, message string) error {
...
+		return githubClient.Repositories.ListHooks(ctx, owner, repo, opt)
+	})
+	if err != nil {
+		return false, err
+	}
+	if found {
+		return true, nil
+	}
+
+	found, err = p.anyHookMatchesSiteURL(userInfo, owner, repo, siteURL, listOrgHooks)
```
avasconcelos114, line 443: We may have to skip this org fallback if the repository's owner is a user instead of an org, otherwise we may get a false-positive where the 404 response from the listOrgHooks request cascades into `isWebhookListAccessError` being true and returning `true` instead of leaving it to the final `return found, nil` which was going to be correctly marked as false since the user hadn't set up webhooks for their personal repo

avasconcelos114 (commented):
Changes are looking good! I just have one possible edge-case that might need to be confirmed

server/plugin/command.go
nang2049: @avasconcelos114 added

avasconcelos114 (approved):
LGTM!

PR #1045, opened 2026-08-02
Fix lost-update race in channel subscriptions

aegonmyy:
# Fix lost-update race in channel subscriptions

## Summary

All channel subscriptions are stored in a single KV key (`subscriptions`).
Every mutation reads the whole blob, changes it in memory, and writes the whole
blob back. That read-modify-write was not atomic, so two subscribe or
unsubscribe operations running at the same time in different channels could
overwrite each other. The plugin reported success to both callers, but only one
of the two changes actually persisted. The other was silently lost.

## Root cause

`StoreSubscriptions` in `server/plugin/subscriptions.go` called
`SetAtomicWithRetries`, but its callback ignored the `oldValue` argument and
returned a snapshot of the subscriptions that was captured before the retry
loop began:

```go
// server/plugin/subscriptions.go (before)
func (p *Plugin) StoreSubscriptions(s *Subscriptions) error {
	return p.store.SetAtomicWithRetries(SubscriptionsKey, func(_ []byte) (any, error) {
		modifiedBytes, err := json.Marshal(s)   // s was read before the loop
		...
	})
}
```

`AddSubscription` (`subscriptions.go:306`) and `Unsubscribe`
(`subscriptions.go:422`) both did `GetSubscriptions()` to read the whole blob,
mutated the in-memory copy, then handed that stale copy to `StoreSubscriptions`.

`SetAtomicWithRetries` is supposed to re-run its callback on each retry so the
change is re-applied on top of the freshly read value. Because the callback
ignored `oldValue` and always marshaled the same pre-read snapshot, a retry
after a conflicting concurrent write just rewrote stale data. The compare-and-set
still succeeded, so the concurrent writer's change was dropped without any error.

## The fix

Move the read-modify-write inside the atomic callback. A new
`modifySubscriptions` helper decodes `oldValue` on every attempt, applies a
mutate closure to that fresh state, and returns the re-encoded result. On a
conflicting write the retry re-reads the latest blob and re-applies the change,
so nothing is lost.

`AddSubscription` and `Unsubscribe` now express their change as a mutate closure
passed to the helper. `GetSubscriptions` and `GetSubscriptionsByChannel` are
unchanged. Error types and return values are preserved.

## Regression test

`TestSubscriptionRace` (`server/plugin/subscription_race_test.go`) fires 200
concurrent `AddSubscription` calls against the vendored `pluginapi.MemoryStore`,
whose `SetAtomicWithRetries` is byte-identical to the production KV service. It
then compares how many calls reported success against how many subscriptions
actually persisted. It fails on the old code and passes with this fix.

### Before the fix

```
=== RUN   TestSubscriptionRace
    subscription_race_test.go:63: concurrent AddSubscription calls : 200
    subscription_race_test.go:64: AddSubscription returned success : 188
    subscription_race_test.go:65: subscriptions actually persisted : 1
    subscription_race_test.go:66: silently lost (success but gone) : 187
    subscription_race_test.go:69: lost-update bug: 187 subscriptions were reported as saved but silently dropped
--- FAIL: TestSubscriptionRace (0.06s)
FAIL
```

### After the fix

```
=== RUN   TestSubscriptionRace
    subscription_race_test.go:63: concurrent AddSubscription calls : 200
    subscription_race_test.go:64: AddSubscription returned success : 111
    subscription_race_test.go:65: subscriptions actually persisted : 111
    subscription_race_test.go:66: silently lost (success but gone) : 0
--- PASS: TestSubscriptionRace (0.07s)
```

After the fix, every call that reports success is durable. Under heavy
contention some calls exhaust the five internal retries and return an error
instead. Those are reported as failures and are not counted as success, which is
the correct behavior.

## Change Impact: 🔴 High

**Reasoning:** The change modifies subscription persistence and data integrity logic. It also removes the public `StoreSubscriptions` method.

**Regression Risk:** Concurrent subscription updates and unsubscribe behavior changed. The main race path has regression coverage, but other mutation and storage-error paths remain possible regression areas.

**QA Recommendation:** Perform focused manual QA for concurrent add, unsubscribe, duplicate, not-found, and storage-error cases. Skipping manual QA carries moderate risk because the change affects shared persistence behavior.

*Generated by CodeRabbitAI*

### DIFF
--- a/server/plugin/subscription_race_test.go
+++ b/server/plugin/subscription_race_test.go
@@ -68,4 +68,11 @@ func TestSubscriptionRace(t *testing.T) {
 	if silentlyLost > 0 {
 		t.Fatalf("lost-update bug: %d subscriptions were reported as saved but silently dropped", silentlyLost)
 	}
+
+	// Guard against a false pass: if every AddSubscription call failed,
+	// reportedSuccess and persisted are both 0, so silentlyLost is 0 and the
+	// check above passes without actually exercising the race.
+	if reportedSuccess == 0 {
+		t.Fatalf("no AddSubscription calls succeeded; the race was not exercised")
+	}
 }
--- a/server/plugin/subscription_race_test.go
+++ b/server/plugin/subscription_race_test.go
@@ -1,0 +1,71 @@
+// Copyright (c) 2018-present Mattermost, Inc. All Rights Reserved.
+// See LICENSE.txt for license information.
+
+package plugin
+
+import (
+	"fmt"
+	"sync"
+	"testing"
+
+	"github.com/mattermost/mattermost/server/public/pluginapi"
+)
+
+// TestSubscriptionRace drives the plugin's real AddSubscription against the
+// vendored pluginapi.MemoryStore, whose SetAtomicWithRetries is byte-identical
+// to the production KVService. No plugin logic is mocked. Many channels
+// subscribe to the same repository concurrently; afterwards we compare how many
+// AddSubscription calls reported success (nil error) against how many
+// subscriptions actually persisted.
+//
+// Correct behaviour: persisted == reported-success (every reported success is
+// durable). Before the fix, the read-modify-write happened outside the atomic
+// callback, so concurrent writers clobbered each other and persisted was far
+// lower than reported-success (silent lost updates).
+func TestSubscriptionRace(t *testing.T) {
+	const numChannels = 200
+
+	p := NewPlugin()
+	p.client = pluginapi.NewClient(p.API, p.Driver)
+	p.store = &pluginapi.MemoryStore{}
+
+	var wg sync.WaitGroup
+	var mu sync.Mutex
+	reportedSuccess := 0
+
+	for i := 0; i < numChannels; i++ {
+		wg.Add(1)
+		go func(i int) {
+			defer wg.Done()
+			sub := &Subscription{
+				ChannelID:  fmt.Sprintf("channel-%d", i),
+				Repository: "org/repo",
+			}
+			if err := p.AddSubscription("org/repo", sub); err == nil {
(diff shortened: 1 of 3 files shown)


server/plugin/subscription_race_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	silentlyLost := reportedSuccess - persisted
+	t.Logf("concurrent AddSubscription calls : %d", numChannels)
+	t.Logf("AddSubscription returned success : %d", reportedSuccess)
+	t.Logf("subscriptions actually persisted : %d", persisted)
+	t.Logf("silently lost (success but gone) : %d", silentlyLost)
+
+	if silentlyLost > 0 {
+		t.Fatalf("lost-update bug: %d subscriptions were reported as saved but silently dropped", silentlyLost)
+	}
```
jgheithcock, line 70: (Similar to https://github.com/mattermost/mattermost-plugin-gitlab/pull/691.)
If all AddSubscription calls fail, subs will be empty and persisted will be 0 and so silentlyLost will be 0 and no failures will be reported. Suggest adding the following after silentlyLost check:

if reportedSuccess == 0 {
    t.Fatalf("No AddSubscription calls succeeded")
}

jgheithcock (changes requested):

server/plugin/subscription_race_test.go
aegonmyy: Good catch, thanks. Added a guard after the silentlyLost check that fails the test if reportedSuccess == 0, so a run where every AddSubscription call failed (persisted 0, silentlyLost 0) no longer passes silently. Pushed in 235d14d.

jgheithcock (approved):
LGTM! Thanks @aegonmyy !

